PR #2, opened 2019-02-08
CDAP-14748 Separate MySQL plugin from other database plugins

ibarbashov:


database-commons/src/main/java/co/cask/ConnectionConfig.java
```diff
@@ -32,20 +33,27 @@
...
+  @Name(HOST)
+  @Description("Database host")
+  public String host;
+
+  @Name(PORT)
+  @Description("Specific database port")
+  public Integer port;
+
+  @Name(DATABASE)
+  @Description("Database name to connect")
```
albertshau: to connect to

database-commons/src/main/java/co/cask/ConnectionConfig.java
```diff
@@ -116,15 +103,23 @@ public static Properties getConnectionArguments(@Nullable String connectionArgum
...
    * @return a {@link Properties} of connection arguments, parsed from the config.
    */
   public Properties getConnectionArguments() {
-    return getConnectionArguments(connectionArguments, user, password);
+    Properties arguments = getConnectionArguments(this.connectionArguments, user, password);
+    arguments.putAll(getDBSpecificArguments());
+    return arguments;
   }
+
+  public Map<String, String> getDBSpecificArguments() {
```
albertshau: better to be protected, since we don't expect anything else to call this.

database-commons/src/main/java/co/cask/db/batch/sink/DBSink.java
```diff
@@ -137,49 +164,49 @@ public void transform(StructuredRecord input, Emitter<KeyValue<DBRecord, NullWri
...
     DBUtils.cleanup(driverClass);
-    dbManager.destroy();
+    if (driverCleanup != null) {
+      driverCleanup.destroy();
+    }
   }
 
   @VisibleForTesting
-  void setColumns(List<String> columns) {
+  public void setColumns(List<String> columns) {
```
albertshau, line 172: is this method needed anymore

database-commons/src/main/java/co/cask/db/batch/sink/DBSink.java
```diff
@@ -34,91 +32,120 @@
...
         "Table %s does not exist. Please check that the 'tableName' property " +
           "has been set correctly, and that the connection string %s points to a valid database.",
-        dbSinkConfig.tableName, dbSinkConfig.connectionString);
+        dbSinkConfig.tableName, connectionString);
     } finally {
       DBUtils.cleanup(driverClass);
     }
-    context.addOutput(Output.of(dbSinkConfig.referenceName, new DBOutputFormatProvider(dbSinkConfig, driverClass)));
+
+    setColumnsInfo(context.getInputSchema().getFields());
```
albertshau: it's possible for the input schema to be null if incoming records can have multiple schemas, or if the schema is not known. If it's null, we should just fail the run with some message about the DBSink requiring a single known schema.

database-commons/src/main/java/co/cask/db/batch/sink/DBSink.java
```diff
@@ -34,91 +32,120 @@
...
+                                new DBOutputFormatProvider(dbSinkConfig, connectionString, dbColumns, driverClass)));
+  }
+
+  private void setColumnsInfo(List<Schema.Field> fields) {
+    columns = fields.stream()
+      .map(Schema.Field::getName)
+      .collect(collectingAndThen(toList(), Collections::unmodifiableList));
+
+    dbColumns = null;
+    if (columns.size() > 1) {
```
albertshau: you can use Collectors.joining() to do this type of logic.

database-commons/src/main/java/co/cask/db/batch/sink/DBSink.java
albertshau: Also, it looks like DB plugins are not recording field level lineage. After this PR, we need to add it (similar to https://github.com/data-integrations/google-cloud/blob/develop/src/main/java/co/cask/gcp/bigquery/BigQuerySink.java#L148).

database-commons/src/main/java/co/cask/db/batch/source/DBSource.java
```diff
@@ -59,58 +58,49 @@
...
 
   public DBSource(DBSourceConfig sourceConfig) {
     super(new ReferencePluginConfig(sourceConfig.referenceName));
     this.sourceConfig = sourceConfig;
-    this.dbManager = new DBManager(sourceConfig);
+  }
+
+  private static String removeConditionsClause(String importQuerySring) {
+    importQuerySring = importQuerySring.replaceAll("\\s{2,}", " ").toUpperCase();
+    if (importQuerySring.contains("WHERE $CONDITIONS AND")) {
```
albertshau: this may catch many of the common possibilities, but it is still possible to specify a query that will cause this to generate an invalid query.

Let's have a final  `else (importQueryString.contains("$CONDITIONS"))` and throw an exception that asks the user to remove the CONDITIONS clause when fetching the input schema.

database-commons/src/main/java/co/cask/db/batch/source/DBSource.java
```diff
@@ -59,58 +58,49 @@
...
+  protected final DBSourceConfig sourceConfig;
+  protected Class<? extends Driver> driverClass;
 
   public DBSource(DBSourceConfig sourceConfig) {
     super(new ReferencePluginConfig(sourceConfig.referenceName));
     this.sourceConfig = sourceConfig;
-    this.dbManager = new DBManager(sourceConfig);
+  }
+
+  private static String removeConditionsClause(String importQuerySring) {
```
albertshau: typo: Sring -> String

database-commons/src/main/java/co/cask/db/batch/source/DBSource.java
```diff
@@ -243,79 +239,69 @@ public void initialize(BatchRuntimeContext context) throws Exception {
   @Override
   public void transform(KeyValue<LongWritable, DBRecord> input, Emitter<StructuredRecord> emitter) throws Exception {
     emitter.emit(StructuredRecordUtils.convertCase(
-      input.getValue().getRecord(), FieldCase.toFieldCase(sourceConfig.columnNameCase)));
+      input.getValue().getRecord(), FieldCase.NONE));
```
albertshau: just remove the call to convertCase()

database-commons/src/main/java/co/cask/db/batch/source/DBSource.java
```diff
@@ -243,79 +239,69 @@ public void initialize(BatchRuntimeContext context) throws Exception {
...
     public static final String BOUNDING_QUERY = "boundingQuery";
     public static final String SPLIT_BY = "splitBy";
     public static final String NUM_SPLITS = "numSplits";
     public static final String SCHEMA = "schema";
     public static final String TRANSACTION_ISOLATION_LEVEL = "transactionIsolationLevel";
-
+    @Nullable
+    @Name(TRANSACTION_ISOLATION_LEVEL)
+    @Description("The transaction isolation level for queries run by this sink. " +
+      "Defaults to TRANSACTION_SERIALIZABLE. See java.sql.Connection#setTransactionIsolation for more details. " +
```
albertshau: it looks like this was added specifically for phoenix. This should not show up for the non-generic db plugins.

Does it make sense to have it for mysql?

database-commons/src/main/java/co/cask/util/DBUtils.java
```diff
@@ -327,6 +332,22 @@ public static void deregisterAllDrivers(Class<? extends Driver> driverClass)
     }
   }
 
+  public static void validateJDBCPluginPipeline(PipelineConfigurer pipelineConfigurer,
+                                                ConnectionConfig config, String jdbcPluginId, String jdbcDriverName) {
+    Preconditions.checkArgument(!(config.user == null && config.password != null),
```
albertshau: in CDAP 6.0.0, we added an InvalidConfigPropertyException that should be thrown when a plugin is performing validation on its config properties and notices that one is invalid. Please change the code so that it is used instead of IllegalArgumentException

generic-database-plugin/docs/Database-action.md
```diff
@@ -0,0 +1,55 @@
...
+**Connection Arguments:** A list of arbitrary string tag/value pairs as connection arguments. These arguments
+will be passed to the JDBC driver, as connection arguments, for JDBC drivers that may need additional configurations.
+This is a semicolon-separated list of key-value pairs, where each pair is separated by a equals '=' and specifies
+the key and value for the argument. For example, 'key1=value1;key2=value' specifies that the connection will be
+given arguments 'key1' mapped to 'value1' and the argument 'key2' mapped to 'value2'. (Macro-enabled)
+
+Example
+-------
+
+    {
```
albertshau: We no longer use json in the examples, as that is confusing to UI users. Instead, just use sentences to describe an example use case.

generic-database-plugin/docs/Database-action.md
```diff
@@ -0,0 +1,55 @@
...
+
+**Username:** User identity for connecting to the specified database.
+
+**Password:** Password to use to connect to the specified database.
+
+**Driver Name:** Name of the JDBC driver to use.
+
+**Connection Arguments:** A list of arbitrary string tag/value pairs as connection arguments. These arguments
+will be passed to the JDBC driver, as connection arguments, for JDBC drivers that may need additional configurations.
+This is a semicolon-separated list of key-value pairs, where each pair is separated by a equals '=' and specifies
```
albertshau: We no longer describe the format expected by the backend, as it's not visible to UI users. You can remove this sentence.

generic-database-plugin/docs/Database-action.md
```diff
@@ -0,0 +1,55 @@
...
+the key and value for the argument. For example, 'key1=value1;key2=value' specifies that the connection will be
+given arguments 'key1' mapped to 'value1' and the argument 'key2' mapped to 'value2'. (Macro-enabled)
```
albertshau: We also no longer include (Macro-enabled), as the UI has an icon whenever a property supports macros.

generic-database-plugin/docs/Database-batchsink.md
```diff
@@ -0,0 +1,62 @@
...
+
+**Username:** User identity for connecting to the specified database. (Macro-enabled)
+
+**Password:** Password to use to connect to the specified database. (Macro-enabled)
+
+**Driver Name:** Name of the JDBC driver to use.
+
+**Connection Arguments:** A list of arbitrary string tag/value pairs as connection arguments. These arguments
+will be passed to the JDBC driver, as connection arguments, for JDBC drivers that may need additional configurations.
+This is a semicolon-separated list of key-value pairs, where each pair is separated by a equals '=' and specifies
```
albertshau: similar comment about removing this.

generic-database-plugin/docs/Database-batchsink.md
```diff
@@ -0,0 +1,62 @@
...
+
+**Transaction Isolation Level:** The transaction isolation level for queries run by this sink.
+Defaults to TRANSACTION_SERIALIZABLE. See java.sql.Connection#setTransactionIsolation for more details.
+The Phoenix jdbc driver will throw an exception if the Phoenix database does not have transactions enabled
+and this setting is set to true. For drivers like that, this should be set to TRANSACTION_NONE.
+
+Example
+-------
+
+    {
```
albertshau: similar comment about writing this example in sentences rather than a json.

generic-database-plugin/docs/Database-batchsource.md
```diff
@@ -0,0 +1,91 @@
...
+
+**Database:** database name.
+
+**Driver Name:** Name of the JDBC driver to use.
+
+**Import Query:** The SELECT query to use to import data from the specified table.
+You can specify an arbitrary number of columns to import, or import all columns using \*. The Query should
+contain the '$CONDITIONS' string. For example, 'SELECT * FROM table WHERE $CONDITIONS'.
+The '$CONDITIONS' string will be replaced by 'splitBy' field limits specified by the bounding query.
+The '$CONDITIONS' string is not required if numSplits is set to one. (Macro-enabled)
```
albertshau: remove (Macro-enabled), as UI now has an icon.

generic-database-plugin/docs/Database-batchsource.md
```diff
@@ -0,0 +1,91 @@
...
+
+
+Example
+-------
+
+    {
+        "name": "Database",
+        "type": "batchsource",
+        "properties": {
+            "importQuery": "select id,name,email,phone from users where $CONDITIONS",
```
albertshau: similar comment about describing the example in text rather than json

generic-database-plugin/docs/Database-postaction.md
```diff
@@ -0,0 +1,73 @@
...
+
+**Driver Name:** Name of the JDBC driver to use.
+
+**Connection Arguments:** A list of arbitrary string tag/value pairs as connection arguments. These arguments
+will be passed to the JDBC driver, as connection arguments, for JDBC drivers that may need additional configurations.
+This is a semicolon-separated list of key-value pairs, where each pair is separated by a equals '=' and specifies
+the key and value for the argument. For example, 'key1=value1;key2=value' specifies that the connection will be
+given arguments 'key1' mapped to 'value1' and the argument 'key2' mapped to 'value2'. (Macro-enabled)
+
+
```
albertshau: remove extra newlines

generic-database-plugin/docs/Database-postaction.md
```diff
@@ -0,0 +1,73 @@
...
+**Password:** Password to use to connect to the specified database. (Macro-enabled)
+
+**Driver Name:** Name of the JDBC driver to use.
+
+**Connection Arguments:** A list of arbitrary string tag/value pairs as connection arguments. These arguments
+will be passed to the JDBC driver, as connection arguments, for JDBC drivers that may need additional configurations.
+This is a semicolon-separated list of key-value pairs, where each pair is separated by a equals '=' and specifies
+the key and value for the argument. For example, 'key1=value1;key2=value' specifies that the connection will be
+given arguments 'key1' mapped to 'value1' and the argument 'key2' mapped to 'value2'. (Macro-enabled)
```
albertshau: remove Macro-enabled in this file

generic-database-plugin/src/main/java/co/cask/jdbc/DatabaseConstants.java
```diff
@@ -0,0 +1,29 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package co.cask.jdbc;
+
+/**
+ * MySQL Constants.
```
albertshau: MySQL -> Database

generic-database-plugin/src/main/java/co/cask/jdbc/DatabasePostAction.java
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  /**
+   * Database post action databaseQueryActionConfig.
+   */
+  public static class DatabaseQueryActionConfig extends QueryActionConfig {
+    @Name(DatabaseConstants.JDBC_DRIVER_NAME)
+    @Description("Name of the JDBC driver to use. This is the value of the 'jdbcDriverName' key defined in the JSON " +
+            "file for the JDBC plugin.")
+    public String jdbcDriverName;
```
albertshau: This should be `@Nullable` and default to "mysql".

generic-database-plugin/src/main/java/co/cask/jdbc/DatabasePostAction.java
```diff
@@ -0,0 +1,55 @@
...
+  public String getJdbcDriverName() {
+    return databaseQueryActionConfig.jdbcDriverName;
+  }
+
+  /**
+   * Database post action databaseQueryActionConfig.
+   */
+  public static class DatabaseQueryActionConfig extends QueryActionConfig {
+    @Name(DatabaseConstants.JDBC_DRIVER_NAME)
+    @Description("Name of the JDBC driver to use. This is the value of the 'jdbcDriverName' key defined in the JSON " +
```
albertshau: Name of the JDBC driver to use. This is only required if you are using a different driver than the default "mysql" driver.

generic-database-plugin/src/main/java/co/cask/jdbc/DatabaseSink.java
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  /**
+   * Generic database sink configuration.
+   */
+  public static class DatabaseSinkConfig extends DBSinkConfig {
+    @Name(DatabaseConstants.JDBC_DRIVER_NAME)
+    @Description("Name of the JDBC driver to use. This is the value of the 'jdbcDriverName' key defined in the JSON " +
+            "file for the JDBC plugin.")
+    public String jdbcDriverName;
```
albertshau: similar comments about making these nullable and default to "mysql"

generic-database-plugin/src/test/java/co/cask/db/batch/action/DBActionTestRun.java
```diff
@@ -38,6 +39,7 @@
 /**
  * Test for DBAction Plugin
  */
+@Ignore // Probably need to be converted to specific database provider and moved in integration tests repo.
```
albertshau: why ignore? The generic DB plugins should still work and be used with hypersql to test the plugins

generic-database-plugin/widgets/Database-action.json
```diff
@@ -58,8 +68,8 @@
             "showDelimiter": "false",
             "key-placeholder": "Key",
             "value-placeholder": "Value",
-            "kv-delimiter": "=",
-            "delimiter": ";"
+            "kv-delimiter" : "=",
```
albertshau: unintentional change?

database-commons/src/main/java/co/cask/ConnectionConfig.java
```diff
@@ -32,20 +33,27 @@
...
-  public static final String JDBC_PLUGIN_NAME = "jdbcPluginName";
-  public static final String JDBC_PLUGIN_TYPE = "jdbcPluginType";
-  public static final String COLUMN_NAME_CASE = "columnNameCase";
-  public static final String ENABLE_AUTO_COMMIT = "enableAutoCommit";
+  public static final String JDBC_PLUGIN_TYPE = "jdbc";
 
-  @Name(CONNECTION_STRING)
-  @Description("JDBC connection string including database name.")
-  @Macro
-  public String connectionString;
```
albertshau, line 48: The generic DB plugins should still require the user to configure the connection string, since JDBC drivers can do whatever they want with the URL. It's only the DB specific plugins that should expose the separate properties for host, port, db, etc.

mysql-plugin/docs/Mysql-action.md
```diff
@@ -0,0 +1,56 @@
...
+The action can be used whenever you want to run a Mysql command before or after a data pipeline.
+For example, you may want to run a sql update command on a database before the pipeline source pulls data from tables.
+
+
+Properties
+----------
+**Database Command:** The database command to execute.
+
+**Host:** host where MySQL running on.
+**Port:** MySQL port default 3306.
```
albertshau: add newline between properties

mysql-plugin/docs/Mysql-action.md
```diff
@@ -0,0 +1,56 @@
...
+
+Use Case
+--------
+The action can be used whenever you want to run a Mysql command before or after a data pipeline.
+For example, you may want to run a sql update command on a database before the pipeline source pulls data from tables.
+
+
+Properties
+----------
+**Database Command:** The database command to execute.
```
albertshau: Database command to execute.

mysql-plugin/docs/Mysql-action.md
```diff
@@ -0,0 +1,56 @@
...
+
+**Host:** host where MySQL running on.
```
albertshau: Host that MySQL is running on

mysql-plugin/docs/Mysql-action.md
albertshau: Port that MySQL is running on.

mysql-plugin/docs/Mysql-action.md
```diff
@@ -0,0 +1,56 @@
...
+**Port:** MySQL port default 3306.
+**Database:** MySQL database name.
+
+**Username:** User identity for connecting to the specified database.
+
+**Password:** Password to use to connect to the specified database.
+
+**Connection Arguments:** A list of arbitrary string tag/value pairs as connection arguments. These arguments
+will be passed to the JDBC driver, as connection arguments, for JDBC drivers that may need additional configurations.
+This is a semicolon-separated list of key-value pairs, where each pair is separated by a equals '=' and specifies
```
albertshau: remove information about the format

mysql-plugin/docs/Mysql-action.md
```diff
@@ -0,0 +1,56 @@
...
+the key and value for the argument. For example, 'key1=value1;key2=value' specifies that the connection will be
+given arguments 'key1' mapped to 'value1' and the argument 'key2' mapped to 'value2'. (Macro-enabled)
+
+**Auto Reconnect** Should the driver try to re-establish stale and/or dead connections.
+
+**Allow Multi Queries** Allow the use of ';' to delimit multiple queries during one statement (true/false).
```
albertshau: I think this should just always be true and not configurable. Or is there a use case for disallowing multiple queries?

mysql-plugin/docs/Mysql-batchsink.md
```diff
@@ -0,0 +1,62 @@
...
+The model is stored in a FileSet and you want to export the contents
+of the FileSet to a Mysql table where it can be served to your users.
+
+Column names would be autodetected from input schema.
+
+Properties
+----------
+**Reference Name:** Name used to uniquely identify this sink for lineage, annotating metadata, etc.
+
+**Host:** host where MySQL running on.
```
albertshau: similar comments about language

mysql-plugin/docs/Mysql-batchsource.md
```diff
@@ -0,0 +1,97 @@
...
+The source is used whenever you need to read from a Mysql. For example, you may want
+to create daily snapshots of a database table by using this source and writing to
+a TimePartitionedFileSet.
+
+
+Properties
+----------
+**Reference Name:** Name used to uniquely identify this sink for lineage, annotating metadata, etc.
+
+**Host:** host where MySQL running on.
```
albertshau: similar comments about language

mysql-plugin/widgets/Mysql-batchsink.json
```diff
@@ -0,0 +1,127 @@
+{
+  "metadata": {
+    "spec-version": "1.5"
+  },
+  "icon": {
```
albertshau: icon should be built by the maven plugin (through files in the icons directory) and not specified inline here.

albertshau: See https://github.com/data-integrations/google-cloud as an example

mysql-plugin/src/main/java/co/cask/mysql/MysqlSource.java
```diff
@@ -0,0 +1,70 @@
...
+   */
+  public static class MysqlSourceConfig extends DBSourceConfig {
+    @Name(MysqlConstants.AUTO_RECONNECT)
+    @Description("Should the driver try to re-establish stale and/or dead connections")
+    public Boolean autoReconnect;
+
+    @Name(MysqlConstants.MAX_ROWS)
+    @Nullable
+    @Description("The maximum number of rows to return (0, the default means return all rows)")
+    public Integer maxRows;
```
albertshau: I don't think we should have this property. It would actually be a max for each split, which translates to something pretty confusing.

albertshau (commented):
Restructuring looks mostly good, but please make sure the generic db plugins work as they did before and that the unit tests are working.

generic-database-plugin/src/main/java/co/cask/jdbc/DatabasePostAction.java
ibarbashov: Are you sure that generic database plugin's driver name should default to mysql?

generic-database-plugin/src/test/java/co/cask/db/batch/action/DBActionTestRun.java
ibarbashov: I temporarily ignored this test before we made decision to move generic plugin to a separate module. We'll use existing hsql tests to verify generic-database-plugin functionality and write dedicated tests for DB-specific plugins.

mysql-plugin/src/main/java/co/cask/mysql/MysqlSource.java
ibarbashov: Agree. Single split batch job is probably not the common real world use case.

generic-database-plugin/widgets/Database-action.json
ibarbashov: yes, fixed

database-commons/src/main/java/co/cask/db/batch/source/DBSource.java
ibarbashov: It's not required by MySQL though a lot of code relies on transactionIsolationLevel  
from DBSource and DBSink configs. It's not possible to move it from DBSinkConfig since  OutputFormatProvider expects it to be there. Although, we can completely remove the dedicated TRANSACTION_ISOLATION property and specify it as part of advanced settings via key-value fields.

database-commons/src/main/java/co/cask/ConnectionConfig.java
ibarbashov: It's now called by other classes. Since we can't directly put host, name and port into ConnectionConfig (otherwise these fields will be displayed for generic DB plugin) we now provide them as DB-specific via Map.

database-commons/src/main/java/co/cask/db/batch/sink/DBSink.java
ibarbashov: I address this in following PR with integration tests.

mysql-plugin/docs/Mysql-action.md
ibarbashov: Can't think of any. Removed from widget.

database-commons/src/main/java/co/cask/db/batch/sink/DBSink.java
ibarbashov: Will do.

generic-database-plugin/src/test/java/co/cask/db/batch/action/DBActionTestRun.java
albertshau: This PR should just be adding a mysql specific plugin. It should *not* break the existing DB plugins.

generic-database-plugin/src/main/java/co/cask/jdbc/DatabasePostAction.java
albertshau: sorry ignore this comment, this is only for the mysql specific plugins.

database-commons/src/main/java/co/cask/db/batch/source/DBSource.java
albertshau: It sounds like it doesn't belong in MySQL, which means it shouldn't be in the base config. The DB specific configs should not extend the config for the generic DB plugin. They should all extend a common parent that only has the properties that are exactly the same (descriptions included) for all of them.

database-commons/src/main/java/co/cask/ConnectionConfig.java
albertshau: Why is this called and not getConnectionProperties()? I thought this was meant to be implemented by the DB specific configs to translate their DB specific properties into generic connection properties.

database-commons/src/main/java/co/cask/db/batch/sink/DBSink.java
```diff
@@ -130,15 +132,7 @@ private void setColumnsInfo(List<Schema.Field> fields) {
...
-    dbColumns = null;
-    if (columns.size() > 1) {
-
-      dbColumns = String.join(",", columns);
-
-    } else if (columns.size() == 1) {
-      dbColumns = columns.get(0);
-    }
-
+    dbColumns = String.join(",", columns);
```
albertshau, line 134: this is not what I commented about. You can using the joining collector above in the stream.

database-commons/src/main/java/co/cask/util/DBUtils.java
```diff
@@ -414,7 +419,17 @@ private static void unregisterOracleMBean(ClassLoader classLoader) {
   }
 
   public static String createConnectionString(ConnectionConfig config, String jdbcDriverName) {
-    return createConnectionString(config.host, config.port, config.database, jdbcDriverName);
+    Map<String, String> dbSpecificArguments = config.getDBSpecificArguments();
+    if (dbSpecificArguments.containsKey(ConnectionConfig.CONNECTION_STRING)) {
```
albertshau: Using special map keys and values is not a good idea. It also means the connection string will be used as a connection property which is not right.

This createConnectionString method does not belong in a central util class. Each individual config should be responsible for creating the connection string, as they are the ones who have knowledge about the their jdbc driver expects. You can add a getConnectionString() method to ConnectionConfig that the various db specific configs can override.

generic-database-plugin/docs/Database-action.md
```diff
@@ -14,42 +14,27 @@ For example, you may want to run a sql update command on a database before the p
...
+**Connection String:** JDBC connection string including database name.
 
-**Database:** database name.
+**Database Command:** Database command to execute.
 
 **Username:** User identity for connecting to the specified database.
 
 **Password:** Password to use to connect to the specified database.
 
-**Driver Name:** Name of the JDBC driver to use.
```
albertshau: why is this removed

generic-database-plugin/docs/Database-action.md
```diff
@@ -14,42 +14,27 @@ For example, you may want to run a sql update command on a database before the p
...
-                "host": "localhost",
-                "port": 3306,
-                "database": "prod",
-                "user": "user123",
-                "password": "password-abc",
-                "jdbcDriverName": "mysql"
-            }
-        }
-    }
+Execute database statement "UPDATE table_name SET price = 20 WHERE ID = 6" against database "prod" running on localhost,
```
albertshau: Use complete sentences in the documentation.

Suppose you want to execute a query against a PostgreSQL database named "prod" that is running on "examplehost" port 5432. Ensure that the driver for PostgreSQL is installed, then configure the plugin with:

```
Driver Name: "postgres"
...
```

generic-database-plugin/docs/Database-batchsink.md
```diff
@@ -19,44 +19,31 @@ Properties
...
-**Table Name:** Name of the table to export to. (Macro-enabled)
+**Table Name:** Name of the table to export to.
 
-**Username:** User identity for connecting to the specified database. (Macro-enabled)
+**Username:** User identity for connecting to the specified database.
 
-**Password:** Password to use to connect to the specified database. (Macro-enabled)
+**Password:** Password to use to connect to the specified database.
 
-**Driver Name:** Name of the JDBC driver to use.
```
albertshau: This should not be removed from the generic plugin

generic-database-plugin/docs/Database-batchsink.md
```diff
@@ -19,44 +19,31 @@ Properties
...
 
 **Connection Arguments:** A list of arbitrary string tag/value pairs as connection arguments. These arguments
 will be passed to the JDBC driver, as connection arguments, for JDBC drivers that may need additional configurations.
-This is a semicolon-separated list of key-value pairs, where each pair is separated by a equals '=' and specifies
-the key and value for the argument. For example, 'key1=value1;key2=value' specifies that the connection will be
-given arguments 'key1' mapped to 'value1' and the argument 'key2' mapped to 'value2'. (Macro-enabled)
-
-**Transaction Isolation Level:** The transaction isolation level for queries run by this sink.
```
albertshau: This should not be removed from the generic plugin

generic-database-plugin/docs/Database-batchsink.md
```diff
@@ -19,44 +19,31 @@ Properties
...
-Defaults to TRANSACTION_SERIALIZABLE. See java.sql.Connection#setTransactionIsolation for more details.
-The Phoenix jdbc driver will throw an exception if the Phoenix database does not have transactions enabled
-and this setting is set to true. For drivers like that, this should be set to TRANSACTION_NONE.
 
 Example
 -------
+Write output records to "users" table of database "prod" running on localhost, port 3306, as "root" user with "root"
```
albertshau: similar comment, use complete sentences.

generic-database-plugin/docs/Database-batchsource.md
```diff
@@ -18,65 +18,49 @@ Properties
...
-**Password:** Password to use to connect to the specified database. (Macro-enabled)
+**Password:** Password to use to connect to the specified database.
 
 **Connection Arguments:** A list of arbitrary string tag/value pairs as connection arguments. These arguments
 will be passed to the JDBC driver, as connection arguments, for JDBC drivers that may need additional configurations.
-This is a semicolon-separated list of key-value pairs, where each pair is separated by a equals '=' and specifies
-the key and value for the argument. For example, 'key1=value1;key2=value' specifies that the connection will be
-given arguments 'key1' mapped to 'value1' and the argument 'key2' mapped to 'value2'. (Macro-enabled)
-
-**Transaction Isolation Level:** The transaction isolation level for queries run by this sink.
```
albertshau: This should not be removed from the generic plugin

generic-database-plugin/src/main/java/co/cask/jdbc/DatabaseSource.java
```diff
@@ -43,20 +43,10 @@ public DatabaseSource(DatabaseSourceConfig databaseSourceConfig) {
...
-  }
-
   /**
    * Generic database source configuration.
    */
   public static class DatabaseSourceConfig extends DBSourceConfig {
-    @Name(DatabaseConstants.JDBC_DRIVER_NAME)
-    @Description("Name of the JDBC driver to use. This is the value of the 'jdbcDriverName' key defined in the JSON " +
-            "file for the JDBC plugin.")
-    public String jdbcDriverName;
```
albertshau: why is the jdbc driver getting removed from all the DB configs?

albertshau:
In the future, please keep PRs confined to a single plugin. This one could have been broken up into sources, sink, action, postaction.

generic-database-plugin/src/main/java/co/cask/jdbc/DatabaseSource.java
ibarbashov: It's moved back to ConnectionConfig because you and I decided to keep this field for DB-specific plugins (with default value pre-populated). So now it is used by all configurations.

generic-database-plugin/docs/Database-action.md
ibarbashov: I was moving the properties so they appear in the same order as on UI but accidentally removed this. Fixed.

database-commons/src/main/java/co/cask/db/batch/sink/DBSink.java
ibarbashov: I'm not sure I understand what do you mean. We have two fields - List columns and String dbColumns. columns list is used several times for validating schema while dbColumns string is used for OutputFormat configuration and in query string. 

Sure, I can use joining collector to assign dbColumns but then I'll need to split the string in order to validate schema.

albertshau: oh didn't realize these were different variables. Ignore this comment

generic-database-plugin/docs/Database-action.md
ibarbashov: updated

generic-database-plugin/docs/Database-batchsink.md
ibarbashov: updated

generic-database-plugin/src/test/java/co/cask/db/batch/action/DBActionTestRun.java
ibarbashov: Since I changed the database-plugin classes to abstract, and also removed the "columns" field from generic connection config class, some tests demanded minor changes to work, and I submitted them in last commit

database-commons/src/main/java/co/cask/DBConfig.java
```diff
@@ -67,4 +52,8 @@ protected String cleanQuery(@Nullable String query) {
     }
     return query.substring(0, idx + 1);
   }
+
+  public String getTransactionIsolationLevel() {
```
albertshau, line 57: annotate methods that can return null with `@Nullable`

database-commons/src/main/java/co/cask/db/batch/action/DBRun.java
```diff
@@ -41,28 +42,24 @@ public DBRun(QueryConfig config, Class<? extends Driver> driverClass) {
...
-    DBManager dbManager = new DBManager(config);
-
+    DriverCleanup driverCleanup = null;
     try {
-      dbManager.ensureJDBCDriverIsAvailable(driverClass);
+      driverCleanup = DBUtils.ensureJDBCDriverIsAvailable(driverClass, config.getConnectionString(),
+                                                          config.jdbcDriverName);
 
       try (Connection connection = getConnection()) {
-        if (!config.enableAutoCommit) {
```
albertshau, line 50: This is still needed for the generic DB action (I think it's a hacky workaround to support the Hive jdbc driver). If it's no longer in the QueryConfig, the boolean should be passed into the constructor.

database-commons/src/main/java/co/cask/db/batch/action/DBRun.java
```diff
@@ -41,28 +42,24 @@ public DBRun(QueryConfig config, Class<? extends Driver> driverClass) {
...
-          connection.setAutoCommit(false);
-        }
         try (Statement statement = connection.createStatement()) {
           statement.execute(config.query);
-          if (!config.enableAutoCommit) {
```
albertshau, line 55: similar comment here

database-commons/src/main/java/co/cask/db/batch/action/DBRun.java
```diff
@@ -41,28 +42,24 @@ public DBRun(QueryConfig config, Class<? extends Driver> driverClass) {
...
       }
     } finally {
-      dbManager.destroy();
+      if (driverCleanup != null) {
+        driverCleanup.destroy();
+      }
     }
   }
 
   private Connection getConnection() throws SQLException {
```
albertshau: Not from your code, but let's remove this one line method and just call DriverManager.getConnection() directly.

database-commons/src/main/java/co/cask/db/batch/sink/DBSink.java
```diff
@@ -34,91 +31,117 @@
...
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.TreeMap;
-import javax.annotation.Nullable;
+import java.util.stream.Collectors;
 
+import static java.util.stream.Collectors.collectingAndThen;
```
albertshau: style nit: we don't use static imports in our projects, code should use 'Collectors.collectingAndThen'

generic-database-plugin/docs/Database-action.md
```diff
@@ -0,0 +1,42 @@
...
+
+**Connection String:** JDBC connection string including database name.
+
+**Database Command:** Database command to execute.
+
+**Username:** User identity for connecting to the specified database.
+
+**Password:** Password to use to connect to the specified database.
+
+
```
albertshau: nit: remove extra newline

generic-database-plugin/docs/Database-batchsink.md
```diff
@@ -0,0 +1,51 @@
...
+
+**Connection String:** JDBC connection string including database name.
+
+**Table Name:** Name of the table to export to.
+
+**Username:** User identity for connecting to the specified database.
+
+**Password:** Password to use to connect to the specified database.
+
+
```
albertshau: nit: remove extra newline

generic-database-plugin/pom.xml
```diff
@@ -0,0 +1,116 @@
...
+    </dependency>
+    <dependency>
+      <groupId>org.hsqldb</groupId>
+      <artifactId>hsqldb</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>co.cask.cdap</groupId>
+      <artifactId>cdap-api</artifactId>
+      <scope>compile</scope>
```
albertshau: cdap-api classes will always be available come from the CDAP system and should be in provided scope.

generic-database-plugin/src/main/java/co/cask/jdbc/DatabaseSource.java
```diff
@@ -0,0 +1,78 @@
...
+import javax.annotation.Nullable;
+
+/**
+ * Batch source to read from generic database.
+ */
+@Plugin(type = "batchsource")
+@Name(DatabaseConstants.PLUGIN_NAME)
+@Description("Reads from a database table(s) using a configurable SQL query." +
+  " Outputs one record for each row returned by the query.")
+public class DatabaseSource extends DBSource {
```
albertshau: Let's rename all the base classes (DBAction, DBSource, DBSink) to start with 'Abstract' (AbstractDBAction, AbstractDBSource, ...). Otherwise it can be a little confusing which one is the actual plugin and which one is the base class.

generic-database-plugin/widgets/Database-action.json
```diff
@@ -84,4 +76,4 @@
       ]
     }
   ]
-}
+}
```
albertshau: files should end with a newline (same for other widgets)

generic-database-plugin/widgets/Database-action.json
```diff
@@ -9,20 +9,12 @@
       "properties": [
         {
           "widget-type": "textbox",
-          "label": "Plugin Name",
-          "name": "jdbcPluginName"
+          "label": "Driver Name",
+          "name": "jdbcDriverName"
```
albertshau: Changing the label from "Plugin Name" to "Driver Name" is good. But let's not change the name in the backend because that breaks backwards compatibility. Older pipelines that are upgraded will no longer work.

generic-database-plugin/widgets/Database-batchsink.json
```diff
@@ -17,20 +17,12 @@
...
-          "label": "Plugin Type",
-          "name": "jdbcPluginType",
-          "widget-attributes": {
-            "default": "jdbc"
-          }
-        },
-        {
-          "widget-type": "textbox",
-          "label": "Connection String",
+          "label": "Connection string",
```
albertshau: By convention, labels capitalize most words. Let's keep it 'Connection String' to make it consistent with everything else.

mysql-plugin/docs/Mysql-batchsource.md
```diff
@@ -0,0 +1,85 @@
...
+Use Case
+--------
+The source is used whenever you need to read from a Mysql. For example, you may want
+to create daily snapshots of a database table by using this source and writing to
+a TimePartitionedFileSet.
+
+
+Properties
+----------
+**Reference Name:** Name used to uniquely identify this sink for lineage, annotating metadata, etc.
```
albertshau: sink -> source

mysql-plugin/docs/Mysql-batchsource.md
```diff
@@ -0,0 +1,85 @@
...
+**Database:** MySQL database name.
+
+**Import Query:** The SELECT query to use to import data from the specified table.
+You can specify an arbitrary number of columns to import, or import all columns using \*. The Query should
+contain the '$CONDITIONS' string. For example, 'SELECT * FROM table WHERE $CONDITIONS'.
+The '$CONDITIONS' string will be replaced by 'splitBy' field limits specified by the bounding query.
+The '$CONDITIONS' string is not required if numSplits is set to one. (Macro-enabled)
+
+**Bounding Query:** Bounding Query should return the min and max of the values of the 'splitBy' field.
+For example, 'SELECT MIN(id),MAX(id) FROM table'. Not required if numSplits is set to one. (Macro-enabled)
```
albertshau: remove (Macro-enabled) from all the properties

mysql-plugin/docs/Mysql-batchsink.md
```diff
@@ -0,0 +1,56 @@
...
+
+**Driver Name:** Name of the JDBC driver to use.
+
+**Host:** Host that MySQL is running on.
+
+**Port:** Port that MySQL is running on.
+
+**Database:** MySQL database name.
+
+**Table Name:** Name of the table to export to. (Macro-enabled)
```
albertshau: remove (Macro-enabled) from all the properties. This was a relic from a time before the UI displayed an icon for macro enabled fields.

mysql-plugin/docs/Mysql-postaction.md
```diff
@@ -0,0 +1,64 @@
...
+
+**Query:** Query to run.
+
+**Host:** Host that MySQL is running on.
+
+**Port:** Port that MySQL is running on.
+
+**Database:** Mysql database name.
+
+**Username:** User identity for connecting to the specified database. (Macro-enabled)
```
albertshau: remove (Macro-enabled) from all properties

mysql-plugin/pom.xml
```diff
@@ -0,0 +1,116 @@
...
+    </dependency>
+    <dependency>
+      <groupId>org.hsqldb</groupId>
+      <artifactId>hsqldb</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>co.cask.cdap</groupId>
+      <artifactId>cdap-api</artifactId>
+      <scope>compile</scope>
```
albertshau: should be provided scope

mysql-plugin/src/main/java/co/cask/mysql/MysqlConstants.java
```diff
@@ -0,0 +1,33 @@
...
+ */
+public final class MysqlConstants {
+  private MysqlConstants() {
+    throw new AssertionError("Should not instantiate static utility class.");
+  }
+
+  public static final String PLUGIN_NAME = "Mysql";
+  public static final String AUTO_RECONNECT = "autoReconnect";
+  public static final String ALLOW_MULTIPLE_QUERIES = "allowMultiQueries";
+  public static final String MYSQL_CONNECTION_STRING_FORMAT = ConnectionConfig.JDBC_PLUGIN_TYPE + ":mysql://%s:%s/%s";
```
albertshau: it's better to just hardcode 'jdbc:mysql' instead of using the JDBC_PLUGIN_TYPE. Though they are the same value, the two concepts are actually unrelated.

albertshau (commented):
a few minor comments

database-commons/src/main/java/co/cask/DBConfig.java
ibarbashov: done

database-commons/src/main/java/co/cask/db/batch/action/DBRun.java
ibarbashov: fixed

database-commons/src/main/java/co/cask/db/batch/action/DBRun.java
ibarbashov: fixed

database-commons/src/main/java/co/cask/db/batch/action/DBRun.java
ibarbashov: done

database-commons/src/main/java/co/cask/db/batch/sink/DBSink.java
ibarbashov: done

generic-database-plugin/docs/Database-action.md
ibarbashov: removed

generic-database-plugin/docs/Database-batchsink.md
ibarbashov: removed

generic-database-plugin/pom.xml
ibarbashov: done

generic-database-plugin/src/main/java/co/cask/jdbc/DatabaseSource.java
ibarbashov: done

generic-database-plugin/widgets/Database-action.json
ibarbashov: fixed

generic-database-plugin/widgets/Database-action.json
ibarbashov: fixed

generic-database-plugin/widgets/Database-batchsink.json
ibarbashov: fixed

mysql-plugin/docs/Mysql-batchsource.md
ibarbashov: fixed

mysql-plugin/docs/Mysql-batchsource.md
ibarbashov: removed

mysql-plugin/docs/Mysql-batchsink.md
ibarbashov: removed

mysql-plugin/docs/Mysql-postaction.md
ibarbashov: removed

mysql-plugin/pom.xml
ibarbashov: fixed

mysql-plugin/src/main/java/co/cask/mysql/MysqlConstants.java
ibarbashov: done

albertshau (approved):
lgtm. Please squash commits and I'll merge

ibarbashov:
> lgtm. Please squash commits and I'll merge

Done

PR #4, opened 2019-02-19
Feature/postgres plugin

ibarbashov:


postgresql-plugin/src/main/java/co/cask/postgres/PostgresAction.java
```diff
@@ -0,0 +1,65 @@
...
+import com.google.common.collect.ImmutableMap;
+
+import java.util.Map;
+
+/**
+ * Action that runs PostgreSQL command.
+ */
+@Plugin(type = Action.PLUGIN_TYPE)
+@Name(PostgresConstants.PLUGIN_NAME)
+@Description("Action that runs a MySQL command")
```
albertshau: MySQL -> PostgreSQL (same for other plugins)

albertshau (commented):
minor comment about plugin descriptions

postgresql-plugin/src/main/java/co/cask/postgres/PostgresAction.java
ibarbashov: done

PR #9, opened 2019-02-26
Feature/oracle plugin tests

ibarbashov:

### DIFF
--- a/database-commons/src/main/java/co/cask/DBRecord.java
+++ b/database-commons/src/main/java/co/cask/DBRecord.java
@@ -117,7 +117,7 @@ public class DBRecord implements Writable, DBWritable, Configurable {
   }
 
   protected void setField(ResultSet resultSet, StructuredRecord.Builder recordBuilder, Schema.Field field, int sqlType,
-                        int sqlPrecision, int sqlScale) throws SQLException {
+                          int sqlPrecision, int sqlScale) throws SQLException {
     Object o = DBUtils.transformValue(sqlType, sqlPrecision, sqlScale, resultSet, field.getName());
     if (o instanceof Date) {
       recordBuilder.setDate(field.getName(), ((Date) o).toLocalDate());
--- a/database-commons/src/main/java/co/cask/DBRecord.java
+++ b/database-commons/src/main/java/co/cask/DBRecord.java
@@ -52,15 +52,15 @@ import javax.sql.rowset.serial.SerialBlob;
  * @see DBWritable DBWritable
  */
 public class DBRecord implements Writable, DBWritable, Configurable {
-  private StructuredRecord record;
-  private Configuration conf;
+  protected StructuredRecord record;
+  protected Configuration conf;
 
   /**
    * Need to cache {@link ResultSetMetaData} of the record for use during writing to a table.
    * This is because we cannot rely on JDBC drivers to properly set metadata in the {@link PreparedStatement}
    * passed to the #write method in this class.
    */
-  private int[] columnTypes;
+  protected int[] columnTypes;
 
   /**
    * Used to construct a DBRecord from a StructuredRecord in the ETL Pipeline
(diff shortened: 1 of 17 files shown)


oracle-plugin/pom.xml
```diff
@@ -48,6 +55,19 @@
...
+    <dependency>
+      <groupId>co.cask</groupId>
+      <artifactId>database-commons</artifactId>
+      <version>${project.version}</version>
+      <type>test-jar</type>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.oracle</groupId>
+      <artifactId>ojdbc6</artifactId>
```
albertshau: what's the license for this?

database-commons/src/main/java/co/cask/DBRecord.java
```diff
@@ -105,12 +105,18 @@ public void readFields(ResultSet resultSet) throws SQLException {
...
   }
 
-  private void setField(ResultSet resultSet, StructuredRecord.Builder recordBuilder, Schema.Field field, int sqlType,
+  protected void handleField(ResultSet resultSet, StructuredRecord.Builder recordBuilder, Schema.Field field,
+                             int sqlType, int sqlPrecision, int sqlScale) throws SQLException {
+    setField(resultSet, recordBuilder, field, sqlType, sqlPrecision, sqlScale);
+  }
+
+  protected void setField(ResultSet resultSet, StructuredRecord.Builder recordBuilder, Schema.Field field, int sqlType,
                         int sqlPrecision, int sqlScale) throws SQLException {
```
albertshau: nit: indentation

oracle-plugin/src/main/java/co/cask/oracle/OracleDBRecord.java
```diff
@@ -0,0 +1,75 @@
...
+  private void handleOracleSpecificType(ResultSet resultSet,
+                                        StructuredRecord.Builder recordBuilder, Schema.Field field,
+                                        int sqlType) throws SQLException {
+
+    Object original = resultSet.getObject(field.getName());
+
+    switch (sqlType) {
+      case INTERVAL_YM:
+      case INTERVAL_DS:
+        recordBuilder.set(field.getName(), original.toString());
```
albertshau, line 71: what do these interval types represent? I'm wondering if toString() is a proper value for them or if some other type is more appropriate

oracle-plugin/src/main/java/co/cask/oracle/OracleSink.java
```diff
@@ -42,12 +52,30 @@ public OracleSink(OracleSinkConfig oracleSinkConfig) {
     this.oracleSinkConfig = oracleSinkConfig;
   }
 
+  @Override
+  public void transform(StructuredRecord input, Emitter<KeyValue<DBRecord, NullWritable>> emitter) throws Exception {
+    List<Schema.Field> outputFields = new ArrayList<>();
+    for (String column : columns) {
+      Schema.Field field = input.getSchema().getField(column);
+      Preconditions.checkNotNull(field, "Missing schema field for column '%s'", column);
```
albertshau: Column '%s' not found in an input record.

oracle-plugin/src/main/java/co/cask/oracle/OracleSink.java
```diff
@@ -42,12 +52,30 @@ public OracleSink(OracleSinkConfig oracleSinkConfig) {
...
+      outputFields.add(field);
+    }
+    StructuredRecord.Builder output = StructuredRecord.builder(
+      Schema.recordOf(input.getSchema().getRecordName(), outputFields));
+    for (String column : columns) {
```
albertshau: is this done in the base DB plugin? Is this copy being created in case the columns are a subset of the input record's schema? Does the OutputFormat not handle that case?

oracle-plugin/pom.xml
ibarbashov: I believe it's proprietary, but It's one of the few Oracle JDBC driver .jars available publicly. To get driver from Oracle one have to register on their website filling lots of info (including company name, job title, corporate phone №, etc.) and then accept OTN licence agreement.

oracle-plugin/src/main/java/co/cask/oracle/OracleDBRecord.java
ibarbashov: Interval represents time period, e.g. for INTERVAL_YM string value '300-5' means '300 years and 5 months'. The toString() method of Oracle's datatype implementation is overridden to return string representation of interval. The invocation of resultSet.getString("interval") returns the same value as resultSet.getObject("interval').toString(). The above also applies to INTERVAL DAY TO SECOND type.

database-commons/src/main/java/co/cask/DBRecord.java
ibarbashov: done

oracle-plugin/src/main/java/co/cask/oracle/OracleSink.java
ibarbashov: done

oracle-plugin/pom.xml
albertshau: If it's not apache2, mit, or some other similar permissive license, we can't use it. I couldn't find the license after a quick search, do you have a link to it or anything?

ibarbashov: Yeah, it's not,
https://www.oracle.com/technetwork/licenses/distribution-license-152002.html, 
I will send you an email with a possible solutions

albertshau:
the changes look good, but we still need a solution without bringing in that dependency.

pom.xml
```diff
@@ -367,6 +367,9 @@
           <systemPropertyVariables>
             <java.io.tmpdir>${project.build.directory}</java.io.tmpdir>
           </systemPropertyVariables>
+          <additionalClasspathElements>
+            <additionalClasspathElement>${jdbcDriverJars}</additionalClasspathElement>
```
albertshau, line 373: If you run a test that doesn't require this and you don't specify the property (ex: the generic db test), does it complain about this property not being defined? If so, it would be better to default this to some value so that it's not always required.

ibarbashov: done, I provided empty placeholder for that case

PR #10, opened 2019-02-27
Feature/db2 plugin

ibarbashov:


db2-plugin/pom.xml
```diff
@@ -0,0 +1,118 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  Copyright © 2015-2018 Cask Data, Inc.
```
albertshau: year should be 2019 (did we miss this in the other poms too?)

albertshau (commented):
just one comment about copyright

albertshau (approved):
lgtm

PR #11, opened 2019-03-01
CDAP-14747 Netezza plugins

ibarbashov:


netezza-plugin/docs/Netezza-batchsource.md
```diff
@@ -0,0 +1,83 @@
...
+
+**Port:** Port that Netezza is running on.
+
+**Database:** Netezza database name.
+
+**Import Query:** The SELECT query to use to import data from the specified table.
+You can specify an arbitrary number of columns to import, or import all columns using \*. The Query should
+contain the '$CONDITIONS' string. For example, 'SELECT * FROM table WHERE $CONDITIONS'.
+The '$CONDITIONS' string will be replaced by 'splitBy' field limits specified by the bounding query.
+The '$CONDITIONS' string is not required if numSplits is set to one. (Macro-enabled)
```
albertshau: Don't need the (Macro-enabled). I may have missed this in the other PRs, can you check if they are there in the others?

netezza-plugin/docs/Netezza-batchsource.md
```diff
@@ -0,0 +1,83 @@
...
+
+**Bounding Query:** Bounding Query should return the min and max of the values of the 'splitBy' field.
+For example, 'SELECT MIN(id),MAX(id) FROM table'. Not required if numSplits is set to one.
+
+**Split-By Field Name:** Field Name which will be used to generate splits. Not required if numSplits is set to one. (Macro-enabled)
```
albertshau: Don't need the (Macro-enabled).

netezza-plugin/pom.xml
```diff
@@ -0,0 +1,118 @@
...
+    <dependency>
+      <groupId>co.cask.cdap</groupId>
+      <artifactId>cdap-api</artifactId>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.jetbrains</groupId>
+      <artifactId>annotations</artifactId>
+      <version>RELEASE</version>
+      <scope>compile</scope>
```
albertshau: should this be test scope?

netezza-plugin/src/main/java/co/cask/netezza/NetezzaDBRecord.java
```diff
@@ -0,0 +1,78 @@
...
+  protected void handleField(ResultSet resultSet, StructuredRecord.Builder recordBuilder, Schema.Field field,
+                             int sqlType, int sqlPrecision, int sqlScale) throws SQLException {
+    if (netezzaTypes.contains(sqlType)) {
+      handleOracleSpecificType(resultSet, recordBuilder, field, sqlType);
+    } else {
+      setField(resultSet, recordBuilder, field, sqlType, sqlPrecision, sqlScale);
+    }
+  }
+
+  private void handleOracleSpecificType(ResultSet resultSet,
```
albertshau: rename Oracle -> Netezza

netezza-plugin/src/main/java/co/cask/netezza/NetezzaDBRecord.java
```diff
@@ -0,0 +1,78 @@
...
+                                        StructuredRecord.Builder recordBuilder, Schema.Field field,
+                                        int sqlType) throws SQLException {
+
+    Object original = resultSet.getObject(field.getName());
+
+    switch (sqlType) {
+      case Types.VARBINARY:
+        original = resultSet.getBytes(field.getName());
```
albertshau: what is the actual type here? byte[]?

netezza-plugin/src/main/java/co/cask/netezza/NetezzaDBRecord.java
```diff
@@ -0,0 +1,78 @@
...
+    } else {
+      setField(resultSet, recordBuilder, field, sqlType, sqlPrecision, sqlScale);
+    }
+  }
+
+  private void handleOracleSpecificType(ResultSet resultSet,
+                                        StructuredRecord.Builder recordBuilder, Schema.Field field,
+                                        int sqlType) throws SQLException {
+
+    Object original = resultSet.getObject(field.getName());
```
albertshau: this is unused if it's a VARBINARY so it's better to get it in the INTERVAL case instead of here.

netezza-plugin/src/main/java/co/cask/netezza/NetezzaDBRecord.java
```diff
@@ -0,0 +1,78 @@
...
+
+    switch (sqlType) {
+      case Types.VARBINARY:
+        original = resultSet.getBytes(field.getName());
+        recordBuilder.set(field.getName(), original);
+        break;
+      case INTERVAL:
+        recordBuilder.set(field.getName(), original.toString());
```
albertshau: What is the actual type returned in this case?

What does toString() for that type return?

netezza-plugin/src/main/java/co/cask/netezza/NetezzaDBRecord.java
ibarbashov: done

netezza-plugin/pom.xml
ibarbashov: done

netezza-plugin/docs/Netezza-batchsource.md
ibarbashov: removed

netezza-plugin/docs/Netezza-batchsource.md
ibarbashov: removed

netezza-plugin/src/main/java/co/cask/netezza/NetezzaDBRecord.java
ibarbashov: when you call getBytes() it returns byte[], but when you call getObject() on resultSet it returns only first byte as Byte (seems like a bug in netezza driver), so the default mapping is not working, and this type need to be handled  properly

netezza-plugin/src/main/java/co/cask/netezza/NetezzaDBRecord.java
ibarbashov: done

netezza-plugin/src/main/java/co/cask/netezza/NetezzaDBRecord.java
ibarbashov: replaced with resultSet.getString()

netezza-plugin/src/main/java/co/cask/netezza/NetezzaDBRecord.java
```diff
@@ -53,25 +53,22 @@ public NetezzaDBRecord() {}
...
 
     switch (sqlType) {
       case Types.VARBINARY:
-        original = resultSet.getBytes(field.getName());
-        recordBuilder.set(field.getName(), original);
+        recordBuilder.set(field.getName(), resultSet.getBytes(field.getName()));
         break;
       case INTERVAL:
-        recordBuilder.set(field.getName(), original.toString());
+        recordBuilder.set(field.getName(), resultSet.getString(field.getName()));
```
albertshau, line 71: how come this changed from the column to the whole row? My previous question was just about what object was being used here to make sure .toString() is meaningful

ibarbashov: Yeah it's meaningful, it prints interval in 'n years n mons n days hh:mm:ss' format  , also INTERVAL type mapping is covered in integration tests

albertshau (approved):
lgtm

PR #12, opened 2019-03-04
Feature/sink schema validation and inference

ibarbashov:


database-commons/src/main/java/co/cask/db/batch/sink/AbstractDBSink.java
```diff
@@ -102,22 +102,18 @@ public void prepareRun(BatchSinkContext context) {
               dbSinkConfig.jdbcPluginName,
               connectionString, "columns");
 
+    Schema inputSchema = context.getInputSchema();
+    Objects.requireNonNull(inputSchema, "AbstractDBSink requires a single known schema.");
```
albertshau: Now that we are looking up the schema, we should allow the input schema to be null. If it is null, try to use whatever schema the output table has.

database-commons/src/main/java/co/cask/db/batch/sink/AbstractDBSink.java
```diff
@@ -221,18 +218,60 @@ private boolean tableExists(Class<? extends Driver> jdbcDriverClass, String tabl
...
+      if (ResultSetMetaData.columnNullable == rsMetaData.isNullable(columnIndex)) {
+        columnSchema = Schema.nullableOf(DBUtils.getSchema(type, precision, scale));
+      } else {
+        columnSchema = DBUtils.getSchema(type, precision, scale);
+      }
+
+      // we check for null-assignment compatibility, because schema.isCompatible() doesn't
+      // handle this case correctly
+      boolean isNotNullAssignable = !columnSchema.isNullable() && field.getSchema().isNullable();
+      boolean isNotCompatible = !field.getSchema().isCompatible(columnSchema);
```
albertshau: isCompatible is used primarily for avro-encoding, whether a field that is written as one type can be read as another. I'm pretty sure the same rules don't apply for databases. 

I think we should just be checking that the schema types match here.

database-commons/src/main/java/co/cask/db/batch/sink/AbstractDBSink.java
```diff
@@ -221,18 +218,60 @@ private boolean tableExists(Class<? extends Driver> jdbcDriverClass, String tabl
...
+
+      if (isNotCompatible || isNotNullAssignable) {
+        invalidFields.add(field.getName());
+      }
+    }
+
+    Preconditions.checkArgument(invalidFields.isEmpty(), "Couldn't find matching database column(s) " +
```
albertshau, line 298: This error message won't give the full information about what went wrong here (does the column not exist, is the field nullable while the column is not nullable, do the types not match, etc). 

It's better to error log the problems as we come to them in addition to throwing this exception.

database-commons/src/main/java/co/cask/db/batch/sink/AbstractDBSink.java
ibarbashov: I'm not sure, how can Sink be used without explicit schema. I mean, it should always be connected to either source or transform, which have output schema that is passed to sink. Or am I getting something wrong?

database-commons/src/main/java/co/cask/db/batch/sink/AbstractDBSink.java
ibarbashov: Got it. Does this mean that we don't care about complex union/array types? The only union we are interested in is [NULL | SimpleType] ?

database-commons/src/main/java/co/cask/db/batch/sink/AbstractDBSink.java
ibarbashov: OK

database-commons/src/main/java/co/cask/db/batch/sink/AbstractDBSink.java
albertshau: A null schema happens if the previous stage set their output schema to be null. It indicates that the schema is either not known or that it can be multiple schemas. This is normally not the case, but we have seen instances where the previous stage might sometimes emit records with fields a,b,c and might sometimes emit records with fields a,b,d. In that scenario, it's possible they might want to connect it to a db sink where the table has columns a,b,c,d.

It's not a common scenario, but it looks like we could support that without much work.

database-commons/src/main/java/co/cask/db/batch/sink/AbstractDBSink.java
albertshau: Yeah we don't need to support union. I don't think any of these relational dbs support union anyway

ibarbashov:
WIP Do not merge, pls.

ibarbashov:
Feature is complete. Ready for re-review and merging.

database-commons/src/main/java/co/cask/db/batch/sink/AbstractDBSink.java
```diff
@@ -221,18 +242,63 @@ private boolean tableExists(Class<? extends Driver> jdbcDriverClass, String tabl
...
+
+      Schema.Type columnType = columnSchema.getType();
+      Schema.Type fieldType = field.getSchema().isNullable() ? field.getSchema().getNonNullable().getType()
+                                                             : field.getSchema().getType();
+      boolean isNotNullAssignable = !isColumnNullable && field.getSchema().isNullable();
+      boolean isNotCompatible = !Objects.equals(fieldType, columnType);
+
+      if (isNotCompatible) {
+        invalidFields.add(field.getName());
+        LOG.error("Can't assign value {} of type {} to column {} of type {}", field.getName(),
```
albertshau: "Field {} was given as type {} but the database column is actually of type {}."

database-commons/src/main/java/co/cask/db/batch/sink/AbstractDBSink.java
```diff
@@ -221,18 +242,63 @@ private boolean tableExists(Class<? extends Driver> jdbcDriverClass, String tabl
...
+                  field.getSchema().getType(), field.getName(), columnSchema.getType());
+      }
+      if (isNotNullAssignable) {
+        invalidFields.add(field.getName());
+        LOG.error("Can't assign nullable value {} to non-null column {}", field.getName(), field.getName());
```
albertshau: "Field {} was given as nullable but the database column is not nullable"

database-commons/src/main/java/co/cask/db/batch/sink/AbstractDBSink.java
```diff
@@ -210,7 +230,8 @@ private void setResultSetMetadata() throws Exception {
     }
   }
 
-  private boolean tableExists(Class<? extends Driver> jdbcDriverClass, String tableName) {
+  private void validateSchema(Class<? extends Driver> jdbcDriverClass, String tableName,
+                                 Schema inputSchema) {
```
albertshau: indentation

database-commons/src/main/java/co/cask/db/batch/sink/AbstractDBSink.java
```diff
@@ -138,10 +140,28 @@ private void setColumnsInfo(List<Schema.Field> fields) {
...
     setResultSetMetadata();
   }
 
+  private Schema inferSchema(Class<? extends Driver> driverClass) {
+    List<Schema.Field> inferredFields = new ArrayList<>();
+    try {
+      DBUtils.ensureJDBCDriverIsAvailable(driverClass, dbSinkConfig.getConnectionString(), dbSinkConfig.jdbcPluginName);
+      try (Connection connection = DriverManager.getConnection(dbSinkConfig.getConnectionString(),
+                                                               dbSinkConfig.getConnectionArguments());
+          Statement statement = connection.createStatement();
```
albertshau: indentation

albertshau (approved):
lgtm

PR #17, opened 2019-03-21
Feature/tests cleanup

ibarbashov:
- Removed JDBC drivers with incompatible licenses from dependencies.
- Replaced configuration properties keys with constants in tests.
- Removed StructuredRecordUtils and related tests as unused.


jayfjin:
LGTM. Could I get a confirmation that all tests and builds pass? I'll merge after that

ibarbashov:
@jayfjin Thanks for double check, it helped to find missing properties in tests. Now everything is good to go

PR #19, opened 2019-03-22
Improve Oracle types support

ibarbashov:


oracle-plugin/src/main/java/co/cask/oracle/OracleSchemaReader.java
```diff
@@ -0,0 +1,66 @@
...
+import java.sql.ResultSetMetaData;
+import java.sql.SQLException;
+import java.util.Set;
+
+/**
+ * Oracle schema reader.
+ */
+public class OracleSchemaReader extends CommonSchemaReader {
+
+  public static final int INTERVAL_YM = -103;
```
jayfjin, line 34: what do these constants mean?

oracle-plugin/src/main/java/co/cask/oracle/OracleSchemaReader.java
```diff
@@ -0,0 +1,66 @@
...
+    if (ORACLE_TYPES.contains(sqlType)) {
+      switch (sqlType) {
+        case INTERVAL_DS:
+        case INTERVAL_YM:
+          return Schema.of(Schema.Type.STRING);
+        case TIMESTAMP_LTZ:
+        case TIMESTAMP_TZ:
+          return Schema.of(Schema.LogicalType.TIMESTAMP_MICROS);
+        default:
+          throw new SQLException(new UnsupportedTypeException(
```
jayfjin: is this code accessable based on the if condition you have above?

oracle-plugin/src/main/java/co/cask/oracle/OracleSchemaReader.java
ibarbashov: It is Oracle type constant, from Oracle JDBC Driver Implementation, and it doesn't match standard SQLTypes constants

oracle-plugin/src/main/java/co/cask/oracle/OracleSchemaReader.java
ibarbashov: yes, you right. Removed

oracle-plugin/src/main/java/co/cask/oracle/OracleSchemaReader.java
jayfjin: I see, would be good to have a comment about this

ibarbashov: added

PR #25, opened 2019-04-03
Feature/aurora postgres integration tests

ibarbashov:


aurora-postgresql-plugin/src/test/java/co/cask/auroradb/postgres/AuroraPostgresPluginTestBase.java
```diff
@@ -0,0 +1,214 @@
...
+import java.sql.Timestamp;
+import java.util.Calendar;
+import java.util.Collections;
+import java.util.Map;
+import java.util.TimeZone;
+
+public class AuroraPostgresPluginTestBase extends DatabasePluginTestBase {
+  protected static final ArtifactId DATAPIPELINE_ARTIFACT_ID = NamespaceId.DEFAULT.artifact("data-pipeline", "3.2.0");
+  protected static final ArtifactSummary DATAPIPELINE_ARTIFACT = new ArtifactSummary("data-pipeline", "3.2.0");
+  protected static final long CURRENT_TS = System.currentTimeMillis();
```
jayfjin, line 55: seen this a few times now. Should we move to the base class?

aurora-postgresql-plugin/src/test/java/co/cask/auroradb/postgres/AuroraPostgresSourceTestRun.java
```diff
@@ -0,0 +1,325 @@
...
+                                              DATAPIPELINE_ARTIFACT, "testDBSourceWithMultipleTables");
+    runETLOnce(appManager);
+
+    // records should be written
+    DataSetManager<Table> outputManager = getDataset(outputDatasetName);
+    List<StructuredRecord> outputRecords = MockSink.readOutput(outputManager);
+    Assert.assertEquals(2, outputRecords.size());
+    String userid = outputRecords.get(0).get("NAME");
+    StructuredRecord row1 = "user1".equals(userid) ? outputRecords.get(0) : outputRecords.get(1);
+    StructuredRecord row2 = "user1".equals(userid) ? outputRecords.get(1) : outputRecords.get(0);
```
jayfjin, line 194: the rows come in a nondeterministic order?

aurora-postgresql-plugin/src/test/java/co/cask/auroradb/postgres/AuroraPostgresPluginTestBase.java
ibarbashov: I struggled a lot with setting up test framework (due to its static nature) when I was working on our very first MySQL plugin. There was a workaround with startCount in original database-plugins tests. The DATAPIPELINE_ARTIFACT was registered in base class once for the whole test suite and when other plugin's tests where trying to set up a test pipeline with their own Sources and Sinks the error occurred that DATAPIPELINE_ARTIFACT was already registered. So I had to duplicate this code and register it for every plugin's test group, while leaving only static utility methods in base tests class (I mean the class that all tests inherit from). 

On the other hand, this allows us to run tests for one or several particular modules at a time instead of triggering build for the whole database-plugins project.

aurora-postgresql-plugin/src/test/java/co/cask/auroradb/postgres/AuroraPostgresSourceTestRun.java
ibarbashov: These lines are also legacy from old plugin test. From database perspective the order is deterministic (i.e. we read rows in order we wrote them) but I'm not sure about test pipeline implementation, for example it might simulate data shuffling, etc.

aurora-postgresql-plugin/src/test/java/co/cask/auroradb/postgres/AuroraPostgresPluginTestBase.java
jayfjin: Hmm ok sg

PR #31, opened 2019-04-12
Use Sqoop library for DB plugins

sreevatsanraman:


sreevatsanraman:
Merging this. Since we tested the change

PR #32, opened 2019-04-17
CDAP-15100 Improve performance of DBRecord class

mlozbin-cybervisiontech:
JIRA: https://issues.cask.co/browse/CDAP-15100
WIKI: https://wiki.cask.co/display/CE/Database+plugin+enhancements

In scope of this PR:
- Removed input schema parsing on each incoming row
- Removed field validation on each incoming row
- Added schema validation on pipeline deployment
- Schema parsing changed to lazy computation with cache
- Actual DB schema resolving was moved to "prepareRun" stage

Attachments ([CDAP-15100_report.zip](https://github.com/data-integrations/database-plugins/files/3088404/CDAP-15100_report.zip)):
- CPU utilization flamegraphs before and after improvement
- CDAP report for 500k records read before and after improvement for same pipeline
- Test pipeline

### DIFF
--- a/database-commons/src/main/java/io/cdap/plugin/util/Lazy.java
+++ b/database-commons/src/main/java/io/cdap/plugin/util/Lazy.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright © 2019 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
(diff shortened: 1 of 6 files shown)


mlozbin-cybervisiontech:
@jayfjin could you, please, take a look

database-commons/src/main/java/io/cdap/plugin/util/Lazy.java
```diff
@@ -0,0 +1,47 @@
...
+public final class Lazy<T> {
+  private final Supplier<T> supplier;
+  private volatile T value;
+
+  public Lazy(Supplier<T> supplier) {
+    this.supplier = supplier;
+  }
+
+  public T getOrCompute() {
+    final T result = value; // Just one volatile read
```
jayfjin, line 37: does this function need to be atomic as well?

jayfjin (commented):
Great work! Just one question

database-commons/src/main/java/io/cdap/plugin/util/Lazy.java
mlozbin-cybervisiontech: do you mean synchronized? If so, then answer is no. This helper uses Double-checked locking in its best implementation.

PR #35, opened 2019-05-16
Feature/get schema improvement

ibarbashov:


database-commons/src/main/java/io/cdap/plugin/db/batch/source/AbstractDBSource.java
```diff
@@ -296,6 +278,8 @@ private String getJDBCPluginId() {
 
   protected abstract String createConnectionString(String host, Integer port, String database);
 
+  protected abstract String createConnectionString();
```
albertshau, line 279: since everything needs to implement this now, can the other version that takes host, port, database be removed?

albertshau (commented):
one question

database-commons/src/main/java/io/cdap/plugin/db/batch/source/AbstractDBSource.java
ibarbashov: Yep, doesn't need it anymore. Removed

albertshau (approved):
lgtm

albertshau:
please squash commits and i'll merge

PR #36, opened 2019-05-21
CDAP-15264 Investigate the reason of database plugins performance gap…

mlozbin-cybervisiontech:
JIRA: https://issues.cask.co/browse/CDAP-15264

### In scope of this PR
- improved performance of operations with LOB type values
- improved performance of DBRecord.handleField operation

### What was done
- removed "free()" method call for BLOB/CLOB types so it work the same way as in sqoop and Spring framework (see: [spring issue](https://github.com/spring-projects/spring-framework/issues/16042))
- reworked ResultSet.get* methods to use column index instead of column key to get rid of lookup time

### Test results for Oracle DB
- 500k rows (before patch) = 52 sec
- 500k rows (after patch) = 14 sec
- 500k rows (sqoop-libs branch) = 54 sec
- 10kk rows (before patch) = 17:57 min
- 10kk rows (after patch) = 3:05 min
- 10kk rows (sqoop-libs branch) = 15:55 min

### Attachments
- CPU and allocation flamegraphs: [oracle-flamegraphs.zip](https://github.com/data-integrations/database-plugins/files/3204627/oracle-flamegraphs.zip)

### Results explanation
**Why does plugin perform that much faster after patch?**
Comparing db plugins and sqoop CPU utilization we can see that Sqoop does not call "free()" method call for BLOB/CLOB types (see oracle_sqoop-cpu.svg from attached flamegraphs). The reason of performance improvement is shown comparing "oracle-cpu (before).svg" and "oracle-cpu (after).svg".

**What is the difference between CDAP plugin and Sqoop application?**
After applying this PR the only difference between CDAP and Sqoop logic will be in StructuredRecord creation and CDAP core related logic. See "oracle-cpu (after).svg" and "oracle_sqoop-cpu.svg".

**Why do we have old bad results after migrating to sqoop in sqoop-libs branch?**
See "oracle-cpu (after).svg" and "oracle-cdap-sqoop-cpu_flamegraph.svg".
We can not really tell that changing implemented interface from "org.apache.hadoop.mapreduce.lib.db.DBWritable" to "org.apache.sqoop.mapreduce.DBWritable" is a migration to Sqoop. The main differences between CDAP plugins and Sqoop are:
- Sqoop generates its own optimized QueryResults class to get rid of SQL schema inspection on each raw. But we still use our DBRecord class in sqoop-libs branch.
- Sqoop does not generate StructuredRecords and just copies bytes from BLOBs to output file (which is way faster)

### Conclusion
- we will achieve almost maximum possible performance after applying this PR without any migrations
- we will not have any benefits from migration to Sqoop

### DIFF
--- a/aurora-postgresql-plugin/src/main/java/io/cdap/plugin/auroradb/postgres/AuroraPostgresDBRecord.java
+++ b/aurora-postgresql-plugin/src/main/java/io/cdap/plugin/auroradb/postgres/AuroraPostgresDBRecord.java
@@ -41,17 +41,17 @@ public class AuroraPostgresDBRecord extends DBRecord {
 
   @Override
   protected void handleField(ResultSet resultSet, StructuredRecord.Builder recordBuilder, Schema.Field field,
-                             int sqlType, int sqlPrecision, int sqlScale) throws SQLException {
+                             int columnIndex, int sqlType, int sqlPrecision, int sqlScale) throws SQLException {
     if (AuroraPostgresSchemaReader.POSTGRES_TYPES.contains(sqlType)) {
-      handleSpecificType(resultSet, recordBuilder, field);
+      handleSpecificType(resultSet, recordBuilder, field, columnIndex);
     } else {
-      setField(resultSet, recordBuilder, field, sqlType, sqlPrecision, sqlScale);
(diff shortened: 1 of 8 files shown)


mlozbin-cybervisiontech:

albertshau (approved):
lgtm

PR #37, opened 2019-05-30
[CDAP-15440] Update MySQL configuration to look similar to native

vladhlinskiy:
JIRA: https://issues.cask.co/browse/CDAP-15440
WIKI: https://wiki.cask.co/display/CE/MySQL+database+plugin

In scope of this PR:

- Support of the SQL_MODE, ANSI quotes, compression and SSL configuration properties added 
- DB-specific arguments passed to `DataDrivenETLDBInputFormat`, `ETLDBOutputFormat`


vladhlinskiy:

mysql-plugin/docs/Mysql-batchsink.md
```diff
@@ -38,6 +38,26 @@ will be passed to the JDBC driver as connection arguments for JDBC drivers that
 
 **Auto Reconnect** Should the driver try to re-establish stale and/or dead connections.
 
+**Use SSL** Turns on SSL encryption. The connection will fail if SSL is not available.
+
+**Keystore URL** URL to the client certificate KeyStore (if not specified, use defaults). Must be accessible at the
+same location on host where CDAP Master is running and all hosts on which at least one HDFS, MapReduce, or YARN daemon
+role is running.
+
+**Keystore password** Password for the client certificates KeyStore.
```
albertshau: By convention, these labels should be capitalized. Keystore Password.

mysql-plugin/widgets/Mysql-batchsource.json
```diff
@@ -107,9 +107,80 @@
...
+          }
+        },
+        {
+          "widget-type": "textbox",
+          "label": "Keystore URL",
+          "name": "clientCertificateKeyStoreUrl"
+        },
+        {
+          "widget-type": "password",
+          "label": "Keystore password",
```
albertshau: By convention, these labels should be capitalized. Keystore Password

mysql-plugin/widgets/Mysql-batchsource.json
```diff
@@ -107,9 +107,80 @@
...
+            "off": {
+              "value": "false",
+              "label": "Off"
+            },
+            "default": "false"
+          }
+        },
+        {
+          "widget-type": "toggle",
+          "label": "Use ANSI quotes to quote identifiers",
```
albertshau: We try not to make these too many words. 'Use ANSI Quotes' should be enough

mysql-plugin/widgets/Mysql-batchsource.json
```diff
@@ -107,9 +107,80 @@
...
       "label": "Advanced",
       "properties": [
+        {
+          "widget-type": "toggle",
+          "label": "Use compression protocol",
+          "name": "useCompression",
+          "widget-attributes": {
+            "on": {
+              "value": "true",
+              "label": "On"
```
albertshau: If this is labeled 'Use Compression', then 'Yes' and 'No' are better labels (and more consistent with 'Use SSL' above). If it were labeled 'Compression', then 'On' and 'Off' make more sense. For consistency, let make it 'Use Compression' with 'Yes' and 'No'.

mysql-plugin/widgets/Mysql-batchsource.json
```diff
@@ -107,9 +107,80 @@
...
+    {
+      "label": "SSL",
+      "properties": [
+        {
+          "label": "Use SSL",
+          "name": "useSSL",
+          "widget-type": "select",
+          "widget-attributes": {
+            "default": "If available",
+            "values": ["No", "If available", "Require"]
```
albertshau: 'No' and 'Require' seem like an odd pair. I think it's better to have:

'If Available', 'Yes', 'No'

mysql-plugin/widgets/Mysql-batchsource.json
```diff
@@ -107,9 +107,80 @@
...
+          }
+        },
+        {
+          "widget-type": "toggle",
+          "label": "Use ANSI quotes to quote identifiers",
+          "name": "useAnsiQuotes",
+          "widget-attributes": {
+            "on": {
+              "value": "true",
+              "label": "On"
```
albertshau: similar comment about 'Yes' and 'No'

mysql-plugin/widgets/Mysql-batchsource.json
```diff
@@ -123,22 +194,19 @@
...
-                "label": "True"
-              },
-              {
-                "id": "false",
-                "label": "False"
-              }
-            ]
+            "on": {
+              "value": "true",
+              "label": "On"
```
albertshau: similar comment about 'Yes' and 'No'.

mysql-plugin/src/main/java/io/cdap/plugin/mysql/MysqlSource.java
```diff
@@ -58,18 +58,70 @@ protected String createConnectionString() {
     @Nullable
     public Boolean autoReconnect;
 
+    @Name(MysqlConstants.USE_COMPRESSION)
```
albertshau, line 63: Do all of the mysql plugins have the same properties with the same descriptions? If so, can another class be added with the common properties?

mysql-plugin/src/main/java/io/cdap/plugin/mysql/MysqlUtil.java
```diff
@@ -0,0 +1,90 @@
...
+   * @param autoReconnect                     should the driver try to re-establish stale and/or dead connections.
+   * @param useCompression                    specifies if compression must be enabled.
+   * @param useSSL                            specifies if SSL encryption must be turned on.
+   * @param clientCertificateKeyStoreUrl      URL of the client certificate KeyStore.
+   * @param clientCertificateKeyStorePassword password for the client certificates KeyStore.
+   * @param trustCertificateKeyStoreUrl       URL of the trusted root certificate KeyStore.
+   * @param trustCertificateKeyStorePassword  password for the trusted root certificates KeyStore.
+   * @return immutable map of the MySQL specific arguments
+   */
+  public static Map<String, String> composeDbSpecificArgumentsMap(Boolean autoReconnect,
```
albertshau, line 43: Is it possible to have a base MysqlConf class that has all of these properties and move this logic to that class?

database-commons/src/main/java/io/cdap/plugin/db/ConnectionConfig.java
```diff
@@ -111,17 +132,55 @@ public Properties getConnectionArguments() {
...
    */
   public abstract String getConnectionString();
 
+  /**
+   * Returns list of initialization queries as a single string separated by ';' character. Initialization queries
+   * supposed to be executed preserving order right after connection establishing.
+   * In the case when there are no initialization queries, an empty string will be returned.
+   * @return list of initialization queries as a single string separated by ';' character.
+   */
+  public String getInitQueriesString() {
```
albertshau: similar comment here. It's less error prone if this returns a List rather than an encoded string.

database-commons/src/main/java/io/cdap/plugin/db/ConnectionConfig.java
```diff
@@ -111,17 +132,55 @@ public Properties getConnectionArguments() {
     return arguments;
   }
 
+  /**
+   * Returns all configuration properties including database-specific ones as single string.
+   * In the case when there are no connection arguments, an empty string will be returned.
+   * @return connection arguments as string with '=' as key-value delimiter and ';' as pair delimiter.
+   */
+  public String getConnectionArgumentsString() {
```
albertshau: We should avoid encoding things in strings when possible. The outputformat should be the one responsible for encoding/decoding the map.

database-commons/src/main/java/io/cdap/plugin/db/ConnectionConfig.java
```diff
@@ -111,17 +132,55 @@ public Properties getConnectionArguments() {
     return arguments;
   }
 
+  /**
+   * Returns all configuration properties including database-specific ones as single string.
+   * In the case when there are no connection arguments, an empty string will be returned.
+   * @return connection arguments as string with '=' as key-value delimiter and ';' as pair delimiter.
```
albertshau: are these delimiters safe to use for every db type? You could also use Gson to serialize/deserialize and then you wouldn't have to worry about special characters.

database-commons/src/main/java/io/cdap/plugin/db/batch/source/AbstractDBSource.java
```diff
@@ -79,14 +81,14 @@ public AbstractDBSource(DBSourceConfig sourceConfig) {
...
-    if (importQueryString.contains("WHERE $CONDITIONS AND")) {
-      importQueryString = importQueryString.replace("$CONDITIONS AND", "");
-    } else if (importQueryString.contains("WHERE $CONDITIONS")) {
-      importQueryString = importQueryString.replace("WHERE $CONDITIONS", "");
-    } else if (importQueryString.contains("AND $CONDITIONS")) {
-      importQueryString = importQueryString.replace("AND $CONDITIONS", "");
-    } else if (importQueryString.contains("$CONDITIONS")) {
+    importQueryString = importQueryString.replaceAll("\\s{2,}", " ");
+    if (importQueryString.toUpperCase().contains("WHERE $CONDITIONS AND")) {
+      importQueryString = importQueryString.replaceAll("(?i)" + Pattern.quote("$CONDITIONS AND"), "");
```
albertshau, line 86: can you add a comment about what this is doing and also include an example?

vladhlinskiy: There was a bug related to queries case insensitiveness. I was not able to specify lowercase table names for MySQL Source. Thus, I removed `toUpperString` conversion for the import query and changed the following `replace` methods to ignore case, since they were expecting string in upper case. This method was added to exclude conditions clause from the query in get schema request in DbSource(https://github.com/cdapio/hydrator-plugins/pull/528).

mysql-plugin/src/main/java/io/cdap/plugin/mysql/MysqlSource.java
vladhlinskiy: With existing class hierarhy for config classes we have to extend specific parent classes for different plugin types:
1) `DBSpecificQueryActionConfig` and `DBSpecificQueryConfig` for action/post-action plugins.
2) `DBSpecificSinkConfig` for sink plugins.
3) `DBSpecificSourceConfig` for source plugins.

So, there is no way to add a single class with common MySQL properties. We could change existing hierarchy or do not use it for mysql-plugin but it will lead to duplication of common source/sink/action configs between plugins' modules. For example, we would duplicate code from `DBSourceConfig`(https://github.com/data-integrations/database-plugins/blob/develop/database-commons/src/main/java/io/cdap/plugin/db/batch/source/AbstractDBSource.java#L284).

mysql-plugin/src/main/java/io/cdap/plugin/mysql/MysqlUtil.java
vladhlinskiy: There is no straightforward way to do so due to the existing hierarchy of config classes.

vladhlinskiy:
@albertshau could you, please, review?

mysql-plugin/src/main/java/io/cdap/plugin/mysql/MysqlSource.java
albertshau: Hm I see, makes sense to have it the way you've done it then.

database-commons/src/main/java/io/cdap/plugin/db/ConnectionConfig.java
```diff
@@ -74,14 +74,39 @@ public ConnectionConfig() {
...
+   * Parses connection arguments into a {@link Map}.
    *
    * @param connectionArguments See {@link ConnectionConfig#connectionArguments}.
    * @param user                See {@link ConnectionConfig#user}.
    * @param password            See {@link ConnectionConfig#password}.
    */
-  public static Properties getConnectionArguments(@Nullable String connectionArguments,
-                                                  @Nullable String user, @Nullable String password) {
+  protected static Map<String, String> getConnectionArguments(@Nullable String connectionArguments,
+                                                     @Nullable String user, @Nullable String password) {
```
albertshau: nit: fix indentation

albertshau (approved):
one minor nit, otherwise lgtm. Please fix and squash commits, then I'll merge.

PR #43, opened 2019-06-21
CDAP-15519 Mapping changed to use decimal logical type

vladhlinskiy:
JIRA: https://issues.cask.co/browse/CDAP-15519

In scope of this PR:

* Mapping changed to use decimal logical type
* Tests changed to use decimal logical type

### DIFF
--- a/aurora-mysql-plugin/src/test/java/io/cdap/plugin/aurora/mysql/AuroraMysqlSinkTestRun.java
+++ b/aurora-mysql-plugin/src/test/java/io/cdap/plugin/aurora/mysql/AuroraMysqlSinkTestRun.java
@@ -33,6 +33,8 @@ import io.cdap.plugin.db.batch.sink.AbstractDBSink;
 import org.junit.Assert;
 import org.junit.Test;
 
+import java.math.BigDecimal;
+import java.math.MathContext;
 import java.sql.Connection;
 import java.sql.Date;
 import java.sql.ResultSet;
@@ -85,6 +87,10 @@ public class AuroraMysqlSinkTestRun extends AuroraMysqlPluginTestBase {
       Assert.assertEquals("user2", Bytes.toString(resultSet.getBytes("TINYBLOB_COL"), 0, 5));
       Assert.assertEquals("user2", Bytes.toString(resultSet.getBytes("MEDIUMBLOB_COL"), 0, 5));
       Assert.assertEquals("user2", Bytes.toString(resultSet.getBytes("LONGBLOB_COL"), 0, 5));
+      Assert.assertEquals(new BigDecimal(3.458, new MathContext(PRECISION)).setScale(SCALE),
+                          resultSet.getBigDecimal("NUMERIC_COL"));
+      Assert.assertEquals(new BigDecimal(3.459, new MathContext(PRECISION)).setScale(SCALE),
+                          resultSet.getBigDecimal("DECIMAL_COL"));
       users.add(resultSet.getString("NAME"));
       Assert.assertEquals(ImmutableSet.of("user1", "user2"), users);
     }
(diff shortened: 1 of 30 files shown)


vladhlinskiy:

aurora-mysql-plugin/src/test/java/io/cdap/plugin/aurora/mysql/AuroraMysqlSinkTestRun.java
```diff
@@ -85,6 +87,8 @@ public void testDBSink() throws Exception {
       Assert.assertEquals("user2", Bytes.toString(resultSet.getBytes("TINYBLOB_COL"), 0, 5));
       Assert.assertEquals("user2", Bytes.toString(resultSet.getBytes("MEDIUMBLOB_COL"), 0, 5));
       Assert.assertEquals("user2", Bytes.toString(resultSet.getBytes("LONGBLOB_COL"), 0, 5));
+      Assert.assertEquals(3.458, resultSet.getBigDecimal("NUMERIC_COL").doubleValue(), 0.00001);
```
albertshau: seems like you should be able to compare BigDecimal objects directly instead of double values.

database-commons/src/main/java/io/cdap/plugin/util/DBUtils.java
```diff
@@ -212,10 +213,7 @@ public static Schema getSchema(int sqlType, int precision, int scale) throws SQL
 
       case Types.NUMERIC:
       case Types.DECIMAL:
-        // if there are no digits after the point, use integer types
-        type = scale != 0 ? Schema.Type.DOUBLE :
-          // with 10 digits we can represent 2^32 and LONG is required
-          precision > 9 ? Schema.Type.LONG : Schema.Type.INT;
+        type = Schema.Type.BYTES;
```
albertshau: how come this isn't returning a decimal?

aurora-mysql-plugin/src/test/java/io/cdap/plugin/aurora/mysql/AuroraMysqlSinkTestRun.java
albertshau: similar comment for rest of the tests.

vladhlinskiy: Changed to compare `BigDecimal` objects directly.

database-commons/src/main/java/io/cdap/plugin/util/DBUtils.java
vladhlinskiy: It certainly must return `Schema.decimalOf(precision, scale)` just like in `CommonSchemaReader#getSchema`, thanks. 
Fixed.

albertshau: does this mean this code path wasn't getting tested in the unit tests?

albertshau (approved):
lgtm. squash and we'll merge.

database-commons/src/main/java/io/cdap/plugin/util/DBUtils.java
vladhlinskiy: Decimal logical type is based on the `Schema.Type.BYTES` primitive type: 
https://github.com/cdapio/cdap/blob/develop/cdap-api-common/src/main/java/io/cdap/cdap/api/data/schema/Schema.java#L124
```
DECIMAL(Schema.Type.BYTES, "decimal");
```

Thus it's ok to use `Schema.Type.BYTES` in that case and unit tests didn't help to find this.

vladhlinskiy: Also we have to ensure that we test sink plugins with explicit input schema set. I will add these tests in the following PRs for all data types support.

PR #45, opened 2019-07-04
CDAP-15549 MS SQL db plugin enhancements: all data types support and proper test coverage

vladhlinskiy:
JIRA: https://issues.cask.co/browse/CDAP-15549
WIKI: https://wiki.cask.co/display/CE/Microsoft+SQL+Server+database+plugin

**In scope of this PR:**

* Added tests for Source plugin for next types: 
```
DATETIME, DATETIMEOFFSET, IMAGE, MONEY, NCHAR, NTEXT, NVARCHAR, NVARCHAR(MAX), SMALLDATETIME, SMALLMONEY, TIMESTAMP, UDT, UNIQUEIDENTIFIER, VARBINARY(MAX), VARCHAR(MAX), XML, SQL_VARIANT, GEOMETRY, GEOGRAPHY
```

* Added tests for Sink plugin for the next types: 
```
BIGINT, BIT, CHAR, DATETIME, DATETIMEOFFSET, DECIMAL, FLOAT, IMAGE, INT, MONEY, NCHAR, NTEXT, NUMERIC, NVARCHAR, NVARCHAR(MAX), REAL, SMALLDATETIME, SMALLINT, SMALLMONEY, TEXT, TINYINT, UDT, UNIQUEIDENTIFIER, VARBINARY, VARBINARY(MAX), VARCHAR(MAX), XML, SQL_VARIANT, GEOMETRY, GEOGRAPHY
```

* Added support of `DATETIMEOFFSET` data type, currently mapped to the `Type.STRING`

* Added support of `TIMESTAMP` type for Source plugin

`TIMESTAMP` is the synonym for the `ROWVERSION` data type, values of which are automatically generated. Thus `TIMESTAMP` can not be supported by Sink plugin.

* Added support of the `GEOMETRY` and `GEOGRAPHY` types, mapped to the `Type.BYTES`

Values of these types can be [set from Well Known Text](https://docs.microsoft.com/en-us/sql/t-sql/spatial-geometry/stgeomfromtext-geometry-data-type?view=sql-server-2017). Added support for Well Known Text in Sink plugin.

* Mappings added to the reference documentation

**Open questions:**

* `DATETIMEOFFSET` mapping

`DATETIMEOFFSET` data type, currently mapped to the `Type.STRING` in order to avoid modification of the original time zone. 
This can be changed as soon as the following scenario, [described by](https://wiki.cask.co/display/CE/Oracle+database+plugin?focusedCommentId=17990518#comment-17990518) @CuriousVini  is supported:
>  ... one approach is to store timezone information as an optional string along with epoch timestamp in platform. When reading from StrucutredRecord, getTimestamp(fieldName) method can read timestamp with provided timezone (Default timezone can be UTC). 

* `TIME` data type accuracy

MS SQL Server `TIME` data type currently mapped to the `LogicalType.TIME_MICROS`.
`TIME` data type [has the accuracy of 100 nanoseconds](https://docs.microsoft.com/en-us/sql/t-sql/data-types/time-transact-sql?view=sql-server-2017) which can not be supported using 'TIME_MICROS' logical type. This, current implementation will loose 0.1-microsecond accuracy. 

We can avoid this by mapping to the `Schema.Type.STRING`, but seems that this is not the best solution.

### DIFF
--- a/database-commons/src/main/java/io/cdap/plugin/db/CommonSchemaReader.java
+++ b/database-commons/src/main/java/io/cdap/plugin/db/CommonSchemaReader.java
@@ -37,6 +37,9 @@ public class CommonSchemaReader implements SchemaReader {
     ResultSetMetaData metadata = resultSet.getMetaData();
     // ResultSetMetadata columns are numbered starting with 1
     for (int i = 1; i <= metadata.getColumnCount(); i++) {
+      if (shouldIgnoreColumn(metadata, i)) {
+        continue;
+      }
       String columnName = metadata.getColumnName(i);
       Schema columnSchema = getSchema(metadata, i);
       if (ResultSetMetaData.columnNullable == metadata.isNullable(i)) {
@@ -119,5 +122,10 @@ public class CommonSchemaReader implements SchemaReader {
     }
 
     return Schema.of(type);
+  }
+
+  @Override
+  public boolean shouldIgnoreColumn(ResultSetMetaData metadata, int index) throws SQLException {
+    return false;
   }
 }
--- a/database-commons/src/main/java/io/cdap/plugin/db/SchemaReader.java
+++ b/database-commons/src/main/java/io/cdap/plugin/db/SchemaReader.java
@@ -22,7 +22,6 @@ import java.sql.ResultSet;
 import java.sql.ResultSetMetaData;
 import java.sql.SQLException;
 import java.util.List;
-import javax.annotation.Nullable;
 
 /**
  * Main Interface to read db specific types.
@@ -49,4 +48,20 @@ public interface SchemaReader {
    * @throws SQLException
    */
   Schema getSchema(ResultSetMetaData metadata, int index) throws SQLException;
+
+  /**
+   * Given a sql metadata indicates if sql column must be ignored and not included in the output schema. Thus we can
+   * support different data types for Sink and Source plugins.
+   * <p/>
+   * For example, in MS SQL 'TIMESTAMP' is the synonym for the 'ROWVERSION' data type, values of which are
+   * automatically generated and can not be inserted or updated. Therefore 'TIMESTAMP' can not be supported by Sink
+   * plugin. 'TIMESTAMP' reported as non-nullable column by JDBC connector, so inferred schema will be also
+   * non-nullable for this field, requiring to set a value. By ignoring this column we will avoid schema validation
+   * failure.
+   * @param metadata resultSet metadata
+   * @param index sql column index
+   * @return 'true' if sql column must not included in the output schema, 'false' otherwise
+   * @throws SQLException
+   */
+  boolean shouldIgnoreColumn(ResultSetMetaData metadata, int index) throws SQLException;
 }
--- a/database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java
+++ b/database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java
@@ -322,44 +322,47 @@ public abstract class AbstractDBSink extends ReferenceBatchSink<StructuredRecord
 
     Preconditions.checkNotNull(inputSchema.getFields());
     Set<String> invalidFields = new HashSet<>();
-
     for (Schema.Field field : inputSchema.getFields()) {
       int columnIndex = rs.findColumn(field.getName());
-      int type = rsMetaData.getColumnType(columnIndex);
-      int precision = rsMetaData.getPrecision(columnIndex);
-      int scale = rsMetaData.getScale(columnIndex);
-
-      Schema columnSchema = DBUtils.getSchema(type, precision, scale);
-      boolean isColumnNullable = (ResultSetMetaData.columnNullable == rsMetaData.isNullable(columnIndex));
-
-      Schema.Type columnType = columnSchema.getType();
-      Schema.LogicalType columnLogicalType = columnSchema.getLogicalType();
-
-      Schema fieldSchema = field.getSchema().isNullable() ? field.getSchema().getNonNullable() : field.getSchema();
-      Schema.Type fieldType = fieldSchema.getType();
-      Schema.LogicalType fieldLogicalType = fieldSchema.getLogicalType();
-
-      boolean isNotNullAssignable = !isColumnNullable && field.getSchema().isNullable();
-      boolean isNotCompatible = !(Objects.equals(fieldType, columnType)
-        && Objects.equals(fieldLogicalType, columnLogicalType));
-
-      if (isNotCompatible) {
+      if (!isFieldCompatible(field, rsMetaData, columnIndex)) {
         invalidFields.add(field.getName());
-        LOG.error("Field {} was given as type {} but the database column is actually of type {}.",
-                  field.getName(),
-                  fieldLogicalType != null ? fieldLogicalType.getToken() : fieldType,
-                  columnLogicalType != null ? columnLogicalType.getToken() : columnType
-        );
-      }
-      if (isNotNullAssignable) {
-        invalidFields.add(field.getName());
-        LOG.error("Field {} was given as nullable but the database column is not nullable", field.getName());
       }
     }
 
     Preconditions.checkArgument(invalidFields.isEmpty(),
-                                "Couldn't find matching database column(s) for input field(s) %s.",
+                                "Couldn't find matching database column(s) for input field(s) '%s'.",
                                 String.join(",", invalidFields));
+  }
+
+  /**
+   * Checks if field of the input schema is compatible with corresponding database column.
+   * @param field field of the explicit input schema.
+   * @param metadata resultSet metadata.
+   * @param index sql column index.
+   * @return 'true' if field is compatible, 'false' otherwise.
+   */
+  protected boolean isFieldCompatible(Schema.Field field, ResultSetMetaData metadata, int index) throws SQLException {
+    boolean isColumnNullable = (ResultSetMetaData.columnNullable == metadata.isNullable(index));
+    boolean isNotNullAssignable = !isColumnNullable && field.getSchema().isNullable();
+    if (isNotNullAssignable) {
+      LOG.error("Field '{}' was given as nullable but the database column is not nullable", field.getName());
+      return false;
+    }
+
+    int type = metadata.getColumnType(index);
+    int precision = metadata.getPrecision(index);
+    int scale = metadata.getScale(index);
+
+    Schema inputFieldSchema = field.getSchema().isNullable() ? field.getSchema().getNonNullable() : field.getSchema();
+    Schema outputFieldSchema = DBUtils.getSchema(type, precision, scale);
+    if (!Objects.equals(inputFieldSchema.getType(), outputFieldSchema.getType()) ||
+      !Objects.equals(inputFieldSchema.getLogicalType(), outputFieldSchema.getLogicalType())) {
+      LOG.error("Field '{}' was given as type '{}' but the database column is actually of type '{}'.",
+                field.getName(), inputFieldSchema.getType(), outputFieldSchema.getType());
+      return false;
+    }
+
+    return true;
   }
 
   private void emitLineage(BatchSinkContext context, List<Schema.Field> fields) {
--- a/database-commons/src/main/java/io/cdap/plugin/util/DBUtils.java
+++ b/database-commons/src/main/java/io/cdap/plugin/util/DBUtils.java
@@ -34,7 +34,6 @@ import java.lang.management.ManagementFactory;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.math.BigDecimal;
-import java.math.MathContext;
 import java.sql.Blob;
 import java.sql.Clob;
 import java.sql.Driver;
@@ -258,8 +257,7 @@ public final class DBUtils {
           return ((Number) original).intValue();
         case Types.NUMERIC:
         case Types.DECIMAL:
-          BigDecimal decimal = (BigDecimal) original;
-          return new BigDecimal(decimal.unscaledValue(), scale, new MathContext(precision));
+          return (BigDecimal) original;
         case Types.DATE:
           return resultSet.getDate(columnIndex);
         case Types.TIME:
--- a/database-commons/src/test/java/io/cdap/plugin/db/CustomAssertions.java
+++ b/database-commons/src/test/java/io/cdap/plugin/db/CustomAssertions.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright © 2019 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.cdap.plugin.db;
+
+import org.junit.Assert;
+
+/**
+ * Test util methods for custom assertions.
+ */
+public final class CustomAssertions {
+
+  /**
+   * The maximum delta between expected and actual floating point number for which both numbers are still considered
+   * equal.
+   */
+  public static final double DELTA = 0.000001;
+
+  private CustomAssertions() {
+    throw new AssertionError("Should not instantiate static utility class.");
+  }
+
+  /**
+   * Reuses {@link Assert#assertEquals(Object, Object)}. Added to prevent 'Ambiguous method call' issue.
+   * Asserts that two objects are equal. If they are not, an {@link AssertionError} without a message is thrown.
+   * If expected and actual are null, they are considered equal.
+   *
+   * @param expected expected value
+   * @param actual   the value to check against expected
+   */
+  public static void assertObjectEquals(Object expected, Object actual) {
+    Assert.assertEquals(expected, actual);
+  }
+
+  /**
+   * Reuses {@link Assert#assertEquals(double, double, double)} with default {@link CustomAssertions#DELTA}.
+   * Added to prevent repetitive casts to 'double' and specifying delta. Asserts that two doubles are equal to within
+   * the delta. If they are not, an AssertionError is thrown with the given message.
+   *
+   * @param expected expected value
+   * @param actual   the value to check against expected
+   */
+  public static void assertNumericEquals(double expected, double actual) {
+    Assert.assertEquals(expected, actual, 0.01);
+  }
+
+  /**
+   * Reuses {@link Assert#assertEquals(double, double, double)} with default {@link CustomAssertions#DELTA}.
+   * Added to prevent repetitive casts to 'float' and specifying delta. Asserts that two doubles are equal to within
+   * the delta. If they are not, an AssertionError is thrown with the given message.
+   *
+   * @param expected expected value
+   * @param actual   the value to check against expected
+   */
+  public static void assertNumericEquals(float expected, float actual) {
+    Assert.assertEquals(expected, actual, 0.01);
+  }
+}
--- a/mssql-plugin/docs/SqlServer-batchsink.md
+++ b/mssql-plugin/docs/SqlServer-batchsink.md
@@ -75,6 +75,58 @@ https://docs.microsoft.com/en-us/sql/relational-databases/system-compatibility-v
 **Connection Arguments:** A list of arbitrary string key/value pairs as connection arguments. These arguments
 will be passed to the JDBC driver as connection arguments for JDBC drivers that may need additional configurations.
 
+Data Types Mapping
+----------
+
+
+    | MS SQL Data Type | CDAP Schema Data Type  | Comment                                                        |
+    | ---------------- | ---------------------- | -------------------------------------------------------------- |
+    | BIGINT           | long                   |                                                                |
+    | BINARY           | bytes                  |                                                                |
+    | BIT              | boolean                |                                                                |
+    | CHAR             | string                 |                                                                |
+    | DATE             | date                   |                                                                |
+    | DATETIME         | timestamp              |                                                                |
+    | DATETIME2        | timestamp              |                                                                |
+    | DATETIMEOFFSET   | string                 |                                                                |
+    | DECIMAL          | decimal                |                                                                |
+    | FLOAT            | double                 |                                                                |
+    | IMAGE            | bytes                  |                                                                |
+    | INT              | int                    |                                                                |
+    | MONEY            | decimal                |                                                                |
+    | NCHAR            | string                 |                                                                |
+    | NTEXT            | string                 |                                                                |
+    | NUMERIC          | decimal                |                                                                |
+    | NVARCHAR         | string                 |                                                                |
+    | NVARCHAR(MAX)    | string                 |                                                                |
+    | REAL             | float                  |                                                                |
+    | SMALLDATETIME    | timestamp              |                                                                |
+    | SMALLINT         | int                    |                                                                |
+    | SMALLMONEY       | decimal                |                                                                |
+    | TEXT             | string                 |                                                                |
+    | TIME             | time                   | TIME data type has the accuracy of 100 nanoseconds which is    |
+    |                  |                        | not currently supported. Values of this type will be rounded   |
+    |                  |                        | to microsecond.                                                |
+    | TINYINT          | int                    |                                                                |
+    | UDT              | bytes                  | UDT types are mapped according to the type they are an alias   |
+    |                  |                        | of. For example, is there is an 'SSN' type that was created as |
+    |                  |                        | 'CREATE TYPE SSN FROM varchar(11);', that type would get       |
+    |                  |                        | mapped to a CDAP string. Common Language Runtime UDTs are      |
+    |                  |                        | mapped to CDAP bytes.                                          |
+    | UNIQUEIDENTIFIER | string                 |                                                                |
+    | VARBINARY        | bytes                  |                                                                |
+    | VARBINARY(MAX)   | bytes                  |                                                                |
+    | VARCHAR          | string                 |                                                                |
+    | VARCHAR(MAX)     | string                 |                                                                |
+    | XML              | string                 |                                                                |
+    | SQLVARIANT       | string                 |                                                                |
+    | GEOMETRY         | bytes                  |                                                                |
+    | GEOGRAPHY        | bytes                  |                                                                |
+    | GEOMETRY         | string                 | Values of this type can be set from Well Known Text strings,   |
+    |                  |                        | such as "POINT(3 40 5 6)".                                     |
+    | GEOGRAPHY        | string                 | Values of this type can be set from Well Known Text strings,   |
+    |                  |                        | such as "POINT(3 40 5 6)".                                     |
+
 
 Example
 -------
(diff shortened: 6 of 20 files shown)


database-commons/src/main/java/io/cdap/plugin/db/SchemaReader.java
```diff
@@ -49,4 +48,22 @@
...
   Schema getSchema(ResultSetMetaData metadata, int index) throws SQLException;
+
+  /**
+   * Given a sql metadata indicates if sql column must be ignored and not included in the output schema
+   * @param metadata resultSet metadata
+   * @param index sql column index
+   * @return 'true' if sql column must not included in the output schema, 'false' otherwise
+   * @throws SQLException
+   */
+  boolean ignoreColumn(ResultSetMetaData metadata, int index) throws SQLException;
```
vladhlinskiy: Allows to not include a column in the inferred schema. Thus we can support different data types for Sink and Source plugins.

For example, in MS SQL `TIMESTAMP` is the synonym for the `ROWVERSION` data type, values of which are automatically generated and can not be inserted or updated. Therefore `TIMESTAMP` can not be supported by Sink plugin.

`TIMESTAMP` reported as non-nullable column by JDBC connector, so inferred schema will be also non-nullable for this field, requiring to set a value. By ignoring this column we will avoid schema validation failure.

vladhlinskiy:

database-commons/src/main/java/io/cdap/plugin/db/CommonSchemaReader.java
```diff
@@ -37,13 +39,15 @@
     ResultSetMetaData metadata = resultSet.getMetaData();
     // ResultSetMetadata columns are numbered starting with 1
     for (int i = 1; i <= metadata.getColumnCount(); i++) {
-      String columnName = metadata.getColumnName(i);
-      Schema columnSchema = getSchema(metadata, i);
-      if (ResultSetMetaData.columnNullable == metadata.isNullable(i)) {
-        columnSchema = Schema.nullableOf(columnSchema);
+      if (!ignoreColumn(metadata, i)) {
```
albertshau: nit: it's a little more readable if it is:

```
if (ignoreColumn(metadata, i)) {
  continue;
}
...
```

as that removes one level of nesting and has a shorter block within the condition.

database-commons/src/main/java/io/cdap/plugin/db/CommonSchemaReader.java
```diff
@@ -120,4 +124,22 @@ public Schema getSchema(ResultSetMetaData metadata, int index) throws SQLExcepti
 
     return Schema.of(type);
   }
+
+  @Override
+  public boolean ignoreColumn(ResultSetMetaData metadata, int index) throws SQLException {
```
albertshau: nit: rename to shouldIgnoreColumn()

database-commons/src/main/java/io/cdap/plugin/db/SchemaReader.java
albertshau: thanks for this comment. Can you add it to the javadoc as well?

database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java
```diff
@@ -319,32 +319,22 @@ private void validateFields(Schema inputSchema, ResultSet rs) throws SQLExceptio
...
       boolean isNotNullAssignable = !isColumnNullable && field.getSchema().isNullable();
-      boolean isNotCompatible = !Objects.equals(fieldType, columnType);
 
-      if (isNotCompatible) {
+      if (!getSchemaReader().isTypeCompatible(field, rsMetaData, columnIndex)) {
         invalidFields.add(field.getName());
-        LOG.error("Field {} was given as type {} but the database column is actually of type {}.", field.getName(),
-                  fieldType, columnSchema.getType());
+        LOG.error("Type of the field '{}' of the given input schema is not compatible with database column at " +
+                    "index: '{}'.", field, columnIndex);
```
albertshau: this message is not as nice as the previous one. The type is much more useful than the index, as it tells the user what needs to be changed.

database-commons/src/main/java/io/cdap/plugin/util/DBUtils.java
```diff
@@ -258,8 +257,7 @@ public static Object transformValue(int sqlType, int precision, int scale,
           return ((Number) original).intValue();
         case Types.NUMERIC:
         case Types.DECIMAL:
-          BigDecimal decimal = (BigDecimal) original;
-          return new BigDecimal(decimal.unscaledValue(), scale, new MathContext(precision));
```
albertshau, line 262: how come this is removed? Is the precision/scale guaranteed to be the same as what is in the schema?

database-commons/src/test/java/io/cdap/plugin/db/CustomAssertions.java
```diff
@@ -0,0 +1,92 @@
...
+  /**
+   * Reuses {@link Assert#assertArrayEquals(byte[], byte[])}.
+   * Asserts that values of two byte arrays are equal for the length of the expected one, since it's common that actual
+   * arrays are larger. If values are not equal size of actual array is less than size of expected,
+   * an {@link AssertionError} is thrown.
+   *
+   * @param expected byte array with expected values.
+   * @param actual   byte array with actual values
+   */
+  public static void assertBytesEquals(byte[] expected, byte[] actual) {
```
albertshau: this method looks like it behaves the same as Arrays.equals()

mssql-plugin/docs/SqlServer-batchsink.md
```diff
@@ -75,6 +75,49 @@ https://docs.microsoft.com/en-us/sql/relational-databases/system-compatibility-v
...
+	| NTEXT            | Type.STRING                  |                        |
+	| NUMERIC          | LogicalType.DECIMAL          |                        |
+	| NVARCHAR         | Type.STRING                  |                        |
+	| NVARCHAR(MAX)    | Type.STRING                  |                        |
+	| REAL             | Type.FLOAT                   |                        |
+	| SMALLDATETIME    | LogicalType.TIMESTAMP_MICROS |                        |
+	| SMALLINT         | Type.INT                     |                        |
+	| SMALLMONEY       | LogicalType.DECIMAL          |                        |
+	| TEXT             | Type.STRING                  |                        |
+	| TIME             | LogicalType.TIMESTAMP_MICROS | Microsecond precision  |
```
albertshau: this isn't TIME_MICROS?

mssql-plugin/docs/SqlServer-batchsink.md
```diff
@@ -75,6 +75,49 @@ https://docs.microsoft.com/en-us/sql/relational-databases/system-compatibility-v
 **Connection Arguments:** A list of arbitrary string key/value pairs as connection arguments. These arguments
 will be passed to the JDBC driver as connection arguments for JDBC drivers that may need additional configurations.
 
+Data Types Mapping
+----------
+
+	| MS SQL Data Type | CDAP Schema Data Type        | Comment                |
+	| ---------------- | ---------------------------- | ---------------------- |
+	| BIGINT           | Type.LONG                    |                        |
```
albertshau: The reference doc is targeted towards UI users, who will not know what Type.XYZ or LogicalType.ABC means. They will only understand the types they can see in the UI. So these should match what is used in the UI, 'long', 'bytes', 'timestamp', 'time', etc.

mssql-plugin/docs/SqlServer-batchsink.md
```diff
@@ -75,6 +75,49 @@ https://docs.microsoft.com/en-us/sql/relational-databases/system-compatibility-v
...
+	| BINARY           | Type.BYTES                   |                        |
+	| BIT              | Type.BOOLEAN                 |                        |
+	| CHAR             | Type.STRING                  |                        |
+	| DATE             | LogicalType.DATE             |                        |
+	| DATETIME         | LogicalType.TIMESTAMP_MICROS |                        |
+	| DATETIME2        | LogicalType.TIMESTAMP_MICROS |                        |
```
albertshau: if there is a loss of precision in any of these types, it should be mentioned in the comment column.

mssql-plugin/docs/SqlServer-batchsink.md
```diff
@@ -75,6 +75,49 @@ https://docs.microsoft.com/en-us/sql/relational-databases/system-compatibility-v
...
+	| NVARCHAR         | Type.STRING                  |                        |
+	| NVARCHAR(MAX)    | Type.STRING                  |                        |
+	| REAL             | Type.FLOAT                   |                        |
+	| SMALLDATETIME    | LogicalType.TIMESTAMP_MICROS |                        |
+	| SMALLINT         | Type.INT                     |                        |
+	| SMALLMONEY       | LogicalType.DECIMAL          |                        |
+	| TEXT             | Type.STRING                  |                        |
+	| TIME             | LogicalType.TIMESTAMP_MICROS | Microsecond precision  |
+	| TINYINT          | Type.INT                     |                        |
+	| UDT              | Type.BYTES                   | Depends on basic type  |
```
albertshau: what does the comment mean?

mssql-plugin/docs/SqlServer-batchsink.md
```diff
@@ -75,6 +75,49 @@ https://docs.microsoft.com/en-us/sql/relational-databases/system-compatibility-v
...
+	| UNIQUEIDENTIFIER | Type.STRING                  |                        |
+	| VARBINARY        | Type.BYTES                   |                        |
+	| VARBINARY(MAX)   | Type.BYTES                   |                        |
+	| VARCHAR          | Type.STRING                  |                        |
+	| VARCHAR(MAX)     | Type.STRING                  |                        |
+	| XML              | Type.STRING                  |                        |
+	| SQLVARIANT       | Type.STRING                  |                        |
+	| GEOMETRY         | Type.BYTES                   |                        |
+	| GEOGRAPHY        | Type.BYTES                   |                        |
+	| GEOMETRY         | Type.STRING                  | Well Known Text String |
```
albertshau: What format does the string take?

mssql-plugin/src/main/java/io/cdap/plugin/mssql/SqlServerDBRecord.java
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  public SqlServerDBRecord() {}
+
+  @Override
+  protected void handleField(ResultSet resultSet, StructuredRecord.Builder recordBuilder, Schema.Field field,
+                             int columnIndex, int sqlType, int sqlPrecision, int sqlScale) throws SQLException {
+    switch (sqlType) {
+      case Types.TIME:
+        // Handle reading SQL Server 'TIME' data type to avoid accuracy loss.
```
albertshau: nice comment

mssql-plugin/src/main/java/io/cdap/plugin/mssql/SqlServerDBRecord.java
```diff
@@ -0,0 +1,135 @@
...
+          if (fieldValue == null) {
+            // Handle setting GEOGRAPHY and GEOMETRY 'null' values. Using 'stmt.setNull(sqlIndex, GEOMETRY_TYPE)' leads
+            // to com.microsoft.sqlserver.jdbc.SQLServerException: The conversion from OBJECT to GEOMETRY is unsupported
+            stmt.setString(sqlIndex, "Null");
+          } else if (fieldValue instanceof String) {
+            // Handle setting GEOGRAPHY and GEOMETRY values from Well Known Text.
+            // For example, "POINT(3 40 5 6)"
+            stmt.setString(sqlIndex, (String) fieldValue);
+          } else {
+            super.writeToDB(stmt, field, fieldIndex);
```
albertshau: will this fail?

mssql-plugin/src/main/java/io/cdap/plugin/mssql/SqlServerDBRecord.java
```diff
@@ -0,0 +1,135 @@
...
+   * SQL Server Source implementation of {@link org.apache.hadoop.mapreduce.lib.db.DBWritable} and
+   * {@link org.apache.hadoop.io.Writable}.
+   */
+  public static class Source extends SqlServerDBRecord {
+
+    public Source(StructuredRecord record, int[] columnTypes) {
+      super(record, columnTypes);
+    }
+
+    public Source() {}
```
albertshau: Does this exist because it's required for Hadoop serialization/deserialization?
```
public Source() {
  // comment about why this exists
}
```

mssql-plugin/src/main/java/io/cdap/plugin/mssql/SqlServerSchemaReader.java
```diff
@@ -0,0 +1,87 @@
...
+    if (GEOMETRY_TYPE == columnSqlType || GEOGRAPHY_TYPE == columnSqlType) {
+      return Schema.of(Schema.Type.BYTES);
+    }
+    return super.getSchema(metadata, index);
+  }
+
+  /**
+   * SQL Server Source schema reader.
+   */
+  public static class Source extends SqlServerSchemaReader {
```
albertshau: can just make SqlServerSchemaReader non-abstract rather than having an empty concrete class.

mssql-plugin/src/main/java/io/cdap/plugin/mssql/SqlServerDBRecord.java
```diff
@@ -0,0 +1,135 @@
...
+      default:
+        setField(resultSet, recordBuilder, field, columnIndex, sqlType, sqlPrecision, sqlScale);
+    }
+  }
+
+  /**
+   * SQL Server Source implementation of {@link org.apache.hadoop.mapreduce.lib.db.DBWritable} and
+   * {@link org.apache.hadoop.io.Writable}.
+   */
+  public static class Source extends SqlServerDBRecord {
```
albertshau: This doesn't look like it behaves any different than SqlServerDBRecord, can just make the parent non-abstract.

mssql-plugin/src/main/java/io/cdap/plugin/mssql/SqlServerDBRecord.java
```diff
@@ -0,0 +1,135 @@
...
+    protected SchemaReader getSchemaReader() {
+      return new SqlServerSchemaReader.Source();
+    }
+  }
+
+  /**
+   * SQL Server Sink implementation of {@link org.apache.hadoop.mapreduce.lib.db.DBWritable} and
+   * {@link org.apache.hadoop.io.Writable}.
+   */
+  public static class Sink extends SqlServerDBRecord {
```
albertshau: think it's better to move this to its own top-level class rather than an inner class.

mssql-plugin/src/main/java/io/cdap/plugin/mssql/SqlServerSchemaReader.java
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  /**
+   * SQL Server Sink schema reader.
+   */
+  public static class Sink extends SqlServerSchemaReader {
```
albertshau: this should be moved out into its own top-level class if we make SqlServerSchemaReader non-abstract.

mysql-plugin/src/test/java/io/cdap/plugin/mysql/MysqlSinkTestRun.java
```diff
@@ -40,9 +40,12 @@
 import java.sql.Timestamp;
 import java.time.LocalDateTime;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.List;
 
+import static io.cdap.plugin.db.CustomAssertions.assertBytesEquals;
```
albertshau: style nit: we don't use static method imports so that it is more clear when a static call is being made. Method calls should be changed to CustomAssertions.assertBytesEquals().

netezza-plugin/src/test/java/io/cdap/plugin/netezza/NetezzaPluginTestBase.java
```diff
@@ -51,6 +50,8 @@
 import java.util.Map;
 import java.util.TimeZone;
 
+import static java.math.MathContext.DECIMAL64;
```
albertshau: similar comment, avoid static variable imports.

database-commons/src/test/java/io/cdap/plugin/db/CustomAssertions.java
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * Test util methods for custom assertions.
+ */
+public final class CustomAssertions {
+
+  /**
+   * The maximum delta between expected and actual floating point number for which both numbers are still considered
+   * equal.
+   */
+  public static final double DELTA = 0.01;
```
albertshau: this seems like a pretty large delta. The original calls look like they were much smaller.

database-commons/src/main/java/io/cdap/plugin/db/SchemaReader.java
vladhlinskiy: Added to the javadoc.

database-commons/src/main/java/io/cdap/plugin/db/CommonSchemaReader.java
vladhlinskiy: Thanks. Changed.

database-commons/src/main/java/io/cdap/plugin/db/CommonSchemaReader.java
vladhlinskiy: Renamed.

database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java
vladhlinskiy: Changed to use the previous message. Also, [moved logic](https://github.com/data-integrations/database-plugins/pull/45/files#diff-5570db2ea5dd53e68f60ef41d9c8f6e8R338) from `CommonSchemaReader#isTypeCompatible` method to `AbstractDBSink#isFieldCompatible` so now we are able to access(and log) the actual type.

database-commons/src/main/java/io/cdap/plugin/util/DBUtils.java
vladhlinskiy: These are SQL `scale`/`precision` and `original` instance will have the same scale/precision values in most cases(except Oracle, but this is handled in the second PR).

database-commons/src/test/java/io/cdap/plugin/db/CustomAssertions.java
vladhlinskiy: It asserts that the values of the expected array are equal to the corresponding values of the actual one since it's common that actual arrays returned by `ResultSet#getBytes` are larger. `Arrays.equals()` will return `false` in such cases since it compares the lengths.

mssql-plugin/docs/SqlServer-batchsink.md
vladhlinskiy: Yes, this is actually `TIME_MICROS`, copied accidentally. Changed to the `time`, thanks.

mssql-plugin/docs/SqlServer-batchsink.md
vladhlinskiy: Changed.

mssql-plugin/docs/SqlServer-batchsink.md
vladhlinskiy: Added the following comment:
```
...
| TIME             | time                   | TIME data type has the accuracy of 100 nanoseconds which is not       |
|                  |                        | currently supported. Values of this type will be rounded to           |
|                  |                        | microsecond.                                                          |
...
```

mssql-plugin/docs/SqlServer-batchsink.md
vladhlinskiy: UDT will be mapped as basic type if it's an alias of this type. Added the following comment:
```
...
| UDT              | bytes                  | UDT mapped as basic type if it's an alias of this type. 'SSN'  |
|                  |                        | type created as 'CREATE TYPE SSN FROM varchar(11);' is mapped  |
|                  |                        | to 'string'. Common Language Runtime UDTs mapped to bytes.     |
...
```

mssql-plugin/docs/SqlServer-batchsink.md
vladhlinskiy: Added the following comment:
```
...
| GEOGRAPHY        | string                 | Values of this type can be set from Well Known Text strings,   |
|                  |                        | such as "POINT(3 40 5 6)".                                     |
...
```

mssql-plugin/src/main/java/io/cdap/plugin/mssql/SqlServerDBRecord.java
vladhlinskiy: Should i change anything?

mssql-plugin/src/main/java/io/cdap/plugin/mssql/SqlServerDBRecord.java
vladhlinskiy: It depends on the field's value. There is one more valid case, when GEOGRAPHY/GEOMETRY value is bytes. Changed to:
```
      case SqlServerSourceSchemaReader.GEOGRAPHY_TYPE:
      case SqlServerSourceSchemaReader.GEOMETRY_TYPE:
        if (fieldValue == null) {
          // Handle setting GEOGRAPHY and GEOMETRY 'null' values. Using 'stmt.setNull(sqlIndex, GEOMETRY_TYPE)' leads
          // to com.microsoft.sqlserver.jdbc.SQLServerException: The conversion from OBJECT to GEOMETRY is unsupported
          stmt.setString(sqlIndex, "Null");
        } else if (fieldValue instanceof String) {
          // Handle setting GEOGRAPHY and GEOMETRY values from Well Known Text.
          // For example, "POINT(3 40 5 6)"
          stmt.setString(sqlIndex, (String) fieldValue);
        } else {
          super.writeBytes(stmt, fieldIndex, sqlIndex, fieldValue);
        }
        break;
```

mssql-plugin/src/main/java/io/cdap/plugin/mssql/SqlServerDBRecord.java
vladhlinskiy: It's required by Hadoop `DBRecordReader` to create an instance via reflection. Comment added.

mssql-plugin/src/main/java/io/cdap/plugin/mssql/SqlServerSchemaReader.java
vladhlinskiy: Done.

mssql-plugin/src/main/java/io/cdap/plugin/mssql/SqlServerDBRecord.java
vladhlinskiy: Done.

mssql-plugin/src/main/java/io/cdap/plugin/mssql/SqlServerDBRecord.java
vladhlinskiy: Done.

mssql-plugin/src/main/java/io/cdap/plugin/mssql/SqlServerSchemaReader.java
vladhlinskiy: Done.

mysql-plugin/src/test/java/io/cdap/plugin/mysql/MysqlSinkTestRun.java
vladhlinskiy: Thanks. Changed.

netezza-plugin/src/test/java/io/cdap/plugin/netezza/NetezzaPluginTestBase.java
vladhlinskiy: Changed.

database-commons/src/test/java/io/cdap/plugin/db/CustomAssertions.java
vladhlinskiy: [Changed to:](https://github.com/data-integrations/database-plugins/pull/45/files#diff-8edf8ec3951da6baf44da046b785ee0fR32)
```
public static final double DELTA = 0.000001;
```

database-commons/src/main/java/io/cdap/plugin/util/DBUtils.java
vladhlinskiy: Missed this in the previous PR.

vladhlinskiy:
Addressed comments and rebased.

mssql-plugin/src/main/java/io/cdap/plugin/mssql/SqlServerDBRecord.java
albertshau: nope, just saying these types of comments are good

database-commons/src/test/java/io/cdap/plugin/db/CustomAssertions.java
albertshau: I see. But doesn't that mean the actual isn't what is expected? Why is the length of the actual array greater? It seems wrong to allow that.

mssql-plugin/docs/SqlServer-batchsink.md
```diff
@@ -75,6 +75,56 @@ https://docs.microsoft.com/en-us/sql/relational-databases/system-compatibility-v
...
+    | REAL             | float                  |                                                                |
+    | SMALLDATETIME    | timestamp              |                                                                |
+    | SMALLINT         | int                    |                                                                |
+    | SMALLMONEY       | decimal                |                                                                |
+    | TEXT             | string                 |                                                                |
+    | TIME             | time                   | TIME data type has the accuracy of 100 nanoseconds which is    |
+    |                  |                        | not currently supported. Values of this type will be rounded   |
+    |                  |                        | to microsecond.                                                |
+    | TINYINT          | int                    |                                                                |
+    | UDT              | bytes                  | UDT mapped as basic type if it's an alias of this type. 'SSN'  |
```
albertshau: I'm not familiar with the UDT type, but from the comment it seems like some user defined type that is just an alias of something else? So it gets mapped to whatever it is an alias of?

UDT types are mapped according to the type they are an alias of. For example, is there is an 'SSN' type that was created as 'CREATE TYPE SSN FROM varchar(11);', that type would get mapped to a CDAP string. Common Language Runtime UDTs are mapped to CDAP bytes.

vladhlinskiy: Yes, you are right. It gets mapped to whatever it is an alias of. 
Type created as `CREATE TYPE BIG_UDT FROM BIGINT NOT NULL` is mapped to `long`. 
Added `BIG_UDT` user defied type to tests, comment changed to:
```
| UDT              | bytes                  | UDT types are mapped according to the type they are an alias   |
|                  |                        | of. For example, is there is an 'SSN' type that was created as |
|                  |                        | 'CREATE TYPE SSN FROM varchar(11);', that type would get       |
|                  |                        | mapped to a CDAP string. Common Language Runtime UDTs are      |
|                  |                        | mapped to CDAP bytes.                                          |
```

vladhlinskiy:
`CustomAssertions#assertBytesEquals` method removed. 
Missed that binary columns were created as `BINARY(100)`. `BINARY` is fixed-length data-type, in our case `100` is greater than the length of the actual arrays. Sorry for the confusion.

@albertshau could you, please, review?

mssql-plugin/src/main/java/io/cdap/plugin/mssql/SqlServerDBRecord.java
vladhlinskiy: Ok. Thanks!

albertshau (approved):
lgtm, please squash commits and I'll merge

PR #1, opened 2019-01-31
Moved database-plugins code to database-commons

ibarbashov:
Except database-plugins code, this PR contains maven configuration to handle every specific database as a different submodule.

PR #46, opened 2019-07-05
 CDAP-15546 Oracle db plugin enhancements: all data types support and proper test coverage

vladhlinskiy:
JIRA: https://issues.cask.co/browse/CDAP-15546
WIKI: https://wiki.cask.co/display/CE/Oracle+database+plugin

**In scope of this PR:**

* Added tests for Source plugin for next types:
```
VARCHAR2,  NVARCHAR2,  FLOAT,  LONG,  BINARY_FLOAT,  BINARY_DOUBLE,  TIMESTAMP WITH TIME ZONE,  TIMESTAMP WITH LOCAL TIME ZONE,  LONG RAW,  ROWID,  UROWID
```

* Added tests for Sink plugin for next types:
```
VARCHAR2, NVARCHAR2, NUMBER, FLOAT, LONG, BINARY_FLOAT, BINARY_DOUBLE, TIMESTAMP, TIMESTAMP WITH TIME ZONE, TIMESTAMP WITH LOCAL TIME ZONE, INTERVAL YEAR TO MONTH, INTERVAL DAY TO SECOND, RAW, LONG RAW, ROWID, UROWID, CHAR, NCHAR
```

* Added support of the following types:
```
LONG, LONG RAW, BINARY_FLOAT, BINARY_DOUBLE, TIMESTAMP WITH TIME ZONE, TIMESTAMP WITH LOCAL TIME ZONE, INTERVAL YEAR TO MONTH, INTERVAL DAY TO SECOND 
```

* Added support of writing primitives as values of decimal logical type in the case of valid precision

Since all Oracle numeric types are based on NUMBER(i.e. INTEGER type is actually NUMBER(38, 0)) we won't be able to use Oracle Sink Plugin with other sources.
It's safe to write primitives as values of decimal logical type in the case of valid precision.

* Mappings added to the reference documentation

**Open questions:**

* `TIMESTAMP WITH TIME ZONE` mapping

`TIMESTAMP WITH TIME ZONE` data type currently converted to UTC time, modifying original time zone.


vladhlinskiy:

database-commons/src/main/java/io/cdap/plugin/db/DBRecord.java
```diff
@@ -65,14 +65,20 @@
...
+   */
+  protected List<String> columns;
+
   /**
    * Used to construct a DBRecord from a StructuredRecord in the ETL Pipeline
    *
    * @param record the {@link StructuredRecord} to construct the {@link DBRecord} from
    */
-  public DBRecord(StructuredRecord record, int[] columnTypes) {
+  public DBRecord(StructuredRecord record, int[] columnTypes, List<String> columns) {
```
albertshau: it seems odd to have a mix of types where one is an array and the other is a list. Additionally, having two separate arguments makes it possible to construct a record where the types and column names don't match up.

We can avoid these problems, if we pass in an array or list of a new ColumnType class, which just contains a column name and type. Or if there is an existing object that the columnTypes array is derived from, that would work too.

database-commons/src/main/java/io/cdap/plugin/util/DBUtils.java
```diff
@@ -258,8 +257,7 @@ public static Object transformValue(int sqlType, int precision, int scale,
           return ((Number) original).intValue();
         case Types.NUMERIC:
         case Types.DECIMAL:
-          BigDecimal decimal = (BigDecimal) original;
-          return new BigDecimal(decimal.unscaledValue(), scale, new MathContext(precision));
```
albertshau: left a question in the other PR about why this is no longer needed

oracle-plugin/docs/Oracle-batchsink.md
```diff
@@ -38,6 +38,38 @@ will be passed to the JDBC driver as connection arguments for JDBC drivers that
 
 **Default Batch Value:** The default batch value that triggers an execution request.
 
+
+Data Types Mapping
+----------
+
+	| Oracle Data Type               | CDAP Schema Data Type        | Comment                |
+	| ------------------------------ | ---------------------------- | ---------------------- |
+	| VARCHAR2                       | Type.STRING                  |                        |
```
albertshau: similar comment in other PR about using the names that people see in the UI.

oracle-plugin/docs/Oracle-batchsource.md
```diff
@@ -53,6 +53,37 @@ will be passed to the JDBC driver as connection arguments for JDBC drivers that
...
+	| RAW                            | Type.BYTES                   |                        |
+	| LONG RAW                       | Type.BYTES                   |                        |
+	| ROWID                          | Type.STRING                  |                        |
+	| UROWID                         | Type.STRING                  |                        |
+	| CHAR                           | Type.STRING                  |                        |
+	| NCHAR                          | Type.STRING                  |                        |
+	| CLOB                           | Type.STRING                  |                        |
+	| NCLOB                          | Type.STRING                  |                        |
+	| BLOB                           | Type.BYTES                   |                        |
+	| BFILE                          | Type.BYTES                   | Java API is deprecated |
```
albertshau: An end user is not going to know what to do with the fact that the Java API is deprecated. Is BFile deprecated in Oracle  in general or just in the Java API?

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleDBRecord.java
```diff
@@ -75,6 +111,51 @@ private void handleOracleSpecificType(ResultSet resultSet,
...
+        recordBuilder.set(field.getName(), resultSet.getBytes(columnIndex));
+        break;
+      case Types.DECIMAL:
+      case Types.NUMERIC:
+        // This is the only way to differentiate FLOAT/REAL columns from other numeric columns, that based on NUMBER.
+        // Since FLOAT is a subtype of the NUMBER data type, 'getColumnType' and 'getColumnTypeName' can not be used.
+        if (Double.class.getTypeName().equals(resultSet.getMetaData().getColumnClassName(columnIndex))) {
+          recordBuilder.set(field.getName(), resultSet.getDouble(columnIndex));
+        } else {
+          // It's required to pass 'scale' parameter since in the case of Oracle, scale of 'BigDecimal' depends on the
```
albertshau: nice comment

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSchemaReader.java
```diff
@@ -22,38 +22,110 @@
...
+  public boolean isTypeCompatible(Schema.Field field, ResultSetMetaData metadata, int index) throws SQLException {
+
+    Schema outputFieldSchema = getSchema(metadata, index);
+    Schema outputFieldNonNullableSchema = outputFieldSchema.isNullable()
+      ? outputFieldSchema.getNonNullable()
+      : outputFieldSchema;
+    Schema inputFieldNonNullableSchema = field.getSchema().isNullable()
+      ? field.getSchema().getNonNullable()
+      : field.getSchema();
+    // Handle the case when output schema expects decimal logical type but we got valid primitive.
```
albertshau: When does this happen?

database-commons/src/main/java/io/cdap/plugin/db/DBRecord.java
vladhlinskiy: Changed to use `ColumnType` class.

database-commons/src/main/java/io/cdap/plugin/util/DBUtils.java
vladhlinskiy: These are SQL scale/precision and original instance will have the same scale/precision values in most cases(except of Oracle). We [handle this](https://github.com/data-integrations/database-plugins/pull/46/files#diff-08960aa54a73eedac06b8583a0c5e81bR133) by passing `scale` parameter:
```
BigDecimal decimal = resultSet.getBigDecimal(columnIndex, scale);
```

oracle-plugin/docs/Oracle-batchsink.md
vladhlinskiy: Changed.

oracle-plugin/docs/Oracle-batchsource.md
vladhlinskiy: Comment deleted. Only `BFile` Java API is deprecated and I faced issues reading non-null values of this type. This data type contains a locator to a large binary file stored outside the database. There is no validation while creating these locators and users must ensure that the locator points to an existing file and database process has access to that file. 

I've managed to create such locators but faced issues reading them via JDBC connector. Should I investigate this deeper?

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleDBRecord.java
vladhlinskiy: Thanks!

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSchemaReader.java
vladhlinskiy: In the case when we specify explicit output schema. 

Since all Oracle numeric types are based on NUMBER(i.e. INTEGER type is actually NUMBER(38, 0)) we won't be able to use Oracle Sink Plugin with other sources.
It's safe to write primitives as values of decimal logical type in the case of valid precision.

Thus the following schema compatibility is supported:
1) Schema.Type.INT -> Schema.LogicalType.DECIMAL (if precision of actual decimal logical type >= 10)
2) Schema.Type.LONG -> Schema.LogicalType.DECIMAL (if precision of actual decimal logical type >= 19)
3) Schema.Type.FLOAT -> Schema.LogicalType.DECIMAL (primitive value can be rounded to match actual schema)
4) Schema.Type.DOUBLE -> Schema.LogicalType.DECIMAL (primitive value can be rounded to match actual schema)

vladhlinskiy:

oracle-plugin/docs/Oracle-batchsource.md
albertshau: What is a locator? Is it some custom class in the Oracle JDBC driver? A URI to a file that exists on the DB's local filesystem? A URI that points to an external file?

We definitely shouldn't be in a situation where the pipeline fails if it encounters a non-null value for a BFile column. If we can't get the actual content, then is there a string representation of the locator? What does that look like?

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSchemaReader.java
albertshau: I see. I guess this is where you ran into that UI bug where there is no way to specify the precision and scale?

Can you add this information to the comment in the code?

database-commons/src/main/java/io/cdap/plugin/db/ColumnType.java
```diff
@@ -0,0 +1,46 @@
...
+
+  public ColumnType(String name) {
+    this.name = name;
+  }
+
+  public String getName() {
+    return name;
+  }
+
+  public void setName(String name) {
```
albertshau: Instead of setters, there should be a constructor that takes name and type and assigns them to final fields. It is much easier to reason about correctness when fields are immutable.

database-commons/src/main/java/io/cdap/plugin/db/ColumnType.java
```diff
@@ -0,0 +1,46 @@
...
+
+package io.cdap.plugin.db;
+
+/**
+ * Stores SQL column name and type.
+ */
+public class ColumnType {
+
+  private String name;
+  private Integer type;
```
albertshau: can this ever be null? If not, we should use an int instead of Integer

database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java
```diff
@@ -165,11 +164,14 @@ public void prepareRun(BatchSinkContext context) {
    *
    */
   protected void setColumnsInfo(List<Schema.Field> fields) {
-    columns = fields.stream()
+    columnTypes = fields.stream()
```
albertshau: we shouldn't use the ColumnType class to represent just the column names. Do we need to build this column type list here? Can it be built when the ResultSetMetadata is fetched?

oracle-plugin/docs/Oracle-batchsource.md
```diff
@@ -53,6 +53,37 @@ will be passed to the JDBC driver as connection arguments for JDBC drivers that
...
+    | NVARCHAR2                      | string                |                        |
+    | VARCHAR                        | string                |                        |
+    | NUMBER                         | decimal               |                        |
+    | FLOAT                          | double                |                        |
+    | LONG                           | string                |                        |
+    | DATE                           | timestamp             |                        |
+    | BINARY_FLOAT                   | float                 |                        |
+    | BINARY_DOUBLE                  | double                |                        |
+    | TIMESTAMP                      | timestamp             |                        |
+    | TIMESTAMP WITH TIME ZONE       | timestamp             | Converted to UTC time  |
```
albertshau: In https://wiki.cask.co/display/CE/Oracle+database+plugin, I mentioned that we should read these as strings until there is a way to represent this natively in the CDAP Schema. That way no data is lost (users can always convert from a string to a timestamp using a wrangler transform after the source).

Was this just an issue for Oracle? I imagine the other DBs have the same issue?

oracle-plugin/docs/Oracle-batchsink.md
```diff
@@ -38,6 +38,38 @@ will be passed to the JDBC driver as connection arguments for JDBC drivers that
...
+    | LONG RAW                       | bytes                 |                        |
+    | ROWID                          | string                |                        |
+    | UROWID                         | string                |                        |
+    | CHAR                           | string                |                        |
+    | NCHAR                          | string                |                        |
+    | CLOB                           | string                |                        |
+    | NCLOB                          | string                |                        |
+    | BLOB                           | bytes                 |                        |
+    | BFILE                          | bytes                 |                        |
+
```
albertshau, line 75: Have we been using the same table for source and sink? Aren't there some types that are not supported by the sink? How would a BFile work for example? How do other types like interval work? Do they assume the string is a specific format?

Can we document which data types are not supported by the sink? And if there are requirements around what makes a valid string, we need to document those as well, and document what happens if an invalid value is given.

albertshau: I didn't think about this for other DBs, but I imagine we have similar issues in those docs.

oracle-plugin/docs/Oracle-batchsource.md
vladhlinskiy: BFILE is a data type used to store a locator (link) to an external file, which is stored outside of the database.
Values of this data type created as: 
```
CREATE TABLE files (ID INT, BFILE_COL BFILE);
CREATE DIRECTORY my_dir AS '/tmp';
INSERT INTO files VALUES (1, BFILENAME('MY_DIR', 'pic1.gif'));

```

We can select such values as:
```
SQL> select BFILE_COL from files;

BFILE_COL
--------------------------------------------------------------------------------
bfilename('MY_DIR', 'pic1.gif')
```

The [documentation contains](https://docs.oracle.com/cd/B19306_01/java.102/b14355/oralob.htm#CHDIGFBA) the following note: 
> You can create a BFILE. However, you cannot create an operating system file that a BFILE would refer to. Those are created only externally.

I finally managed to read BFILE locators via JDBC connector, so i added support of this data type for Source and Sink plugins(Yes, there is a custom [BFILE](https://docs.oracle.com/database/121/JAJDB/oracle/sql/BFILE.html) class in the Oracle JDBC driver, instances of which can be mapped to the `Type.BYTES`).

The current implementation allows to read and write locators(and not content of the external files), mapping BFILE to Type.BYTES.
Also, BFILE locator of nonexistent file [added to Source/Sink tests](https://github.com/data-integrations/database-plugins/pull/46/files#diff-9c6424c7b677d0b14bfdab5253a3a18dR295).

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSchemaReader.java
vladhlinskiy: Yes, I faced the UI bug testing the Oracle Sink with other sources. 

[Added the info to the comment](https://github.com/data-integrations/database-plugins/pull/46/files#diff-02bcda70486e82bad237c444b0d7f592R66):
```
    // Handle the case when explicit output schema expects decimal logical type but we got valid primitive.
    // Since all Oracle numeric types are based on NUMBER(i.e. INTEGER type is actually NUMBER(38, 0)) we won't be able
    // to use Oracle Sink Plugin with other sources. It's safe to write primitives as values of decimal logical type in
    // the case of valid precision.
    // Thus the following schema compatibility is supported:
    // 1) Schema.Type.INT -> Schema.LogicalType.DECIMAL (if precision of actual decimal logical type >= 10)
    // 2) Schema.Type.LONG -> Schema.LogicalType.DECIMAL (if precision of actual decimal logical type >= 19)
    // 3) Schema.Type.FLOAT -> Schema.LogicalType.DECIMAL (primitive value can be rounded to match actual schema)
    // 4) Schema.Type.DOUBLE -> Schema.LogicalType.DECIMAL (primitive value can be rounded to match actual schema)
    if (Schema.LogicalType.DECIMAL == outputFieldNonNullableSchema.getLogicalType()) {
      int precision = metadata.getPrecision(index);
      switch (inputFieldNonNullableSchema.getType()) {
    ...
```

database-commons/src/main/java/io/cdap/plugin/db/ColumnType.java
vladhlinskiy: Changed.

database-commons/src/main/java/io/cdap/plugin/db/ColumnType.java
vladhlinskiy: Changed to use `int` instead of `Integer`.

database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java
vladhlinskiy: Changed to build type list in `AbstractDBSink#setResultSetMetadata` method.

oracle-plugin/docs/Oracle-batchsource.md
vladhlinskiy: Mapping of `TIMESTAMP WITH TIME ZONE` changed to `Types.STRING`. There was the same issue with 'DATETIMEOFFSET' MS SQL Server data type, but we mapped it to the `Types.STRING`.

oracle-plugin/docs/Oracle-batchsink.md
vladhlinskiy: With the most recent updates Source and Sink support the same data types.
BFile is mapped to the `Types.BYTE` and locator will be written fine as it would be created via CLI using `BFILENAME`  function([from the comment above](https://github.com/data-integrations/database-plugins/pull/46#discussion_r303600292)).

Interval types expect string literals in the standard format, [documented by Oracle](https://docs.oracle.com/database/121/SQLRF/sql_elements003.htm#SQLRF00221).

Added the following comments:
```
| TIMESTAMP WITH TIME ZONE       | string                | Timestamp string in the following format:              |
|                                |                       | "2019-07-15 15:57:46.65 GMT"                           |
| TIMESTAMP WITH LOCAL TIME ZONE | timestamp             |                                                        |
| INTERVAL YEAR TO MONTH         | string                | Oracle's 'INTERVAL YEAR TO MONTH' literal in the       |
|                                |                       | standard format: "year[-month]"                        |
| INTERVAL DAY TO SECOND         | string                | Oracle's 'INTERVAL DAY TO SECOND' literal in the       |
|                                |                       | standard format:                                       |
|                                |                       | "[day] [hour][:minutes][:seconds[.milliseconds]"       |

...
| BFILE                          | bytes                 | BFILE is a data type used to store a locator (link)    |
|                                |                       | to an external file, which is stored outside of the    |
|                                |                       | database. Only the locator will be written to an       |
|                                |                       | Oracle table and not the content of the external file. |
```

Currently, in the case of invalid string literals, pipeline will fail with the following exception:
```
java.lang.IllegalStateException: MapReduce JobId job_local269563738_0003 failed
	at com.google.common.base.Preconditions.checkState(Preconditions.java:176) ~[com.google.guava.guava-13.0.1.jar:na]
	at io.cdap.cdap.internal.app.runtime.batch.MapReduceRuntimeService.run(MapReduceRuntimeService.java:416) ~[na:na]
	at com.google.common.util.concurrent.AbstractExecutionThreadService$1$1.run(AbstractExecutionThreadService.java:52) ~[com.google.guava.guava-13.0.1.jar:na]
	at io.cdap.cdap.internal.app.runtime.batch.MapReduceRuntimeService$2$1.run(MapReduceRuntimeService.java:450) [na:na]
	at java.lang.Thread.run(Thread.java:748) [na:1.8.0_144]
```
Should this be changed to something meaningful?

> I didn't think about this for other DBs, but I imagine we have similar issues in those docs.

Reference doc of the MS SQL Sink [contains examples](https://github.com/data-integrations/database-plugins/blob/develop/mssql-plugin/docs/SqlServer-batchsink.md#data-types-mapping) of `GEOMETRY` and `GEOGRAPHY` string values, but there is no example of valid `DATETIMEOFFSET` string value.
Also, there are missing mappings in the reference docs for MySQL and Netezza plugins. Should i add them in the scope of this PR?

mysql-plugin/src/test/java/io/cdap/plugin/mysql/MysqlSinkTestRun.java
```diff
@@ -104,12 +103,13 @@ public void testDBSinkWithInvalidFieldLogicalType() throws Exception {
 
   @Test
   public void testDBSinkWithDBSchemaAndInvalidData() throws Exception {
-    String stringColumnName = "NAME";
-    startPipelineAndWriteInvalidData(stringColumnName, getSinkConfig(), DATAPIPELINE_ARTIFACT);
+    // MySQL JDBC connector allows to write integer values to STRING column. Use ENUM column instead.
+    String enumColumnName = "ENUM_COL";
+    startPipelineAndWriteInvalidData(enumColumnName, getSinkConfig(), DATAPIPELINE_ARTIFACT);
```
vladhlinskiy, line 108: In the scope of this PR, `AbstractDBSink#transform` [method was modified](https://github.com/data-integrations/database-plugins/pull/46/files#diff-5570db2ea5dd53e68f60ef41d9c8f6e8L211) to emit original input record and not create a new one(based on the output schema). 
```
  @Override
  public void transform(StructuredRecord input, Emitter<KeyValue<DBRecord, NullWritable>> emitter) {
    emitter.emit(new KeyValue<>(getDBRecord(input), null));
  }
```

This allows us to provide the following schema compatibility:
1. Schema.Type.INT -> Schema.LogicalType.DECIMAL (if precision of actual decimal logical type >= 10)
2. Schema.Type.LONG -> Schema.LogicalType.DECIMAL (if precision of actual decimal logical type >= 19)
3. Schema.Type.FLOAT -> Schema.LogicalType.DECIMAL (primitive value can be rounded to match actual schema)
4. Schema.Type.DOUBLE -> Schema.LogicalType.DECIMAL (primitive value can be rounded to match actual schema)

These changes broke `testDBSinkWithDBSchemaAndInvalidData` test, since now we are trying to write original input record(and not transformed one) and will fail only if JDBC connector does not allow to perform the write. 

MySQL JDBC connector allows to write integer values to STRING column, converting to string implicitly, so here we use `ENUM` column instead.

Another approach would be revert `AbstractDBSink#transform` changes and transform primitives into corresponding instances of `BigDecimal` explicitly.

vladhlinskiy:

vladhlinskiy:
There is an option to map Oracle `NUMBER` types as:
1. scale == 0 && precision < 10                      -> Type.INT (any 9-digits value can stored in java.lang.Long value)
2. scale == 0 && precision >= 10 && precision < 19   -> Type.LONG (any 18-digits value can stored in java.lang.Long value)
3. scale != 0 || scale == 0 && precision >= 19       -> LogicalType.DECIMAL

But `INT`, `SMALLINT` will still be mapped to the `LogicalType.DECIMAL`, since they are actually `NUMBER(38, 0)`.

Here is [old implementation](https://github.com/data-integrations/database-plugins/blob/5f2118190047afe0828794fc52e91716de5080bb/database-commons/src/main/java/io/cdap/plugin/db/CommonSchemaReader.java#L90):
```
      case Types.NUMERIC:
      case Types.DECIMAL:
        int precision = metadata.getPrecision(index); // total number of digits
        int scale = metadata.getScale(index); // digits after the decimal point
        // if there are no digits after the point, use integer types
        type = scale != 0 ? Schema.Type.DOUBLE :
          // with 10 digits we can represent 2^32 and LONG is required
          precision > 9 ? Schema.Type.LONG : Schema.Type.INT;
        break;
```
Such implementation will map `INT`, `SMALLINT` to the `Type.LONG` and this could lead to an overflow on our side.

oracle-plugin/docs/Oracle-batchsource.md
albertshau: What is the byte value of a BFILE locator? That seems really odd to me. Is it https://docs.oracle.com/database/121/JAJDB/oracle/jdbc/OracleBfile.html#getBytes_long__int_ ? That seems like the actual file contents.

oracle-plugin/docs/Oracle-batchsink.md
albertshau: Open separate PRs for the other DB plugins. 

I still don't understand what happens with BFiles. What does a CLI have anything to do with what is happening here? What are the bytes that are somehow being interpreted as a location?

mysql-plugin/src/test/java/io/cdap/plugin/mysql/MysqlSinkTestRun.java
albertshau: I'm confused how that has to do with data types. Previously, the record was being created in order to filter out fields that are present in the input schema but not in the output table.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleDBRecord.java
```diff
@@ -50,31 +61,208 @@ protected SchemaReader getSchemaReader() {
...
+   * @return bytes representation of 'oracle.sql.BFILE' obtained via 'oracle.sql.BFILE#getBytes'.
+   */
+  private byte[] getBfileBytes(ResultSet resultSet, Object bfile) {
+    if (bfile == null) {
+      return null;
+    }
+    try {
+      ClassLoader classLoader = resultSet.getClass().getClassLoader();
+      Class<?> bfileClass = classLoader.loadClass("oracle.sql.BFILE");
+      return (byte[]) bfileClass.getMethod("getBytes").invoke(bfile);
```
albertshau: the javadoc for BFile says it's deprecated and OracleBfile should be used instead, which doesn't have the getBytes() method, but does have a way to get an InputStream for reading the file, and also has a method to check if the file exists. 

You said this returns a byte representation of the locator? I just don't see how that would be useful to anybody, why not return the contents of the file?

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleDBRecord.java
```diff
@@ -50,31 +61,208 @@ protected SchemaReader getSchemaReader() {
...
+    } catch (ClassNotFoundException e) {
+      throw new InvalidStageException("Unable to load 'oracle.sql.BFILE'.", e);
```
albertshau: If this happens, the user should know what went wrong and how to proceed from the error message. It would be more useful to say something like:

Column '%s' is of type 'BFILE', which is not supported with this version of the JDBC driver.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleDBRecord.java
```diff
@@ -50,31 +61,208 @@ protected SchemaReader getSchemaReader() {
...
+        // Note, that ResultSet#getObject retrieves only locator (link) to an external binary file (file stored outside
+        // of the database) and not content of the file.
+        Object bfile = resultSet.getObject(columnIndex);
+        recordBuilder.set(field.getName(), getBfileBytes(resultSet, bfile));
+        break;
+      case OracleSchemaReader.LONG_RAW:
+        recordBuilder.set(field.getName(), resultSet.getBytes(columnIndex));
+        break;
+      case Types.DECIMAL:
+      case Types.NUMERIC:
```
albertshau: do the 'integer', 'smallint', etc Oracle types come back as NUMERIC?

In general, I'm confused where those types are being mapped to a CDAP numeric logical type with correct precision and scale. If you hit the 'Get Schema' button, does the plugin return decimal for those columns?

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleDBRecord.java
```diff
@@ -50,31 +61,208 @@ protected SchemaReader getSchemaReader() {
...
+   * @param connection sql connection.
+   * @param bytes BFILE locator's bytes.
+   * @return instance of 'oracle.sql.BFILE'.
+   */
+  private Object createOracleBfile(Connection connection, byte[] bytes) {
+    try {
+      ClassLoader classLoader = connection.getClass().getClassLoader();
+      Class<?> oracleConnectionClass = classLoader.loadClass("oracle.jdbc.OracleConnection");
+      Class<?> bfileClass = classLoader.loadClass("oracle.sql.BFILE");
+      return bfileClass.getConstructor(oracleConnectionClass, byte[].class).newInstance(connection, bytes);
```
albertshau: It seems like a bad idea to create a BFile since the javadoc says it's deprecated and explicitly mentions to use OracleBfile instead, which has no way to be instantiated. I think we should just not support BFile for the sink.

vladhlinskiy: Removed. Sink changed to not support `BFILE` data type.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleDBRecord.java
vladhlinskiy: Message changed.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleDBRecord.java
vladhlinskiy: > You said this returns a byte representation of the locator? I just don't see how that would be useful to anybody, why not return the contents of the file?

Maybe it will be useful when Source and Sink use Oracle tables of the same database. In other cases, the user will be forced to move external files manually. I was trying to represent only locators because Oracle's CLI does so.

BFILE mapping changed to retrieve actual file contents.

mysql-plugin/src/test/java/io/cdap/plugin/mysql/MysqlSinkTestRun.java
vladhlinskiy: > I'm confused how that has to do with data types. 

From your discussion with Maksym at the meeting on June 25, I heard that we should not transform a field, instead, we should fail writing field if it's not compatible.
Seems that I got it wrong.

We can support writing primitives to decimal columns in two ways:
1. Transform primitives into corresponding instances of BigDecimal explicitly to match the output schema

Handle the case when output schema expects decimal logical type but we got valid primitive. Transform primitives into corresponding instances of BigDecimal.

2. Emit original record and write primitives using corresponding methods of PreparedStatement

For example, it's possible to write `int` to `NUMBER(38, 0)` column using `PreparedStatement#setInt` method.

Drawbacks: 
We lose some sort of validation. The broken test is a good example.
With the **old implementation**, if we create a new record based on the output schema, we will **fail writing an integer to string** column in MySQL Sink.
However, the MySQL JDBC connector allows doing so and **proposed implementation will store integer to string column.**  

* Also, we can map Oracle NUMBER types as I mentioned in the https://github.com/data-integrations/database-plugins/pull/46#issuecomment-511931265
and do not support writing primitives to decimal columns at all.

*******************************************************************

> Previously, the record was being created in order to filter out fields that are present in the input schema but not in the output table.

Proposed implementation [assumes that some of the output schema fields can be absent in the actual record](https://github.com/data-integrations/database-plugins/pull/46/files#diff-4799ee249e76f7b18abb28c4d52191b7R270):

```
  protected void writeToDB(PreparedStatement stmt, @Nullable Schema.Field field, int fieldIndex) throws SQLException {

    int sqlType = columnTypes.get(fieldIndex).getType();
    if (field == null) {
      // Some of the fields can be absent in the record
      stmt.setNull(sqlIndex, sqlType);
      return;
    }

   ...
```

I would like to show you the differences in the behavior of proposed and old implementation.
Let's consider the following scenario with simple source-to-sink pipeline. Tested this with Oracle plugins built from this branch and from 5f2118190047afe0828794fc52e91716de5080bb.

Source table created as: 
```
create table source(FIRST VARCHAR(50), SECOND VARCHAR(50), THIRD VARCHAR(50));
insert into source values('first', 'second', 'third');
```

Sink table:
```
create table sink(FIRST VARCHAR(50));
```

1. No schema specified in UI(nor manually, nor via 'Get Schema' button)

**Proposed implementation:** Succeeded;

**Old implementation:**
Deployed successfully, failed with:
```
java.lang.IllegalArgumentException: Input field 'SECOND' is not found in columns
	at com.google.common.base.Preconditions.checkArgument(Preconditions.java:119) ~[com.google.guava.guava-13.0.1.jar:na]
	at io.cdap.plugin.db.batch.sink.AbstractDBSink.transform(AbstractDBSink.java:209) ~[database-commons-1.3.0-SNAPSHOT.jar:na] 

```

2. Output schema with single `FIRST` field specified 

**Proposed implementation:** Succeeded;

**Old implementation:** Succeeded;

3. Output schema with all fields specified(`FIRST`, `SECOND`, `THIRD`)

**Proposed implementation:** 
Failed to deploy with:
```
Failed to configure pipeline: Error configuring stage 'OracleSink': java.sql.SQLException: Invalid column name
```

**Old implementation:**
Deployed successfully, failed with:
```
java.sql.SQLException: Invalid column name
	at oracle.jdbc.driver.OracleStatement.getColumnIndex(OracleStatement.java:3757) ~[na:na]
	at oracle.jdbc.driver.InsensitiveScrollableResultSet.findColumn(InsensitiveScrollableResultSet.java:264) ~[na:na]
```

*********************************************************************

Could you please share your opinion regarding the following questions:
1. What is preferred mapping for Oracle NUMBER types?
2. Should we support writing primitives to decimal columns? If so, what approach is preferred? Should I revert `AbstractDBSink#transform` method changes?

Thanks in advance!

oracle-plugin/docs/Oracle-batchsource.md
vladhlinskiy: I used this method: https://docs.oracle.com/database/121/JAJDB/oracle/sql/Datum.html#getBytes__

Changed to use https://docs.oracle.com/database/121/JAJDB/oracle/jdbc/OracleBfile.html#getBinaryStream__
to retrieve the entire BFILE as a stream.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSink.java
```diff
@@ -45,8 +50,57 @@ public OracleSink(OracleSinkConfig oracleSinkConfig) {
...
   @Override
-  protected DBRecord getDBRecord(StructuredRecord.Builder output) {
-    return new OracleDBRecord(output.build(), columnTypes);
+  protected DBRecord getDBRecord(StructuredRecord output) {
+    return new OracleSinkDBRecord(output, columnTypes);
+  }
+
+  @Override
+  public boolean isFieldCompatible(Schema.Field field, ResultSetMetaData metadata, int index) throws SQLException {
+    Schema outputFieldSchema = getSchemaReader().getSchema(metadata, index);
```
albertshau: OK I think some of my confusion comes from the fact that there is an extra layer of indirection here, where the Oracle types in the ResultSetMetaData are being changed into CDAP types here.  

This should be looking at the Oracle type directly and checking that the CDAP type is compatible, similar to what is happening in the OracleSourceDBRecord.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceSchemaReader.java
```diff
@@ -0,0 +1,93 @@
...
+        return Schema.of(Schema.Type.DOUBLE);
+      case BFILE:
+      case LONG_RAW:
+        return Schema.of(Schema.Type.BYTES);
+      case INTERVAL_DS:
+      case INTERVAL_YM:
+      case LONG:
+        return Schema.of(Schema.Type.STRING);
+      case Types.NUMERIC:
+      case Types.DECIMAL:
```
albertshau, line 80: When there is an 'Integer' column in a table, does the jdbc driver return that it's an int type? Or does it return that its a numeric type? 

What I'm confused about is what the 'Get Schema' button will return for an 'Integer' column. What I would expect is that it used to return a CDAP int type, but now it returns a CDAP decimal with 0 scale and 38 precision. But I don't see how that happens from this code.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceSchemaReader.java
```diff
@@ -0,0 +1,93 @@
...
+        // This is the only way to differentiate FLOAT/REAL columns from other numeric columns, that based on NUMBER.
+        // Since in Oracle FLOAT is a subtype of the NUMBER data type, 'getColumnType' and 'getColumnTypeName' can not
+        // be used.
+        if (Double.class.getTypeName().equals(metadata.getColumnClassName(index))) {
```
albertshau, line 82: what are the possible values for column class name, and why is Double handled specially? Which oracle type becomes a double?

oracle-plugin/docs/Oracle-batchsink.md
vladhlinskiy: I used this method: https://docs.oracle.com/database/121/JAJDB/oracle/sql/Datum.html#getBytes__
to
> Create a new Java byte array containing a copy of the RDBMS data.

After that, the bytes were converted back to BFILE instance.

Changed to use https://docs.oracle.com/database/121/JAJDB/oracle/jdbc/OracleBfile.html#getBinaryStream__
to retrieve the entire BFILE as a stream.

vladhlinskiy: Ok, I will open separate PRs. Thanks.

mysql-plugin/src/test/java/io/cdap/plugin/mysql/MysqlSinkTestRun.java
albertshau: For #1, In general, we should not be transforming values. For example, if the input field is a CDAP int, we should not write it to a database varchar column.

Oracle is a special case because they implement the standard sql type 'integer' as a numeric. So In order to read it, we have to read it as a decimal, like you have here. But when writing, the sink should allow writing a CDAP decimal or an int. This is just an Oracle thing, and only when the column is declared as an 'integer' -- which is really a number(38, 0) -- because of how the standard sql 'integer' type is handled.

For #2, this was my original question, I don't see how it's related to transforming things. It looks like it was just taking a subset of the input fields and creating a new record out of them. There should not be any transformation happening there. Or are you saying the actual values there didn't match what is in the schema? We should *never* set a value that doesn't match the schema.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleDBRecord.java
vladhlinskiy: Yes,  they come back as NUMERIC.

'Get Schema' button returns decimal for those columns, the exported pipeline JSON file will contain the following entry:
```
"schema": "{\"type\":\"record\",\"name\":\"etlSchemaBody\",\"fields\":[{\"name\":\"INT_COL\",\"type\":[{\"type\":\"bytes\",\"logicalType\":\"decimal\",\"precision\":38},\"null\"]},{\"name\":\"INTEGER_COL\",\"type\":[{\"type\":\"bytes\",\"logicalType\":\"decimal\",\"precision\":38},\"null\"]},{\"name\":\"SMALL\",\"type\":[{\"type\":\"bytes\",\"logicalType\":\"decimal\",\"precision\":38},\"null\"]}]}"
```

We read them as:
https://github.com/data-integrations/database-plugins/pull/46/files#diff-aef8f5d3e726ab43b5650285f66bcc1fR219
```
      case Types.DECIMAL:
      case Types.NUMERIC:
        // This is the only way to differentiate FLOAT/REAL columns from other numeric columns, that based on NUMBER.
        // Since FLOAT is a subtype of the NUMBER data type, 'getColumnType' and 'getColumnTypeName' can not be used.
        if (Double.class.getTypeName().equals(resultSet.getMetaData().getColumnClassName(columnIndex))) {
          recordBuilder.set(field.getName(), resultSet.getDouble(columnIndex));
        } else {
          // It's required to pass 'scale' parameter since in the case of Oracle, scale of 'BigDecimal' depends on the
          // scale of actual value. For example for value '77.12' scale will be '2' even if sql scale is '6'
          BigDecimal decimal = resultSet.getBigDecimal(columnIndex, scale);
          recordBuilder.setDecimal(field.getName(), decimal);
        }
```

Currently, if precision/scale of the explicit input schema does not match sql precision/scale, the Source will fail. Should we do rounding in this case?

vladhlinskiy:
@albertshau 
Could you please share your opinion regarding the following questions:
from https://github.com/data-integrations/database-plugins/pull/46#discussion_r304547812
* What is preferred mapping for Oracle NUMBER types?
* Should we support writing primitives to decimal columns? If so, what approach is preferred? Should I revert AbstractDBSink#transform method changes?

from https://github.com/data-integrations/database-plugins/pull/46#discussion_r304610960
* Currently, if precision/scale of the explicit input schema does not match sql precision/scale, the Source will fail. Should we do rounding in this case?

Thanks in advance!

albertshau:
I think it's correct to error if the precision/scale do not match. Rounding should be handled by a transform or wrangler directive, otherwise every source would have to have rounding logic.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceSchemaReader.java
vladhlinskiy: `sqlType` reported by JDBC connector for Oracle 'INTEGER' column equals to `Types.NUMERIC`. Here we handle `FLOAT`/`REAL` columns, that also reported as `Types.NUMERIC` and return `super.getSchema(metadata, index);` for other cases. So `INTEGER` was mapped to the decimal logical type by `CommonSchemaReader`.

Changed to:
```
      case Types.NUMERIC:
      case Types.DECIMAL:
        // This is the only way to differentiate FLOAT/REAL columns from other numeric columns, that based on NUMBER.
        // Since in Oracle FLOAT is a subtype of the NUMBER data type, 'getColumnType' and 'getColumnTypeName' can not
        // be used.
        if (Double.class.getTypeName().equals(metadata.getColumnClassName(index))) {
          return Schema.of(Schema.Type.DOUBLE);
        } else {
          int precision = metadata.getPrecision(index); // total number of digits
          int scale = metadata.getScale(index); // digits after the decimal point
          return Schema.decimalOf(precision, scale);
        }
```

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSink.java
vladhlinskiy: [Method changed](https://github.com/data-integrations/database-plugins/pull/46/files/#diff-02bcda70486e82bad237c444b0d7f592R58) to check compatibility looking at the Oracle type directly:

```
    ....
    switch (fieldType) {
      case FLOAT:
        return sqlType == OracleSinkSchemaReader.BINARY_FLOAT
          || super.isFieldCompatible(field, metadata, index);
      case BYTES:
        return sqlType == OracleSinkSchemaReader.BFILE
          || sqlType == OracleSinkSchemaReader.LONG_RAW
          || super.isFieldCompatible(field, metadata, index);
      case STRING:
        return sqlType == OracleSinkSchemaReader.LONG
          || sqlType == OracleSinkSchemaReader.TIMESTAMP_TZ
          || sqlType == OracleSinkSchemaReader.INTERVAL_DS
          || sqlType == OracleSinkSchemaReader.INTERVAL_YM
          || sqlType == Types.ROWID
    ...
```

This logic [removed](https://github.com/data-integrations/database-plugins/pull/46/files/#diff-960d064534f854ff7d9011de29fdce95L107) from `DBUtils` class.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceSchemaReader.java
vladhlinskiy: `FLOAT`/`REAL` columns are also reported as `Types.NUMERIC` by JDBC driver and this is the only way to differentiate these columns from other numeric columns. 

`metadata.getColumnClassName` returns `java.lang.Double` for these columns, whereas `java.math.BigDecimal` returned for other numeric columns. 

These are mapped to Double since they can store values greater than `Float.MAX_VALUE` and Oracle's JDBC driver returns `java.lang.Double` if we read these columns via `ResultSet#getObject`.

********************************************************************

> Which oracle type becomes a double?

`BINARY_DOUBLE`, `FLOAT`, `REAL`

mysql-plugin/src/test/java/io/cdap/plugin/mysql/MysqlSinkTestRun.java
vladhlinskiy: >  For example, if the input field is a CDAP int, we should not write it to a database varchar column.

Currently, there is no validation for each received input record. Only explicit schema is validated.
We just copying corresponding fields from input record to a new one, created based on the output schema. For instance, with the old implementation, we will be able to write values of decimal logical type to BINARY columns(if there is no schema set, so no validation is triggered). 

> Or are you saying the actual values there didn't match what is in the schema?

Yes. 
Let's consider the following case:

0. Having old implementation of [AbstractDbSink#transform](https://github.com/data-integrations/database-plugins/blob/develop/database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java#L211) method:
```
  @Override
  public void transform(StructuredRecord input, Emitter<KeyValue<DBRecord, NullWritable>> emitter) {
    // Create StructuredRecord that only has the columns in this.columns
    List<Schema.Field> outputFields = new ArrayList<>();
    for (Schema.Field field : input.getSchema().getFields()) {
      Preconditions.checkArgument(columns.contains(field.getName()), "Input field '%s' is not found in columns",
                                  field.getName());
      outputFields.add(field);
    }
    StructuredRecord.Builder output = StructuredRecord.builder(outputSchema);
    for (String column : columns) {
      output.set(column, input.get(column));
    }

    emitter.emit(new KeyValue<>(getDBRecord(output), null));
  }
```

1. Sink table created as:

```
create table sink(INT_COL INTEGER, SMALL_COL SMALLINT); 
```

2. There is a [logic in the Sink plugin, that infers the schema](https://github.com/data-integrations/database-plugins/blob/develop/database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java#L185) if it's not set. 

```
    try {
      if (Objects.nonNull(outputSchema)) {
        validateSchema(driverClass, dbSinkConfig.tableName, outputSchema);
      } else {
        outputSchema = inferSchema(driverClass);
      }
    } finally {
      DBUtils.cleanup(driverClass);
    }
```
This schema will be used in `AbstractDbSink#transform` method to build an actual record.
The inferred schema will contain two fields of decimal logical type.

_I'm wondering is it required to infer a schema in the Sink plugin? Isn't it passed by the Source? Or there are some cases when this will be useful?_ 

3. Transform receives input record with primitives

Input record's schema contains two fields with the same names but of `Type.INT` instead of decimal logical type.

4. `AbstractDbSink#transform` method creates a new `StructuredRecord` with decimal schema, just copying corresponding values.

Here we have the situation when actual values don't match the schema of the newly created `StructuredRecord`.

5. We fail to access primitives as values of decimal logical type while writing

We expect to access values via `StructuredRecord#getDecimal` since according to the record's schema they are of decimal logical type.

*************************************************************************************************
So i see the following options:
* Do not infer a schema in the Sink plugin
* Do not support writing primitives with inferred Sink schema
* Transform primitives of the `inputRecord` into corresponding instances of `BigDecimal`
* Do not transform original `inputRecord`, write it using its schema. So, we will write field of the `inputRecord` of type `INT` as integer(`PreparedStatement#setInt`), and will fail if JDBC connector does not allow us doing so. This is how the proposed implementation works now. Should i change it to one of the options above?

vladhlinskiy:
@albertshau could you, please, take a look and confirm that we should change current implementation for https://github.com/data-integrations/database-plugins/pull/46#discussion_r305024029?

vladhlinskiy:
@albertshau could you, please, take a look and confirm that we should change current implementation for https://github.com/data-integrations/database-plugins/pull/46#discussion_r305024029?

mysql-plugin/src/test/java/io/cdap/plugin/mysql/MysqlSinkTestRun.java
albertshau: OK I think I understand the problem better now.

I must not have looked carefully when inferSchema() was first added. It is never correct to guess the input schema based on the output database table. The input schema is determined by the previous stage in the pipeline -- not by the output table. In the prepareRun() method, it is just configuring the output format so it uses inferSchema() to configure the column names. This is fine. In the initialize() method, it should *not* use inferSchema(). It is incorrect to assume the records have a schema that match the one coming from the db table.

When context.getInputSchema() is not null, it is guaranteed that every record given to the transform() method has that schema. When it is null, there is no guarantee. The schema can be anything. Incoming records could have different schemas.

What this plugin is responsible for is mapping the types from the incoming record into the type expected by the oracle jdbc driver. Does the driver require setting a BigDecimal for 'integer' columns? Then the plugin should ensure that incoming ints are changed to BigDecimal before writing to the DB.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceSchemaReader.java
albertshau: I see. It wasn't clear to me from the comment. Can you modify it to say the BINARY_DOUBLE, FLOAT, and REAL are returned as SQL Types.NUMERIC, but with value that is a java Double?

vladhlinskiy: Sorry, I misunderstood the question. FLOAT/REAL are returned as java.sql.Types.NUMERIC and BINARY_DOUBLE is returned as custom Oracle type:
```
public static final int BINARY_DOUBLE = 101;
```

Comment changed to:
```
...
// FLOAT and REAL are returned as java.sql.Types.NUMERIC but with value that is a java.lang.Double
if (Double.class.getTypeName().equals(metadata.getColumnClassName(index))) {
...
```

mysql-plugin/src/test/java/io/cdap/plugin/mysql/MysqlSinkTestRun.java
vladhlinskiy: >  In the initialize() method, it should not use inferSchema(). It is incorrect to assume the records have a schema that match the one coming from the db table.

[CDAP documentation says](https://docs.cask.co/cdap/current/en/developer-manual/pipelines/developing-plugins/creating-a-plugin.html#H2231) that `initialize()` method must create any resources required by `transform()` method. 
I believe that's why `inferSchema()` was added to this method.

It configures the column names. These names are used by `setResultSetMetadata` method to initialize `columnTypes` list, which is actually used in the `transform()` method. 

The proposed implementation does not use the inferred schema for other cases except this one. 
The old implementation uses the inferred schema in `transfrom()` method to create a new record. 


> When it is null, there is no guarantee. The schema can be anything. Incoming records could have different schemas.

The proposed implementation works exactly in this way. It uses the schema of received record to write record's fields. 
That's why MySQL test, which expects that an attempt to write int to string column must fail, was broken.

The old implementation:
1. the test does not set a schema
2. we receive int field of Type.INT 
3. we copy it to a new record, using inferred schema, which infers that this field is of Type.STRING
4. an attempt to get the int from a StructuredRecord fails

Proposed implementation:
1. the test does not set a schema
2. we receive int field of Type.INT
3. an attempt to write int to string column succeeds since JDBC connector allows doing so

> Does the driver require setting a BigDecimal for 'integer' columns?

No, it allows us to set Java integers and that's how the proposed implementation writes primitives to 'integer' columns.

vladhlinskiy:

database-commons/src/main/java/io/cdap/plugin/db/ColumnType.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+package io.cdap.plugin.db;
+
+/**
+ * Stores SQL column name and type.
+ */
+public class ColumnType {
+
+  private String name;
```
albertshau: these should be final

database-commons/src/main/java/io/cdap/plugin/db/ColumnType.java
```diff
@@ -0,0 +1,39 @@
...
+  public ColumnType(String name, int type) {
+    this.name = name;
+    this.type = type;
+  }
+
+  public String getName() {
+    return name;
+  }
+
+  public Integer getType() {
```
albertshau: int instead of Integer

database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java
```diff
@@ -209,23 +211,11 @@ private Schema inferSchema(Class<? extends Driver> driverClass) {
...
   @Override
   public void transform(StructuredRecord input, Emitter<KeyValue<DBRecord, NullWritable>> emitter) {
-    // Create StructuredRecord that only has the columns in this.columns
-    List<Schema.Field> outputFields = new ArrayList<>();
-    for (Schema.Field field : input.getSchema().getFields()) {
-      Preconditions.checkArgument(columns.contains(field.getName()), "Input field '%s' is not found in columns",
-                                  field.getName());
-      outputFields.add(field);
-    }
-    StructuredRecord.Builder output = StructuredRecord.builder(outputSchema);
```
albertshau, line 219: since outputSchema is no longer used here, it no longer needs to be set in the initialize() call either, and can probably be removed as a variable.

database-commons/src/main/java/io/cdap/plugin/util/DBUtils.java
```diff
@@ -104,148 +98,6 @@ public static DriverCleanup ensureJDBCDriverIsAvailable(Class<? extends Driver>
...
-   * list of {@link io.cdap.cdap.api.data.schema.Schema.Field},
-   * where name of the field is same as column name and type of the field is obtained using
-   * {@link DBUtils#getSchema(int, int, int)}
-   *
-   * @param resultSet result set of executed query
-   * @param schemaStr schema string to override resultant schema
-   * @return list of schema fields
-   * @throws SQLException
-   */
-  public static List<Schema.Field> getSchemaFields(ResultSet resultSet, @Nullable String schemaStr)
```
albertshau, line 118: these are being removed because the logic has been moved into SchemaReader?

albertshau (commented):
small comments otherwise lgtm. Please fix and squash commits, then I'll merge. Thanks for being patient with all the back and forth.

database-commons/src/main/java/io/cdap/plugin/util/DBUtils.java
vladhlinskiy: I removed this method addressing https://github.com/data-integrations/database-plugins/pull/46#discussion_r305006496 because i did not find usages of it.

database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java
vladhlinskiy: `outputSchema` instance variable removed. Changed to use local variables, did not remove them at all in order to maintain readability.

database-commons/src/main/java/io/cdap/plugin/db/ColumnType.java
vladhlinskiy: Changed. Thanks.

database-commons/src/main/java/io/cdap/plugin/db/ColumnType.java
vladhlinskiy: Changed. Thanks.

vladhlinskiy:
@albertshau comments are addressed and commits are squashed.

PR #47, opened 2019-07-08
CDAP-15518 Investigate pipeline failure on demo environment: input schema validation

mlozbin-cybervisiontech:
JIRA: https://issues.cask.co/browse/CDAP-15518

**In scope of this PR:**

* Added input schema validation in sink plugin during pipeline configuration
* Added check of logical types during schema validation
An attempt to write data type LONG instead of data type LONG with logical type TIMESTAMP_MICROS cased pipeline fail in MSSQL sink plugin on demo environment
* Added tests for schema validation in MySQL, MSSQL, Postgresql, Oracle, DB2, Netezza sink plugins for such cases:
   * incorrect data type
   * incorrect data logical type
* Added tests for invalid data write attempt in MySQL, MSSQL, Postgresql, Oracle, DB2, Netezza sink plugins
* Fixed connection close in DB2 sink plugin after attempt to write invalid data

### DIFF
--- a/aurora-mysql-plugin/src/test/java/io/cdap/plugin/aurora/mysql/AuroraMysqlSinkTestRun.java
+++ b/aurora-mysql-plugin/src/test/java/io/cdap/plugin/aurora/mysql/AuroraMysqlSinkTestRun.java
@@ -31,6 +31,7 @@ import io.cdap.plugin.auroradb.mysql.AuroraMysqlConstants;
 import io.cdap.plugin.common.Constants;
 import io.cdap.plugin.db.batch.sink.AbstractDBSink;
 import org.junit.Assert;
+import org.junit.Before;
 import org.junit.Test;
 
 import java.math.BigDecimal;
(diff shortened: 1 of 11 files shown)


mlozbin-cybervisiontech:
@albertshau could you, please, take a look.

database-commons/src/main/java/io/cdap/plugin/db/batch/sink/ETLDBOutputFormat.java
```diff
@@ -110,9 +110,14 @@ public void close(TaskAttemptContext context) throws IOException {
         }
 
         @Override
-        public void write(K key, V value) throws IOException {
-          super.write(key, value);
+        public void write(K key, V value) {
           emptyData = false;
+          try {
+            key.write(getStatement());
```
albertshau, line 117: can you add a comment about why super.write() is not used? It seems like it's because the exception gets lost?

database-commons/src/main/java/io/cdap/plugin/db/batch/sink/ETLDBOutputFormat.java
```diff
@@ -110,9 +110,14 @@ public void close(TaskAttemptContext context) throws IOException {
...
+            getStatement().addBatch();
+          } catch (SQLException e) {
+            LOG.warn("Failed to write value to database", e);
```
albertshau, line 120: is there a reason it logs and continues rather than throwing an exception and failing immediately?

database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java
```diff
@@ -327,15 +333,21 @@ private void validateFields(Schema inputSchema, ResultSet rs) throws SQLExceptio
...
       boolean isNotNullAssignable = !isColumnNullable && field.getSchema().isNullable();
-      boolean isNotCompatible = !Objects.equals(fieldType, columnType);
+      boolean isNotCompatible = !(Objects.equals(fieldType, columnType)
+        && Objects.equals(fieldLogicalType, columnLogicalType));
 
       if (isNotCompatible) {
         invalidFields.add(field.getName());
-        LOG.error("Field {} was given as type {} but the database column is actually of type {}.", field.getName(),
-                  fieldType, columnSchema.getType());
+        LOG.error("Field {} was given as type {} and logical type {} " +
```
albertshau: users don't actually know anything about physical and logical types. They just see a single type in the UI, where timestamp is in the same dropdown as long.

Instead of logging both, we should keep the old message, but use the LogicalType.token() value if the logical type is not null and the type otherwise.

aurora-mysql-plugin/src/test/java/io/cdap/plugin/aurora/mysql/AuroraMysqlSinkTestRun.java
```diff
@@ -54,20 +56,59 @@
...
+  public void setup() throws Exception {
+    try (Connection connection = createConnection();
+         Statement stmt = connection.createStatement()) {
+      stmt.execute("TRUNCATE TABLE MY_DEST_TABLE");
+    }
+  }
+
+  @Test
+  public void testDBSinkWithInvalidFieldType() throws Exception {
+    String inputDatasetName = "input-dbsinktest-invalid-field-type";
```
albertshau: it looks like there is a lot of code duplication, can we move the duplicated code to a common class?

database-commons/src/main/java/io/cdap/plugin/db/batch/sink/ETLDBOutputFormat.java
mlozbin-cybervisiontech: this is used instead of super method to use proper logging.

database-commons/src/main/java/io/cdap/plugin/db/batch/sink/ETLDBOutputFormat.java
mlozbin-cybervisiontech: it just copies logic from parent class. Did not want to break something on which DBOutputFormat relies.

database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java
mlozbin-cybervisiontech: Done

aurora-mysql-plugin/src/test/java/io/cdap/plugin/aurora/mysql/AuroraMysqlSinkTestRun.java
mlozbin-cybervisiontech: Done

mlozbin-cybervisiontech:
@albertshau I addressed all your comments, could you, please, take a look.

albertshau (approved):
lgtm, please squash commits and I'll merge

PR #51, opened 2019-07-22
CDAP-15519 Reference doc updates

vladhlinskiy:
JIRA: https://issues.cask.co/browse/CDAP-15519

In the scope of this PR:
* Missing mappings added to the reference docs for MySQL and Netezza plugins
* Added example of a valid `DATETIMEOFFSET` string value for MS SQL plugin


vladhlinskiy:

albertshau (approved):

PR #54, opened 2019-08-05
CDAP-15658 Make the DB integration tests automated

mlozbin-cybervisiontech:
JIRA: [https://issues.cask.co/browse/CDAP-1565](https://issues.cask.co/browse/CDAP-15658)

In scope of this PR:
- fixed broken integration tests for Generic Plugin
- fixed broken unit tests for DBSink
- moved schema and fields validation methods to separate class
- added unit tests for schema validations
- updated surefire plugin to support multiple attached DB driver artifacts
- documented Aurora DB test properties
- changed default test properties to work on local environment
 classes


mlozbin-cybervisiontech:
@albertshau could you, please, take a look.

database-commons/src/main/java/io/cdap/plugin/util/DBUtils.java
```diff
@@ -184,6 +189,60 @@ public static void validateJDBCPluginPipeline(PipelineConfigurer pipelineConfigu
...
+  /**
+   * Checks if fields from schema are compatible to be written into database.
+   *
+   * @param actualSchema schema from db.
+   * @param configSchema schema from config.
+   */
+  public static void validateSourceSchema(Schema actualSchema, Schema configSchema) {
+    if (configSchema == null) {
+      throw new InvalidConfigPropertyException("Schema should not be null or empty",
+                                               AbstractDBSource.DBSourceConfig.SCHEMA);
```
albertshau: this validate method is only ever correct to be called from a source. It's better to keep it in the AbstractDBSource, but make it a package private static method that you can call from unit tests.

database-commons/pom.xml
```diff
@@ -59,6 +59,12 @@
       <artifactId>cdap-api</artifactId>
       <scope>compile</scope>
     </dependency>
+    <dependency>
+      <groupId>com.mockrunner</groupId>
+      <artifactId>mockrunner-jdbc</artifactId>
```
albertshau, line 64: where is this being used?

generic-database-plugin/src/test/java/io/cdap/plugin/db/batch/sink/DBSinkTest.java
```diff
@@ -1,149 +0,0 @@
...
-import org.junit.Assert;
-import org.junit.Test;
-
-import java.util.Map;
-import java.util.stream.Collectors;
-
-/**
- * Tests for {@link AbstractDBSink}
- */
-public class DBSinkTest {
```
albertshau, line 37: why remove this?

netezza-plugin/src/test/java/io/cdap/plugin/netezza/NetezzaPluginTestBase.java
```diff
@@ -51,7 +51,7 @@
 import java.util.Map;
 import java.util.TimeZone;
 
-public class NetezzaPluginTestBase extends DatabasePluginTestBase {
+public abstract class NetezzaPluginTestBase extends DatabasePluginTestBase {
```
albertshau, line 54: how come these are being changed to abstract classes?

albertshau (commented):
It's hard to tell whether there are logical changes being made, I'm assuming it's just moving code to Validator classes instead of in the plugins. Those changes lgtm assuming it's just moving stuff.

I'm confused about the mockrunner-jdbc changes. Can you include more information about what is happening there? 

In the future, please separate changes like these into multiple PRs. It looks like there are unrelated changes that could have been separated with minimal conflicts.

database-commons/src/main/java/io/cdap/plugin/util/DBUtils.java
mlozbin-cybervisiontech: done

database-commons/pom.xml
mlozbin-cybervisiontech: I’m using mockrunner-jdbc to mock ResultSet and ResultSetMetaData. They are used in CommonFieldsValidatorTest and AbstractDBSinkTest classes.

netezza-plugin/src/test/java/io/cdap/plugin/netezza/NetezzaPluginTestBase.java
mlozbin-cybervisiontech: As *TestBase classes don’t have methods with "@ Test" annotation so we received tests fail for each plugin with cause: “no runnable methods”.

generic-database-plugin/src/test/java/io/cdap/plugin/db/batch/sink/DBSinkTest.java
mlozbin-cybervisiontech: This class tested transform method in Sink plugin. For now transform method doesn't validate schema from db and input. Instead of this class I’ve added unit tests, which cover these cases, in CommonFieldsValidatorTest class in database-commons module.

mlozbin-cybervisiontech:
@albertshau I addressed all your comments, could you, please, take a look.

albertshau (approved):
lgtm

PR #57, opened 2019-10-15
[CDAP-15787] Add failure collector to database plugins

CuriousVini:


database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java
```diff
@@ -91,12 +94,16 @@ private String getJDBCPluginId() {
   @Override
   public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
     super.configurePipeline(pipelineConfigurer);
+    StageConfigurer configurer = pipelineConfigurer.getStageConfigurer();
     DBUtils.validateJDBCPluginPipeline(pipelineConfigurer, dbSinkConfig, getJDBCPluginId());
```
albertshau, line 98: does this need to be updated too?

database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java
```diff
@@ -269,15 +278,16 @@ private void setResultSetMetadata() throws Exception {
...
 
     try {
       DBUtils.ensureJDBCDriverIsAvailable(jdbcDriverClass, connectionString, dbSinkConfig.jdbcPluginName);
     } catch (IllegalAccessException | InstantiationException | SQLException e) {
-      throw new InvalidStageException(String.format("Unable to load or register JDBC driver '%s' while checking for " +
-                                                      "the existence of the database table '%s'.",
-                                                    jdbcDriverClass, tableName), e);
+      collector.addFailure(String.format("Unable to load or register JDBC driver '%s' while checking for " +
+                                           "the existence of the database table '%s'.",
+                                         jdbcDriverClass, tableName), null).withStacktrace(e.getStackTrace());
```
albertshau, line 290: if this fails, doesn't that mean the other logic will fail?

database-commons/src/main/java/io/cdap/plugin/db/batch/sink/CommonFieldsValidator.java
```diff
@@ -35,7 +35,7 @@
   protected static final Logger LOG = LoggerFactory.getLogger(CommonFieldsValidator.class);
 
   @Override
-  public void validateFields(Schema inputSchema, ResultSet resultSet) throws SQLException {
+  public Set<String> validateFields(Schema inputSchema, ResultSet resultSet) throws SQLException {
```
albertshau: should this take a FailureCollector instead of logging errors?

database-commons/src/main/java/io/cdap/plugin/util/DBUtils.java
```diff
@@ -163,17 +163,23 @@ public static void deregisterAllDrivers(Class<? extends Driver> driverClass)
...
                                                 String jdbcPluginId) {
-    if (Objects.isNull(config.user) && Objects.nonNull(config.password)) {
-      throw new InvalidConfigPropertyException("user", "user is null. Please provide both user name and password if " +
-        "database requires authentication. If not, please remove password and retry.");
+    FailureCollector collector = pipelineConfigurer.getStageConfigurer().getFailureCollector();
+    if (!config.containsMacro(DBConfig.USER) && !config.containsMacro(DBConfig.PASSWORD) &&
+      Objects.isNull(config.user) && Objects.nonNull(config.password)) {
+      collector.addFailure("Username and password should be provided together.", "Please provide both " +
+        "user name and password, if database supports it or remove password.")
+        .withConfigProperty(ConnectionConfig.USER).withConfigProperty(ConnectionConfig.PASSWORD);
```
albertshau: this should only highlight the user property, since there is nothing wrong with providing a password. The message can also be simplified to just say that a user name is required when a password is given.

database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java
CuriousVini: yes that has been updated in the PR as well. Since it takes `pipelineConfigurer` we do not need to pass `FailureCollector` explicitly.

database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java
CuriousVini: yea, thats right. Modified the code to incorporate this.

database-commons/src/main/java/io/cdap/plugin/db/batch/sink/CommonFieldsValidator.java
CuriousVini: This returns the set of invalid fields and the caller can decide to either log it or add to failure collector.

albertshau: but the caller doesn't know why the field is invalid, and this method already logs an error for each problem.

database-commons/src/main/java/io/cdap/plugin/util/DBUtils.java
CuriousVini: done

database-commons/src/main/java/io/cdap/plugin/db/batch/sink/CommonFieldsValidator.java
CuriousVini: done

database-commons/src/main/java/io/cdap/plugin/db/batch/sink/CommonFieldsValidator.java
```diff
@@ -35,35 +36,32 @@
...
-    Set<String> invalidFields = new HashSet<>();
     for (Schema.Field field : inputSchema.getFields()) {
       int columnIndex = resultSet.findColumn(field.getName());
       boolean isColumnNullable = (ResultSetMetaData.columnNullable == rsMetaData.isNullable(columnIndex));
       boolean isNotNullAssignable = !isColumnNullable && field.getSchema().isNullable();
+      String name = field.getName();
       if (isNotNullAssignable) {
-        LOG.error("Field '{}' was given as nullable but the database column is not nullable", field.getName());
-        invalidFields.add(field.getName());
+        LOG.error("Field '{}' was given as nullable but the database column is not nullable.", name);
```
albertshau: shouldn't log here. This is similar to how logging and throwing an exception is a bad pattern since it usually results in multiple logs.

database-commons/src/main/java/io/cdap/plugin/db/batch/sink/CommonFieldsValidator.java
```diff
@@ -35,35 +36,32 @@
...
+          String.format("Field '%s' was given as nullable but database column is not nullable.", name),
+          "Ensure that the field is not nullable.").withInputSchemaField(name);
       }
 
       if (!isFieldCompatible(field, rsMetaData, columnIndex)) {
         String sqlTypeName = rsMetaData.getColumnTypeName(columnIndex);
         Schema fieldSchema = field.getSchema().isNullable() ? field.getSchema().getNonNullable() : field.getSchema();
-        Schema.Type fieldType = fieldSchema.getType();
-        Schema.LogicalType fieldLogicalType = fieldSchema.getLogicalType();
         LOG.error("Field '{}' was given as type '{}' but the database column is actually of type '{}'.",
```
albertshau: same here

albertshau (approved):
one minor comment, otherwise lgtm

PR #74, opened 2020-06-25
Mariadb Plugins

amolkokane-cirus:
Added missing javadocs and fixed pom.xml versions
Changed copyright to 2020


mariadb-plugin/src/main/java/io/cdap/plugin/mariadb/MariadbAction.java
```diff
@@ -0,0 +1,113 @@
...
+    super(mariadbActionConfig, false);
+    this.mariadbActionConfig = mariadbActionConfig;
+  }
+
+  /**
+   * MariaDB Action Config.
+   */
+  public static class MariadbActionConfig extends DBSpecificQueryConfig {
+
+    @Name(MariadbConstants.AUTO_RECONNECT)
```
rmstar, line 57: Most of these configs are common to all mariadb plugins. Can you put he common configuration into a class to avoid duplication?

mariadb-plugin/pom.xml
```diff
@@ -0,0 +1,118 @@
...
+
+    <name>Maria DB plugin</name>
+    <artifactId>mariadb-plugin</artifactId>
+    <version>1.4.0-SNAPSHOT</version>
+    <modelVersion>4.0.0</modelVersion>
+
+    <dependencies>
+        <dependency>
+            <groupId>io.cdap.cdap</groupId>
+            <artifactId>cdap-etl-api</artifactId>
```
rmstar, line 34: the scope should be provided.
https://github.com/data-integrations/google-cloud/blob/develop/pom.xml#L150

mariadb-plugin/pom.xml
```diff
@@ -0,0 +1,118 @@
...
+        <dependency>
+            <groupId>io.cdap.plugin</groupId>
+            <artifactId>database-commons</artifactId>
+            <version>${project.version}</version>
+            <type>test-jar</type>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.cdap.cdap</groupId>
+            <artifactId>hydrator-test</artifactId>
```
rmstar, line 60: scope should be test
https://github.com/data-integrations/google-cloud/blob/develop/pom.xml#L416
Please take a look at above pom file for scope of the artifacts.

amolkokane-cirus: This dependency has been already declared with scope as "test" in the parent pom.xml file. Do we still have to explicitly mention the scope here? We checked pom.xml files from other sub-projects under database-plugins and none of them has scope explicitly mentioned. Can you please confirm?

mariadb-plugin/pom.xml
amolkokane-cirus: This dependency has been already declared with scope as "provided" in the parent pom.xml file. Do we still have to explicitly mention the scope here? We checked pom.xml files from other sub-projects under database-plugins and none of them has scope explicitly mentioned. Can you please confirm?

mariadb-plugin/src/main/java/io/cdap/plugin/mariadb/MariadbAction.java
amolkokane-cirus: TODO

shiva-cirus: The code is modified to specify TODO in comment and reference to JIRA - https://issues.cask.co/browse/PLUGIN-358

rmstar (approved):

CuriousVini:
Please squash commits

PR #75, opened 2020-06-26
Teradata Plugins

amolkokane-cirus:
Added missing javadocs and fixed pom.xml versions
Changed copyright to 2020


teradata-plugin/pom.xml
```diff
@@ -0,0 +1,119 @@
...
+
+  <artifactId>teradata-plugin</artifactId>
+  <name>Teradata plugin</name>
+  <version>1.4.0-SNAPSHOT</version>
+  <modelVersion>4.0.0</modelVersion>
+
+  <dependencies>
+    <dependency>
+      <groupId>io.cdap.cdap</groupId>
+      <artifactId>cdap-etl-api</artifactId>
```
rmstar, line 35: the scope should be provided.
https://github.com/data-integrations/google-cloud/blob/develop/pom.xml#L150
Please take a look at above pom file for scope of the artifacts

rmstar (approved):

teradata-plugin/pom.xml
amolkokane-cirus: This dependency has been already declared with scope as "provided" in the parent pom.xml file. Do we still have to explicitly mention the scope here? We checked pom.xml files from other sub-projects under database-plugins and none of them has scope explicitly mentioned. Can you please confirm?

rmstar: Oh I see. Please ignore my comment

shiva-cirus:
Squash complete.

PR #94, opened 2020-10-26
bugfix/PLUGIN-375-PostgresSQL-decimal

ailegion:
added: Validation check for 0 decimal scale issue

JIRA: https://issues.cask.co/browse/PLUGIN-375

### DIFF
--- a/database-commons/src/main/java/io/cdap/plugin/db/CommonSchemaReader.java
+++ b/database-commons/src/main/java/io/cdap/plugin/db/CommonSchemaReader.java
@@ -90,6 +90,15 @@ public class CommonSchemaReader implements SchemaReader {
       case Types.DECIMAL:
         int precision = metadata.getPrecision(index); // total number of digits
         int scale = metadata.getScale(index); // digits after the decimal point
+        // decimal type with scale 0 is not supported
+        // possible cases are:
+        //    - scale is set to 0
+        //    - only precision is set - by default scale is 0
+        //    - precision and scale are set as dynamic - by default scale is 0
+        if (scale == 0) {
+          throw new SQLException(new UnsupportedTypeException(String.format("Unsupported SQL Type: %s with scale 0.",
+                                                                            metadata.getColumnTypeName(index))));
+        }
         return Schema.decimalOf(precision, scale);
 
       case Types.DOUBLE:


postgresql-plugin/src/main/java/io/cdap/plugin/postgres/PostgresSchemaReader.java
```diff
@@ -43,6 +46,14 @@ public Schema getSchema(ResultSetMetaData metadata, int index) throws SQLExcepti
     String typeName = metadata.getColumnTypeName(index);
     int columnType = metadata.getColumnType(index);
 
+    // raise exception if scale is not set for decimal type
+    if (columnType == 2) {
+      if (metadata.getScale(index) == 0) {
+        throw new ValidationException(Collections.singletonList(
```
CuriousVini: Why are we not using failure collector? When is this method `getSchema` being called?

ailegion: This method is being called from ``AbstractDBSource->getSchema->loadSchemaFromDB->getSchemaReader().getSchemaFields`` which is located in ``CommonSchemaReader`` and calls ``getSchema``.

We could not find a way to access failure collector from here.

CuriousVini: This logic should be added in `CommonSchemaReader#getSchema()`. When scale is 0 then throw checked exception and handle it in caller `AbstractDBSource`

CuriousVini: As per our offline discussion we need to handle following cases:

1.) Numeric with precision and scale
If scale is zero we should throw a validation error as we dont support scale to be zero

2.) Numeric with precision
By default, it will have scale zero. So we should throw a validation error.

3.) Numeric with varying precision and scale in a given column
This case we will not support. So throw an error if columns have varying scale and precision.

How are we handling #3?

ailegion: updated PR. As for handling #3 - it was already there as ``setDecimal`` method in ``StructuredRecord`` raises the following error :
375-PostreSQL-Decimal-scale-mistmatch

CuriousVini: For #3 is it possible to raise this at validation time? So that user doesnt have to run the pipeline to find out it wont work?

ailegion: When reading column metadata during schema detection for columns with dynamic scale and precision both are returned as 0 which is caught by current validation condition for scale 0. 
However if user sets output schema manually and sets custom scale and precision value then the dynamic scale issue can only be detected while trying to match the actual record schema to output schema (which is happens currently).
I don't see how we can catch this without reading a record that has mismatch with output schema.

database-commons/src/main/java/io/cdap/plugin/db/CommonSchemaReader.java
```diff
@@ -90,6 +90,10 @@ public Schema getSchema(ResultSetMetaData metadata, int index) throws SQLExcepti
       case Types.DECIMAL:
         int precision = metadata.getPrecision(index); // total number of digits
         int scale = metadata.getScale(index); // digits after the decimal point
+        if (scale == 0) {
+          throw new SQLException(new UnsupportedTypeException(String.format("Unsupported SQL Type: %s with scale 0.",
```
CuriousVini, line 99: Why are we wrapping this with sqlexception. It is thrown in case there is error accessing database
https://docs.oracle.com/javase/7/docs/api/java/sql/SQLException.html#:~:text=An%20exception%20that%20provides%20information,available%20via%20the%20method%20getMesasge%20.

postgresql-plugin/src/main/java/io/cdap/plugin/postgres/PostgresSchemaReader.java
CuriousVini: I see. In that case can you add comment in the code explaining behavior in all 3 cases?

CuriousVini (commented):
Just one comment. Fix and squash

ailegion:
> Just one comment. Fix and squash

added comment and squashed

database-commons/src/main/java/io/cdap/plugin/db/CommonSchemaReader.java
ailegion: This is what method currently throws - we based it on the following line https://github.com/data-integrations/database-plugins/blob/540d5a56f9d33888fee27bc6f5846132fff8dd7a/database-commons/src/main/java/io/cdap/plugin/db/CommonSchemaReader.java#L130
Is there a better solution? Also if there is - how to handle it on other callers as this method is being used from all the other db plugins?

CuriousVini (approved):

database-commons/src/main/java/io/cdap/plugin/db/CommonSchemaReader.java
CuriousVini: There is no need to wrap `UnsupportedTypeException ` into `SQLException`. we can just have unsupported type exception

ailegion: changing it to throw UnsupportedTypeException also updates base method signature which is used by all database plugins

CuriousVini: I see. Looks like we are handling it like that in this method.

PR #106, opened 2021-03-02
PLUGIN-620 Do not process arguments during config time if macro is provided for the database configurations.

sagarkapare:
https://cdap.atlassian.net/browse/PLUGIN-620


database-commons/src/main/java/io/cdap/plugin/db/batch/action/ArgumentSetterConfig.java
```diff
@@ -138,4 +138,19 @@ public void validate(FailureCollector collector) {
     }
     collector.getOrThrowException();
   }
+
+  /**
+   * @return {@code true} if any of the database configuration value is provided as macro, o
+   * therwise {@code false} is returned
```
CuriousVini: nit: otherwise in one line

CuriousVini (approved):

PR #108, opened 2021-03-25
Added example pipeline for DB Argument Setter.

jasir99:
Added example pipeline for DB Argument Setter.

### DIFF
--- a/generic-db-argument-setter/examples/example-db-argument-setter-cdap-data-pipeline.json
+++ b/generic-db-argument-setter/examples/example-db-argument-setter-cdap-data-pipeline.json
@@ -1,0 +1,135 @@
+{
+    "name": "db-argument-setter",
+    "description": "Data Pipeline Application",
+    "artifact": {
+        "name": "cdap-data-pipeline",
+        "version": "6.4.0",
+        "scope": "SYSTEM"
+    },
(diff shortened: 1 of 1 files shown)


generic-db-argument-setter/examples/example-db-argument-setter-cdap-data-pipeline.json
```diff
@@ -0,0 +1,135 @@
...
+                        "scope": "USER"
+                    },
+                    "properties": {
+                        "jdbcPluginName": "mysql",
+                        "connectionString": "${connection_string}",
+                        "user": "${username}",
+                        "password": "${password}",
+                        "databaseName": "${database}",
+                        "tableName": "${table}",
+                        "argumentSelectionConditions": "${argument_selection_conditions}",
```
bdmogal: line 113 and 114 shouldn't need to be macros right? Otherwise it is very hard to understand how this works?

bdmogal (changes requested):

bdmogal:
Also, can we make a couple more fields configurable?

bdmogal:
> Also, can we make a couple more fields configurable?

Was this addressed? I still see only `file_path` is a macro. Can we add a few more?

PR #112, opened 2021-05-19
Bump develop version after cutting release branch

MEseifan:
T


rmstar (approved):

yaojiefeng (approved):
LGTM, can we also update the hydrator version to a released one? https://github.com/data-integrations/database-plugins/blob/f3b5e3ac6153864d6e80b5848198e78feb063716/pom.xml#L64

PR #113, opened 2021-05-19
Bump release version to snapshot

MEseifan:


yaojiefeng (approved):
LGTM, same comment on updating the hydrator version, https://github.com/data-integrations/database-plugins/blob/f3b5e3ac6153864d6e80b5848198e78feb063716/pom.xml#L64

PR #120, opened 2021-06-28
migrate to CDAP 6.5.0 and fix failed test suits

seanzhougoogle:


aurora-mysql-plugin/pom.xml
```diff
@@ -79,40 +85,4 @@
       <scope>compile</scope>
     </dependency>
   </dependencies>
-
-  <build>
```
albertshau, line 83: why remove this? It means the plugin won't build.

aurora-mysql-plugin/pom.xml
```diff
@@ -79,40 +85,4 @@
...
-      <plugin>
-        <groupId>org.apache.felix</groupId>
-        <artifactId>maven-bundle-plugin</artifactId>
-        <version>3.3.0</version>
-        <extensions>true</extensions>
-        <configuration>
-          <instructions>
-            <_exportcontents>
-              io.cdap.plugin.*;
-              org.apache.commons.lang;
```
albertshau, line 94: Not related to this work, but these exports seem problematic. Have opened PLUGIN-710 to track it.

We should try to run a pipeline that reads from two different dbs (eg: mysql and sqlserver) and see if it fails. If not, it's something that we can address later.

aurora-mysql-plugin/src/test/java/io/cdap/plugin/aurora/mysql/AuroraMysqlPluginTestBase.java
```diff
@@ -115,9 +109,19 @@ public static void setupTest() throws Exception {
       BASE_PROPS.get(ConnectionConfig.PORT) + "/" + BASE_PROPS.get(ConnectionConfig.DATABASE);
     Connection conn = createConnection();
     createTestTables(conn);
+    tearDown = true;
     prepareTestData(conn);
   }
 
+  private static void getProperties() {
```
albertshau, line 118: why make this change? is it because the properties unavailable with the previous static initialization?

aurora-mysql-plugin/pom.xml
albertshau: ok I see, you added it to the base pom.

pom.xml
```diff
@@ -332,35 +332,37 @@
...
-              </goals>
-            </execution>
-          </executions>
-        </plugin>
-      </plugins>
-    </pluginManagement>
     <plugins>
+      <plugin>
+        <groupId>io.cdap</groupId>
+        <artifactId>cdap-maven-plugin</artifactId>
```
albertshau: this and maven-bundle-plugin should be in the pluginManagement section, because not all of the submodules are plugins (database-commons). The plugin submodules should have a build section while the common ones do not.

aurora-mysql-plugin/src/test/java/io/cdap/plugin/aurora/mysql/AuroraMysqlPluginTestBase.java
```diff
@@ -115,9 +109,19 @@ public static void setupTest() throws Exception {
       BASE_PROPS.get(ConnectionConfig.PORT) + "/" + BASE_PROPS.get(ConnectionConfig.DATABASE);
     Connection conn = createConnection();
     createTestTables(conn);
+    tearDown = true;
```
albertshau: I assume we're doing this because the tear down can fail if the expected tables don't exist?

This is not a good way to guard against that type of scenario though, as it's possible to successfully create some tables but fail partway through. Then tearDown never gets set and the tables don't get cleaned up.

Instead, it's better not to have this flag at all. In the cleanup section, if we can't drop a table, we log a message and move on.

aurora-mysql-plugin/src/test/java/io/cdap/plugin/aurora/mysql/AuroraMysqlPluginTestBase.java
seanzhougoogle: Because the old code will fail if properties are unavailable. So I use `Assume.assume` to skip the test case if properties are not available.

While `Assume.assume` actually throws a specific runtime exception when assumption is false which is handled by outmost Junit Runner to skip the test case instead of failing the test case. Throwing runtime exception in static initialization will cause class initialization failure.

pom.xml
seanzhougoogle: Good catch! Didn't notice that `database-commons` doesn't have this section.

seanzhougoogle: done:)

aurora-mysql-plugin/src/test/java/io/cdap/plugin/aurora/mysql/AuroraMysqlPluginTestBase.java
seanzhougoogle: Thanks for the suggestion. Done :)

aurora-mysql-plugin/src/test/java/io/cdap/plugin/aurora/mysql/AuroraMysqlPluginTestBase.java
```diff
@@ -254,6 +252,9 @@ public static void tearDownDB() throws SQLException {
       stmt.execute("DROP TABLE postActionTest");
       stmt.execute("DROP TABLE dbActionTest");
       stmt.execute("DROP TABLE MY_DEST_TABLE");
+      throw new NullPointerException();
```
albertshau: is this here by accident?

aurora-mysql-plugin/src/test/java/io/cdap/plugin/aurora/mysql/AuroraMysqlPluginTestBase.java
```diff
@@ -254,6 +252,9 @@ public static void tearDownDB() throws SQLException {
...
+    } catch (Exception e) {
+      logger.warn("Fail to tear down.", e);
```
albertshau: it's better to have a try-catch for each drop statement, so that we try and delete the rest of the tables if an earlier one fails. Probably makes sense to have a utility method to do this.

aurora-mysql-plugin/src/test/java/io/cdap/plugin/aurora/mysql/AuroraMysqlPluginTestBase.java
```diff
@@ -56,6 +58,7 @@
 import javax.sql.rowset.serial.SerialBlob;
 
 public abstract class AuroraMysqlPluginTestBase extends DatabasePluginTestBase {
+  private static final Logger logger = LoggerFactory.getLogger(AuroraMysqlPluginTestBase.class);
```
albertshau: style: private static final variable names should be all caps LOGGER

memsql-plugin/pom.xml
```diff
@@ -102,6 +102,10 @@
           </instructions>
         </configuration>
       </plugin>
+      <plugin>
+        <groupId>io.cdap</groupId>
+        <artifactId>cdap-maven-plugin</artifactId>
+      </plugin>
     </plugins>
   </build>
 </project>
```
albertshau: an existing issue, but seems like this file needs a newline at the end.

albertshau (approved):
minor comments, otherwise lgtm.

aurora-mysql-plugin/src/test/java/io/cdap/plugin/aurora/mysql/AuroraMysqlPluginTestBase.java
seanzhougoogle: good catch. it's by accident.

aurora-mysql-plugin/src/test/java/io/cdap/plugin/aurora/mysql/AuroraMysqlPluginTestBase.java
seanzhougoogle: done

seanzhougoogle: done :)

yaojiefeng (approved):

aurora-mysql-plugin/src/test/java/io/cdap/plugin/aurora/mysql/AuroraMysqlPluginTestBase.java
seanzhougoogle: done:)

memsql-plugin/pom.xml
seanzhougoogle: done :)

PR #121, opened 2021-06-28
migrate DB specific connector plugins

seanzhougoogle:


database-commons/src/main/java/io/cdap/plugin/db/ConnectionConfig.java
```diff
@@ -132,4 +132,15 @@ public ConnectionConfig() {
     return Collections.emptyMap();
   }
 
+  public String getJdbcPlughinName() {
```
tivv: typo: `plughin`

database-commons/src/main/java/io/cdap/plugin/db/batch/config/DBSpecificSourceConfig.java
```diff
@@ -40,4 +46,43 @@
   @Description("Database name to connect to")
   @Macro
   public String database;
+
+  protected AbstractDBSpecificConnectorConfig getConnection() {
```
tivv: `@Nullable`

database-commons/src/main/java/io/cdap/plugin/db/batch/config/DBSpecificSourceConfig.java
```diff
@@ -40,4 +46,43 @@
...
+    AbstractDBSpecificConnectorConfig connection = getConnection();
+    if (connection == null) {
+      return super.getConnectionArguments();
+    }
+    Map<String, String> arguments = new HashMap<>();
+    arguments.putAll(Maps.fromProperties(connection.getConnectionArgumentsProperties()));
+    arguments.putAll(getDBSpecificArguments());
+    return arguments;
+  }
+  public String getJdbcPlughinName() {
```
tivv: nit: empty line between methods

database-commons/src/main/java/io/cdap/plugin/db/connector/AbstractDBSpecificConnectorConfig.java
```diff
@@ -0,0 +1,130 @@
...
+    Properties properties = new Properties();
+    properties.putAll(connectionArgumentsMap);
+    return properties;
+  }
+
+  public String getHost() {
+    return host;
+  }
+
+  public abstract int getPort();
```
tivv: Could you handle default value in some other way? E.g. with a separate abstract `getDefaultPort`. It's a bit confusing to see a field and an abstract getter.

database-commons/src/main/java/io/cdap/plugin/db/connector/DBSpecificPath.java
```diff
@@ -0,0 +1,134 @@
...
+ * as database name as "a" and table name or schema name as empty.
+ *
+ */
+public class DBSpecificPath implements DBConnectorPath {
+  private final boolean supportSchema;
+  private String database;
+  private String schema;
+  private String table;
+
+  public DBSpecificPath(String path, boolean supportSchema) {
```
tivv: Consider making a factory method. This would allow you to have all fields final

database-commons/src/main/java/io/cdap/plugin/db/connector/DBSpecificPath.java
```diff
@@ -0,0 +1,134 @@
...
+
+  private void validateName(String property, String name) {
+    if (name.isEmpty()) {
+      throw new IllegalArgumentException(
+        String.format("%s should not be empty.", property));
+    }
+  }
+
+  @Override
+  public boolean containDatabase() {
```
tivv, line 116: Please add javadocs for public methods

database-commons/src/main/java/io/cdap/plugin/db/ConnectionConfig.java
seanzhougoogle: good catch, done :)

database-commons/src/main/java/io/cdap/plugin/db/batch/config/DBSpecificSourceConfig.java
seanzhougoogle: done :)

database-commons/src/main/java/io/cdap/plugin/db/batch/config/DBSpecificSourceConfig.java
seanzhougoogle: done :)

database-commons/src/main/java/io/cdap/plugin/db/connector/AbstractDBSpecificConnectorConfig.java
seanzhougoogle: Good idea :)

seanzhougoogle: done:)

database-commons/src/main/java/io/cdap/plugin/db/connector/DBSpecificPath.java
seanzhougoogle: good idea, done :)

database-commons/src/main/java/io/cdap/plugin/db/connector/DBSpecificPath.java
seanzhougoogle: the is overriding the parent method which already have the javadoc. And the overriding doesn't change any semantics here.

database-commons/src/main/java/io/cdap/plugin/db/batch/config/AbstractDBSpecificSourceConfig.java
```diff
@@ -0,0 +1,239 @@
...
+
+    if (getTransactionIsolationLevel() != null) {
+      TransactionIsolationLevel.validate(getTransactionIsolationLevel(), collector);
+    }
+
+    if (!containsMacro(IMPORT_QUERY) && Strings.isNullOrEmpty(importQuery)) {
+      collector.addFailure("Import Query must be specified.", null).withConfigProperty(IMPORT_QUERY);
+    }
+
+    if (!hasOneSplit && !containsMacro(IMPORT_QUERY) && !getImportQuery().contains("$CONDITIONS")) {
```
tivv, line 128: What if NUM_SPLITs has macro and this guy does not have macro? Do we require CONDITIONS in this case (same for others below)

database-commons/src/main/java/io/cdap/plugin/db/batch/config/AbstractDBSpecificSourceConfig.java
```diff
@@ -0,0 +1,239 @@
...
+      collector.addFailure("Invalid Import Query.",
+                           String.format("Import Query %s must contain the string '$CONDITIONS'.", importQuery))
```
tivv: Do yo want to clarify "if Number of Splits is not set to 1"?

database-commons/src/main/java/io/cdap/plugin/db/batch/action/ArgumentSetterConfig.java
```diff
@@ -115,10 +115,10 @@ public void validate(FailureCollector collector) {
     if (!containsMacro(CONNECTION_STRING) && Strings.isNullOrEmpty(this.connectionString)) {
       collector.addFailure("Invalid connection string.", "Connection string cannot be empty.");
```
tivv: The second parameter should be an action, e.g. "Fill in connection string". So it makes sense to make seconds parameter first (as it describes the failure in more details) and add an action. Same for other two checks.

database-commons/src/main/java/io/cdap/plugin/db/batch/config/AbstractDBSpecificSourceConfig.java
```diff
@@ -0,0 +1,239 @@
...
+        hasOneSplit = true;
+      }
+    }
+
+    if (getTransactionIsolationLevel() != null) {
+      TransactionIsolationLevel.validate(getTransactionIsolationLevel(), collector);
+    }
+
+    if (!containsMacro(IMPORT_QUERY) && Strings.isNullOrEmpty(importQuery)) {
+      collector.addFailure("Import Query must be specified.", null).withConfigProperty(IMPORT_QUERY);
```
tivv: Please also follow the rule that first parameter is the problem, e.g. "Import Query is empty" and second is a corrective action, e.g. "Fill in Import Query". Same for others

database-commons/src/main/java/io/cdap/plugin/db/connector/AbstractDBSpecificConnector.java
```diff
@@ -0,0 +1,124 @@
...
+    return getConnection(getConnectionString(path.getDatabase()) , config.getConnectionArgumentsProperties());
+  }
+
+  protected String getConnectionString(String database) {
+    return config.getConnectionString();
+  }
+
+  protected String getTableQuery(DBSpecificPath path) {
+    return path.getSchema() == null ? String.format("SELECT * from %s.%s", path.getDatabase(), path.getTable()) : String
+      .format("SELECT * from %s.%s.%s", path.getDatabase(), path.getSchema(), path.getTable());
```
tivv, line 112: Are you sure we need to include database? Cross-datadase queries is not a standard feature. E.g. postgresql docs says "The name (optionally schema-qualified) of an existing table or view". (https://www.postgresql.org/docs/current/sql-select.html). It does not say you can include a database name

database-commons/src/main/java/io/cdap/plugin/db/connector/AbstractDBSpecificConnector.java
```diff
@@ -0,0 +1,124 @@
...
+
+
+    return new SourceInputFormatProvider(DataDrivenETLDBInputFormat.class, connectionConfigAccessor.getConfiguration());
+  }
+
+  protected Connection getConnection(DBConnectorPath path) {
+    return getConnection(getConnectionString(path.getDatabase()) , config.getConnectionArgumentsProperties());
+  }
+
+  protected String getConnectionString(String database) {
```
tivv, line 106: Is this overriden anywhere? I don't see the parameter used

database-commons/src/main/java/io/cdap/plugin/db/connector/AbstractDBSpecificConnector.java
tivv: Also it may produce problems if connection is going through some kind of pooler (e.g. PGPOOL) as there going to be a mapping between pooler DBs and database server DBs

postgresql-plugin/src/main/java/io/cdap/plugin/postgres/PostgresConnector.java
```diff
@@ -0,0 +1,108 @@
...
+      RESULTSET_COLUMN_TABLE_CAT));
+  }
+
+  @Override
+  protected String getConnectionString(@Nullable String database) {
+    if (database == null) {
+      return config.getConnectionString();
+    }
+    return String
+      .format(PostgresConstants.POSTGRES_CONNECTION_STRING_FORMAT, config.getHost(), config.getPort(), database);
```
tivv: Why do we have `POSTGRES_CONNECTION_STRING_FORMAT` and `POSTGRE_CONNECTION_STRING_FORMAT` in `PostgresConnectorConfig`?

database-commons/src/main/java/io/cdap/plugin/db/batch/config/AbstractDBSpecificSourceConfig.java
seanzhougoogle: Copied this code somewhere else, I guess the reason why the action is null was because the message itself explains what need to be the action. But I think your suggestion makes the code better. done :)

database-commons/src/main/java/io/cdap/plugin/db/batch/config/AbstractDBSpecificSourceConfig.java
seanzhougoogle: `NUM_SPLITS` doesn't support MACRO, you can see the property is not annotated with `@MACRO`

tivv: Oh, I see you check for it `containsMacro(NUM_SPLITS)` above, thus my question

database-commons/src/main/java/io/cdap/plugin/db/batch/config/AbstractDBSpecificSourceConfig.java
seanzhougoogle: Good catch, copied this code somewhere else and didn't notice this. Done :)

database-commons/src/main/java/io/cdap/plugin/db/batch/action/ArgumentSetterConfig.java
seanzhougoogle: done :)

database-commons/src/main/java/io/cdap/plugin/db/connector/AbstractDBSpecificConnector.java
seanzhougoogle: Yes, it's overridden in the `PostgresConnector`, because Postgres doesn't support  cross DB connection, need to put database in connection string.

database-commons/src/main/java/io/cdap/plugin/db/connector/AbstractDBSpecificConnector.java
seanzhougoogle: Those DBs that don't support Cross-database queries will only connect to the DB being queried. That's why PostgresConnector overrides the `getConnectionString(database)` methods.

Thanks for bring up the issue for pooler, but could you educate me what the problem will be if the database supports Cross-database queries with pooler ?

postgresql-plugin/src/main/java/io/cdap/plugin/postgres/PostgresConnector.java
seanzhougoogle: Good question. Because initially for a connector to test connection, we don't know what database to connect to (User dont' need to specify a database) , so the format in Connector is without database. Only when user dive into a specific database, we need to reconstruct the connection string for that database and here the format contains database.

seanzhougoogle: renamed them respectively to make it more clear.

database-commons/src/main/java/io/cdap/plugin/db/batch/config/AbstractDBSpecificSourceConfig.java
seanzhougoogle: sorry, my fault. It should support Macro. corrected. Good catch !

seanzhougoogle: `hasOneSplit` will only be true when `containsMacro(NUM_SPLITS)` is false.

database-commons/src/main/java/io/cdap/plugin/db/connector/AbstractDBSpecificConnector.java
tivv: Well, e.g. in pgbouncer there is next setting that allows to map database names. This means that database in URI and database within remote server are different beasts, see https://www.pgbouncer.org/config.html
This allows a single pooler to pool multiple servers with potentially overlapping database names

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleConnector.java
```diff
@@ -0,0 +1,103 @@
...
+  public StructuredRecord transform(LongWritable longWritable, OracleSourceDBRecord record) {
+    return record.getRecord();
+  }
+
+  @Override
+  protected ResultSet queryDatabases(Connection connection) throws SQLException {
+    return connection.createStatement()
+      .executeQuery(String.format("SELECT NAME AS %s FROM V$DATABASE", RESULTSET_COLUMN_TABLE_CAT));
+  }
+  @Override
```
tivv, line 116: nit: empty line between methods

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleConnector.java
```diff
@@ -0,0 +1,103 @@
...
+  protected String getTableQuery(DBSpecificPath path) {
+    return String.format("SELECT * from %s.%s", path.getSchema(), path.getTable());
```
tivv, line 118: I see that for oracle you don't pass database here and don't override `getConnection`. How database would be passed to Oracle?
In IMPORT_QUERY you pass database

database-commons/src/main/java/io/cdap/plugin/db/batch/config/AbstractDBSpecificSourceConfig.java
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    if (!hasOneSplit && !containsMacro(IMPORT_QUERY) && !getImportQuery().contains("$CONDITIONS")) {
+      collector.addFailure(String.format(
+        "Import Query %s must contain the string '$CONDITIONS'. if Number of Splits is not set to 1.", importQuery),
+                           "Include '$CONDITIONS' in the Import Query")
+        .withConfigProperty(IMPORT_QUERY);
+    }
+
+    if (!hasOneSplit && !containsMacro("splitBy") && (splitBy == null || splitBy.isEmpty())) {
```
tivv: Should it be `containsMacro(SPLIT_BY)`?

database-commons/src/main/java/io/cdap/plugin/db/batch/config/AbstractDBSpecificSourceConfig.java
```diff
@@ -0,0 +1,246 @@
...
+      collector.addFailure("Split-By Field Name must be specified if Number of Splits is not set to 1.",
+                           "Specify the Split-by Field Name.").withConfigProperty(SPLIT_BY)
+        .withConfigProperty(NUM_SPLITS);
+    }
+
+    if (!hasOneSplit && !containsMacro("boundingQuery") && (boundingQuery == null || boundingQuery.isEmpty())) {
```
tivv: Should it be `containsMacro(BOUNDING_QUERY)`?

database-commons/src/main/java/io/cdap/plugin/db/connector/AbstractDBSpecificConnector.java
seanzhougoogle: Discussed offline and will leave the current code as is and will keep in mind this situation.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleConnector.java
seanzhougoogle: done :)

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleConnector.java
seanzhougoogle: good catch. I should override getConnection for Oracle. also you remind me when generateSpec I should also populate database properties.

database-commons/src/main/java/io/cdap/plugin/db/batch/config/AbstractDBSpecificSourceConfig.java
seanzhougoogle: done :)

database-commons/src/main/java/io/cdap/plugin/db/batch/config/AbstractDBSpecificSourceConfig.java
seanzhougoogle: good catch, done :)

database-commons/src/main/java/io/cdap/plugin/db/batch/config/DatabaseConnectionConfig.java
```diff
@@ -0,0 +1,55 @@
...
+package io.cdap.plugin.db.batch.config;
+
+import java.util.Map;
+
+/**
+ * Interface for DB connection config
+ */
+public interface DatabaseConnectionConfig {
+
+  boolean containsMacro(String property);
```
yaojiefeng: I don't think you need this method, this method is available in `PluginConfig`, this interface shouldn't need it to compute macro

database-commons/src/main/java/io/cdap/plugin/db/batch/config/DatabaseSourceConfig.java
```diff
@@ -0,0 +1,83 @@
...
+   *
+   * @param schemaFromDB the actual schema got from database
+   * @param collector    the failure collector
+   */
+  void validateSchema(Schema schemaFromDB, FailureCollector collector);
+
+  /**
+   * @return the initial query to be run upon connecting to database
+   */
+  List<String> getInitQueries();
```
yaojiefeng, line 82: Can you add more docs on what init query means here?

database-commons/src/main/java/io/cdap/plugin/db/connector/AbstractDBSpecificConnector.java
```diff
@@ -0,0 +1,124 @@
...
+    DBSpecificPath path = DBSpecificPath.of(request.getPath(), supportSchema());
+    if (path.getTable() == null) {
+      throw new IllegalArgumentException(
+        String.format("Path %s cannot be sampled. Must have table name in the path.", request.getPath()));
+    }
+
+    ConnectionConfigAccessor connectionConfigAccessor = new ConnectionConfigAccessor();
+    if (config.getUser() == null && config.getPassword() == null) {
+      DBConfiguration
+        .configureDB(connectionConfigAccessor.getConfiguration(), driverClass.getName(),
```
yaojiefeng: nit - we normally split line on parameter

database-commons/src/main/java/io/cdap/plugin/db/connector/AbstractDBSpecificConnector.java
```diff
@@ -0,0 +1,124 @@
...
+                     getConnectionString(path.getDatabase()));
+    } else {
+      DBConfiguration
+        .configureDB(connectionConfigAccessor.getConfiguration(), driverClass.getName(),
```
yaojiefeng: same here

database-commons/src/main/java/io/cdap/plugin/db/connector/AbstractDBSpecificConnector.java
yaojiefeng: Can you open a JIRA to track and put a TODO here

generic-database-plugin/src/test/java/io/cdap/plugin/GenericDatabasePluginTestSuite.java
```diff
@@ -23,13 +23,15 @@
...
 import org.junit.BeforeClass;
+import org.junit.Ignore;
 import org.junit.runner.RunWith;
 import org.junit.runners.Suite;
 
 
 /**
  * This is a test suite that runs all the tests for Database plugins.
  */
+@Ignore
```
yaojiefeng: Do you know why this is ignored?

mssql-plugin/src/test/java/io/cdap/plugin/mssql/SqlServerConnectorTest.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+package io.cdap.plugin.mssql;
+
+import io.cdap.plugin.db.connector.DBSpecificConnectorBaseTest;
+import org.junit.Test;
+
+import java.io.IOException;
+
+
```
yaojiefeng: nit - extra line

yaojiefeng: You also need to have some javadoc on these public classes, I think checkstyle file is missing in this repo

mysql-plugin/src/test/java/io/cdap/plugin/mysql/MysqlConnectorTest.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+package io.cdap.plugin.mysql;
+
+import io.cdap.plugin.db.connector.DBSpecificConnectorBaseTest;
+import org.junit.Test;
+
+import java.io.IOException;
+
+
```
yaojiefeng, line 24: same here - please add some javadoc

yaojiefeng (commented):
Mostly looks good. The connector plugin is missing the doc file like https://github.com/data-integrations/database-plugins/blob/develop/mysql-plugin/docs/Mysql-batchsource.md

tivv (approved):

database-commons/src/main/java/io/cdap/plugin/db/connector/AbstractDBSpecificConnector.java
seanzhougoogle: Discussed offline, no need to Add TODO here, filed PLUGIN-714 for documenting that we don't support pooler that proxies multiple db server. Because when we connect to DB initially , we don't specify database name in the connection string, if a pooler is proxying multiple DB server, it won't know which server to connect to.

generic-database-plugin/src/test/java/io/cdap/plugin/GenericDatabasePluginTestSuite.java
seanzhougoogle: it's by accident, removed.

mssql-plugin/src/test/java/io/cdap/plugin/mssql/SqlServerConnectorTest.java
seanzhougoogle: checkstyle is not missing in this repo. Test class doesn't require javadoc. Removed the extra line.

mysql-plugin/src/test/java/io/cdap/plugin/mysql/MysqlConnectorTest.java
seanzhougoogle: same as replied previously.

database-commons/src/main/java/io/cdap/plugin/db/connector/AbstractDBSpecificConnector.java
seanzhougoogle: done :)

database-commons/src/main/java/io/cdap/plugin/db/connector/AbstractDBSpecificConnector.java
seanzhougoogle: done :)

database-commons/src/main/java/io/cdap/plugin/db/batch/config/DatabaseSourceConfig.java
seanzhougoogle: discussed offline, it's already explained in the javadoc

database-commons/src/main/java/io/cdap/plugin/db/batch/config/DatabaseConnectionConfig.java
seanzhougoogle: discussed offline, did a refactoring.

yaojiefeng (approved):

PR #125, opened 2021-07-09
Minor fix for DB specific Source

seanzhougoogle:
1. remove duplicated property "connectionType" on UI
2. skip validate jdbc plugin at deploy time if use connection
3. add "browse database" button

### DIFF
--- a/mysql-plugin/widgets/Mysql-batchsource.json
+++ b/mysql-plugin/widgets/Mysql-batchsource.json
@@ -96,6 +96,14 @@
           "name": "database"
         },
         {
+          "widget-type": "connection-browser",
+          "widget-category": "plugin",
+          "widget-attributes": {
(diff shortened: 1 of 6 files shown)


mysql-plugin/widgets/Mysql-batchsource.json
```diff
@@ -95,6 +95,14 @@
           "label": "Database",
           "name": "database"
         },
+        {
+          "widget-type": "connection-browser",
+          "widget-category": "plugin",
```
yaojiefeng, line 100: Can you just add this to all the plugins we migrated? Should be same code change except the type

seanzhougoogle: done , I was planning to make the changes in other PR :)

yaojiefeng (approved):

PR #126, opened 2021-07-09
Fix Oracle Connector

seanzhougoogle:
1. resolve "ORA-01882: timezone region not found" issue
2. add a role property
3. fix the import query in generate spec

### DIFF
--- a/oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleConnectorConfig.java
+++ b/oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleConnectorConfig.java
@@ -20,6 +20,9 @@ import io.cdap.cdap.api.annotation.Description;
 import io.cdap.cdap.api.annotation.Name;
 import io.cdap.plugin.db.connector.AbstractDBSpecificConnectorConfig;
 
+import java.util.Properties;
+import javax.annotation.Nullable;
+
 /**
  * Configuration for Oracle Database Connector
  */
@@ -27,6 +30,8 @@ public class OracleConnectorConfig extends AbstractDBSpecificConnectorConfig {
 
   private static final String ORACLE_CONNECTION_STRING_SID_WITHOUT_DB_FORMAT = "jdbc:oracle:thin:@%s:%s";
   private static final String ORACLE_CONNECTION_STRING_SERVICE_NAME_WITHOUT_DB_FORMAT = "jdbc:oracle:thin:@//%s:%s";
+  private static final String TIME_ZONE_AS_REGION_PROPERTY = "oracle.jdbc.timezoneAsRegion";
+  private static final String INTERNAL_LOGON_PROPERTY = "internal_logon";
 
   public OracleConnectorConfig(String host, int port, String user, String password, String jdbcPluginName,
                                String connectionArguments) {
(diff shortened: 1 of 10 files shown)


oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleConnectorConfig.java
```diff
@@ -65,4 +75,18 @@ protected int getDefaultPort() {
...
+    return role == null ? "normal" : role;
+  }
+
+  @Override
+  public Properties getConnectionArgumentsProperties() {
+    Properties prop = super.getConnectionArgumentsProperties();
+    // To solve the "ORA-01882: timezone region not found" issue, see:
+    // https://stackoverflow.com/questions/9156379/ora-01882-timezone-region-not-found
+    prop.put(TIME_ZONE_AS_REGION_PROPERTY, "false");
+    prop.put(INTERNAL_LOGON_PROPERTY, getRole());
```
yaojiefeng, line 89: In the stackoverflow question, I don't see this role needs to be included, why do we need to introduce this variable?

oracle-plugin/widgets/Oracle-connector.json
```diff
@@ -64,6 +64,29 @@
...
           "name": "password"
+        },
+        {
+          "label": "Role",
+          "name": "role",
+          "widget-type": "radio-group",
+          "widget-attributes": {
+            "layout": "inline",
+            "default": "normal",
+            "options": [
```
yaojiefeng, line 75: Are these the only available roles in oracle?

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleConnectorConfig.java
```diff
@@ -55,7 +60,12 @@ public String getConnectionString() {
 
   @Name(OracleConstants.CONNECTION_TYPE)
   @Description("Whether to use an SID or Service Name when connecting to the database.")
-  public String connectionType;
+  private String connectionType;
+
+  @Name(OracleConstants.ROLE)
+  @Description("The login role of the user when connecting to the database.")
```
yaojiefeng: nit - Remove "The"

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleConnectorConfig.java
```diff
@@ -55,7 +60,12 @@ public String getConnectionString() {
...
+  @Nullable
+  private String role;
```
yaojiefeng, line 68: there is a oracle source md file, can you also add this doc to that file?

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleConnectorConfig.java
seanzhougoogle: This is irrelevant to the stackoverflow question. 

It is a required property, if you are connecting using a SYSDBA or SYSOPR account, you have to specify this property. Actually for those Oracle sql client, you also need to specify the role

oracle-plugin/widgets/Oracle-connector.json
seanzhougoogle: Yes.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleConnectorConfig.java
seanzhougoogle: done :)

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleConnectorConfig.java
seanzhougoogle: done :)

yaojiefeng (approved):

PR #133, opened 2021-07-20
minor fixes:

seanzhougoogle:
1. don't set transaction isolation level when it's null
2. set limit in select query when sampling
3. set timeout param for mysql connector

### DIFF
--- a/mysql-plugin/src/main/java/io/cdap/plugin/mysql/MysqlConnectorConfig.java
+++ b/mysql-plugin/src/main/java/io/cdap/plugin/mysql/MysqlConnectorConfig.java
@@ -18,13 +18,17 @@ package io.cdap.plugin.mysql;
 
 import io.cdap.plugin.db.connector.AbstractDBSpecificConnectorConfig;
 
+import java.util.Properties;
+
 /**
  * Configuration for Mysql Connector
  */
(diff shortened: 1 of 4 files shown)


mysql-plugin/src/main/java/io/cdap/plugin/mysql/MysqlConnectorConfig.java
```diff
@@ -45,4 +49,12 @@ public String getConnectionString() {
   public int getDefaultPort() {
     return 3306;
   }
+
+  @Override
+  public Properties getConnectionArgumentsProperties() {
+    Properties prop = super.getConnectionArgumentsProperties();
+    prop.put(JDBC_PROPERTY_CONNECT_TIMEOUT, "20000");
```
yaojiefeng, line 57: can you put a comment on what unit this time is?

yaojiefeng (approved):
one minor comment, LGTM

mysql-plugin/src/main/java/io/cdap/plugin/mysql/MysqlConnectorConfig.java
seanzhougoogle: done :)

PR #141, opened 2021-07-27
quote names in select query

seanzhougoogle:
this is to avoid problems with special character in the names

### DIFF
--- a/mysql-plugin/src/main/java/io/cdap/plugin/mysql/MysqlConnector.java
+++ b/mysql-plugin/src/main/java/io/cdap/plugin/mysql/MysqlConnector.java
@@ -81,6 +81,16 @@ public class MysqlConnector extends AbstractDBSpecificConnector<DBRecord> {
   }
 
   @Override
+  protected String getTableQuery(DBConnectorPath path) {
+    return String.format("SELECT * FROM `%s`.`%s`", path.getDatabase(), path.getTable());
+  }
+
+  @Override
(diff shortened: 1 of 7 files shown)


mysql-plugin/src/main/java/io/cdap/plugin/mysql/MysqlConnector.java
```diff
@@ -80,6 +80,16 @@ protected void setConnectorSpec(ConnectorSpecRequest request, DBConnectorPath pa
...
   }
 
+  @Override
+  protected String getTableQuery(DBConnectorPath path) {
+    return String.format("SELECT * FROM `%s`.`%s`", path.getDatabase(), path.getTable());
+  }
+
+  @Override
+  protected String getTableQuery(DBConnectorPath path, int limit) {
+    return String.format("SELECT * FROM `%s`.`%s` LIMIT %d", path.getDatabase(), path.getTable(), limit);
```
yaojiefeng, line 90: so sql server doesn't use `"` to quote?

yaojiefeng (approved):
One question, LGTM

mysql-plugin/src/main/java/io/cdap/plugin/mysql/MysqlConnector.java
seanzhougoogle: Yeah, Mysql use ` instead " to quote

PR #168, opened 2021-08-13
CDAP-18354 put SID/Service name in connector properties

seanzhougoogle:


yaojiefeng (commented):
One question - is the Service name same as database? I saw this pr assumes they are the same thing

seanzhougoogle:
> One question - is the Service name same as database? I saw this pr assumes they are the same thing

service name is not same as database. But there should only be one database under one service name. Here we use `database` as field name to be consistent with what source is using.

yaojiefeng (approved):

PR #177, opened 2021-10-15
Add License file

avirmani:
Add License file to Database-Plugins

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,0 +1,232 @@
+Copyright © 2014-2021 Cask Data, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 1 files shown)


LICENSE.txt
```diff
@@ -0,0 +1,232 @@
+Copyright © 2014-2021 Cask Data, Inc.
```
snehalakshmisha, line 1: Should this still be Cask Data?

avirmani: All current licenses are that way, hence kept it the same.

greeshmaswaminathan, line 1: Is it intentional to have Cask copyright with Apache license 2 ?

avirmani: Same license in cdap and other repos as well - https://github.com/cdapio/cdap/blob/develop/LICENSE.txt

albertshau (approved):

PR #180, opened 2021-11-16
migrate connector DB specific sink plugins

itsankit-google:
[PLUGIN-968](https://cdap.atlassian.net/browse/PLUGIN-968)

### DIFF
--- a/database-commons/src/main/java/io/cdap/plugin/db/batch/config/AbstractDBSpecificSinkConfig.java
+++ b/database-commons/src/main/java/io/cdap/plugin/db/batch/config/AbstractDBSpecificSinkConfig.java
@@ -1,0 +1,118 @@
+/*
+ * Copyright © 2021 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.cdap.plugin.db.batch.config;
+
+import com.google.common.collect.Maps;
+import io.cdap.cdap.api.annotation.Description;
+import io.cdap.cdap.api.annotation.Macro;
+import io.cdap.cdap.api.annotation.Name;
+import io.cdap.cdap.api.plugin.PluginConfig;
+import io.cdap.plugin.common.Constants;
+import io.cdap.plugin.db.connector.AbstractDBSpecificConnectorConfig;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import javax.annotation.Nullable;
+
+/**
+ * Abstract Config for DB Specific Sink plugin
+ */
+public abstract class AbstractDBSpecificSinkConfig extends PluginConfig implements DatabaseSinkConfig {
+  public static final String TABLE_NAME = "tableName";
+  public static final String TRANSACTION_ISOLATION_LEVEL = "transactionIsolationLevel";
+
+  @Name(Constants.Reference.REFERENCE_NAME)
+  @Description(Constants.Reference.REFERENCE_NAME_DESCRIPTION)
+  private String referenceName;
+
+  @Name(TABLE_NAME)
+  @Description("Name of the database table to write to.")
+  @Macro
+  private String tableName;
+
+  @Override
+  public String getTableName() {
+    return tableName;
+  }
+
+  /**
+   * Adds escape characters (back quotes, double quotes, etc.) to the table name for
+   * databases with case-sensitive identifiers.
+   *
+   * @return tableName with leading and trailing escape characters appended.
+   * Default implementation returns unchanged table name string.
+   */
+  @Override
+  public String getEscapedTableName() {
+    return tableName;
+  }
+
+  @Override
+  public boolean canConnect() {
+    return !containsMacro(TABLE_NAME) && getConnection().canConnect();
+  }
+
+  @Nullable
+  @Override
+  public String getTransactionIsolationLevel() {
+    return null;
+  }
+
+  @Override
+  public List<String> getInitQueries() {
+    return Collections.emptyList();
+  }
+
+  protected abstract Map<String, String> getDBSpecificArguments();
+
+  protected abstract AbstractDBSpecificConnectorConfig getConnection();
+
+  @Override
+  public String getConnectionString() {
+    return getConnection().getConnectionString();
+  }
+
+  @Override
+  public Map<String, String> getConnectionArguments() {
+    Map<String, String> arguments = new HashMap<>();
+    arguments.putAll(Maps.fromProperties(getConnection().getConnectionArgumentsProperties()));
+    arguments.putAll(getDBSpecificArguments());
+    return arguments;
+  }
+
+  @Override
+  public String getJdbcPluginName() {
+    return getConnection().getJdbcPluginName();
+  }
+
+  @Override
+  public String getUser() {
+    return getConnection().getUser();
+  }
+
+  @Override
+  public String getPassword() {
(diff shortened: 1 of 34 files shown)


database-commons/src/main/java/io/cdap/plugin/db/batch/config/AbstractDBSpecificSinkConfig.java
```diff
@@ -0,0 +1,109 @@
...
+   * databases with case-sensitive identifiers.
+   *
+   * @return tableName with leading and trailing escape characters appended.
+   * Default implementation returns unchanged table name string.
+   */
+  public String getEscapedTableName() {
+    return tableName;
+  }
+
+  public boolean connectionParamsContainsMacro() {
```
yaojiefeng: This name is not intuitive, maybe call it `canConnect()`?

database-commons/src/main/java/io/cdap/plugin/db/batch/config/AbstractDBSpecificSinkConfig.java
```diff
@@ -0,0 +1,109 @@
...
+import java.util.List;
+import java.util.Map;
+import javax.annotation.Nullable;
+
+/**
+ * Abstract Config for DB Specific Sink plugin
+ */
+public abstract class AbstractDBSpecificSinkConfig extends PluginConfig implements DatabaseSinkConfig {
+  public static final String TABLE_NAME = "tableName";
+  public static final String TRANSACTION_ISOLATION_LEVEL = "transactionIsolationLevel";
```
yaojiefeng, line 38: where is this variable used?

database-commons/src/main/java/io/cdap/plugin/db/batch/config/AbstractDBSpecificSinkConfig.java
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  public boolean connectionParamsContainsMacro() {
+    return containsMacro(TABLE_NAME) || containsMacro(ConnectionConfig.DATABASE)
+      || getConnection().connectionParamsContainsMacro();
+  }
+
+  @Nullable
+  public String getTransactionIsolationLevel() {
+    return null;
```
yaojiefeng, line 74: this just return null and I don't see any field variable related to it, do we still need this method?

database-commons/src/main/java/io/cdap/plugin/db/connector/AbstractDBSpecificConnectorConfig.java
```diff
@@ -136,4 +136,9 @@ public int getPort() {
   public String getTransactionIsolationLevel() {
     return null;
   }
+
+  public boolean connectionParamsContainsMacro() {
```
yaojiefeng: can we just call this canConnect()?

mssql-plugin/src/main/java/io/cdap/plugin/mssql/SqlServerSink.java
```diff
@@ -126,19 +144,26 @@ protected FieldsValidator getFieldsValidator() {
...
+      return SqlServerUtil.composeDbSpecificArgumentsMap(instanceName, connection.getAuthenticationType(), null,
                                                          connectTimeout, columnEncryption, encrypt,
                                                          trustServerCertificate, workstationId, failoverPartner,
                                                          packetSize, queryTimeout);
     }
 
+    @Override
+    public String getConnectionString() {
+      return String
+        .format(SqlServerConstants.SQL_SERVER_CONNECTION_STRING_FORMAT, connection.getHost(), connection.getPort(),
```
yaojiefeng: nit - we normally split on the start of parameters instead of on `.`

yaojiefeng (commented):
Some small comments

database-commons/src/main/java/io/cdap/plugin/db/batch/config/AbstractDBSpecificSinkConfig.java
itsankit-google: Yeah, we need a default implementation of this method because its being called in the base class AbstractDBSink,
https://github.com/data-integrations/database-plugins/blob/873f86ad337403dadd824607d692ba3524d2dbce/database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java#L157

database-commons/src/main/java/io/cdap/plugin/db/connector/AbstractDBSpecificConnectorConfig.java
itsankit-google: Similar comment here as well, its being used by the base class AbstractDBSink, renaming here will lead to method not being recognised there, 
https://github.com/data-integrations/database-plugins/blob/873f86ad337403dadd824607d692ba3524d2dbce/database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java#L104

database-commons/src/main/java/io/cdap/plugin/db/batch/config/AbstractDBSpecificSinkConfig.java
itsankit-google: It is being used by the base class AbstractDBSink, renaming here will lead to method not being recognised there, 
https://github.com/data-integrations/database-plugins/blob/873f86ad337403dadd824607d692ba3524d2dbce/database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java#L104

database-commons/src/main/java/io/cdap/plugin/db/batch/config/AbstractDBSpecificSinkConfig.java
itsankit-google: Its being used in CloudSql MySql and Cloud PostgreSql sink can remove it for now as they don't use connections for now and add it in future PR.

yaojiefeng: If we will add it back later, let's just keep it there.

database-commons/src/main/java/io/cdap/plugin/db/batch/config/AbstractDBSpecificSinkConfig.java
yaojiefeng: Oh so this is an override from the parent class, can you mark `@Override` for all methods that inherit from the parent class?

yaojiefeng: I think it is also fine to just rename this method since it is not an api

database-commons/src/main/java/io/cdap/plugin/db/batch/config/AbstractDBSpecificSinkConfig.java
yaojiefeng: Please mark this method with `@Override`

database-commons/src/main/java/io/cdap/plugin/db/batch/config/DatabaseSinkConfig.java
```diff
@@ -0,0 +1,60 @@
...
+
+  /**
+   * @return the initial query to be run upon connecting to database
+   */
+  List<String> getInitQueries();
+
+  /**
+   * @return true if any of the connection parameters is macro, otherwise false
+   */
+  boolean connectionParamsContainsMacro();
```
yaojiefeng: I think it is fine to just rename all places with same name to `canConnect`, don't think it will result in any incompatibility.

mssql-plugin/src/main/java/io/cdap/plugin/mssql/SqlServerSink.java
yaojiefeng: This is not addressed

yaojiefeng (approved):
Some comments on the function name, otherwise LGTM

yaojiefeng:
Please also open JIRA for any issue found in the sink

mssql-plugin/src/main/java/io/cdap/plugin/mssql/SqlServerSink.java
itsankit-google: Acknowledged.

itsankit-google:
> Please also open JIRA for any issue found in the sink

done, https://cdap.atlassian.net/browse/PLUGIN-982

PR #183, opened 2021-11-23
PLUGIN-986

itsankit-google:
[PLUGIN-986](https://cdap.atlassian.net/browse/PLUGIN-986)

Postgress automatically converts unquoted column names to lowercase in the query which creates problem when the table itself contains columns in CamelCase or UpperCase. To deal with this, we pass the column names in quotes to the query. So, if a table contains column names in lowercase then inputSchema should also have column names in lowercase otherwise plugin validation will fail.

After fix:-
image

### DIFF
--- a/database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java
+++ b/database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java
@@ -305,9 +305,10 @@ public abstract class AbstractDBSink extends ReferenceBatchSink<StructuredRecord
           return;
         }
       }
-
-      try (PreparedStatement pStmt = connection.prepareStatement("SELECT * FROM " + dbSinkConfig.getEscapedTableName()
-                                                                   + " WHERE 1 = 0");
+      setColumnsInfo(inputSchema.getFields());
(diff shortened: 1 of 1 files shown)


database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java
```diff
@@ -305,18 +305,20 @@ private void validateSchema(FailureCollector collector, Class<? extends Driver>
...
       }
     } catch (SQLException e) {
       LOG.error("Exception while trying to validate schema of database table {} for connection {}.",
                 tableName, connectionString, e);
       collector.addFailure(
-        String.format("Exception while trying to validate schema of database table '%s' for connection '%s'.",
-                      tableName, connectionString), null).withStacktrace(e.getStackTrace());
+        String.format("Exception while trying to validate schema of database table '%s' for connection '%s' with %s",
+                      tableName, connectionString, e.getMessage()), null)
+        .withStacktrace(e.getStackTrace());
```
yaojiefeng: nit - indentation is off

yaojiefeng (approved):
Minor nit, LGTM

database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java
itsankit-google: fixed.

PR #185, opened 2021-12-01
Added logic to ensure batches are submitted to the engine every 1k records.

fernst:
This prevents the workers from running out of memory for large partitions.


sanjanasandeep (approved):
LGTM

snehalakshmisha (approved):
lgtm

database-commons/src/main/java/io/cdap/plugin/db/batch/sink/ETLDBOutputFormat.java
```diff
@@ -116,6 +117,13 @@ public void write(K key, V value) {
           try {
             key.write(getStatement());
             getStatement().addBatch();
+
+            // Submit a batch to the SQL engine every 10k records
+            // This is done to reduce memory usage in the worker, as processed records can now be GC'd.
+            numWrittenRecords++;
+            if (numWrittenRecords % 10000 == 0) {
```
tivv: Can we have this 10K configurable? In some rare cases 10K can be too high (E.g. in CDF our internal records are multimeg, customer may have it). Don't need ui, but a runtime argument would be nice

tivv (approved):

PR #187, opened 2021-12-03
set Double values to structured record for NUMERIC type when oracle return a double value

seanzhougoogle:
the value set should match the schema read here :
https://github.com/data-integrations/database-plugins/blob/345c77fced9a472312f7423f01d9b05451ec12d7/oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceSchemaReader.java#L83


itsankit-google (approved):
LGTM, please ensure oracle db connector in wrangler is also able to sample such data.

PR #190, opened 2021-12-03
use db specific way instead of generic way to get table schema for db specific plugin

seanzhougoogle:


database-commons/src/main/java/io/cdap/plugin/db/connector/AbstractDBSpecificConnector.java
```diff
@@ -144,4 +144,11 @@ protected void setConnectionProperties(Map<String, String> properties) {
     properties.put(ConnectionConfig.PASSWORD, rawProperties.get(ConnectionConfig.PASSWORD));
     properties.put(ConnectionConfig.CONNECTION_ARGUMENTS, rawProperties.get(ConnectionConfig.CONNECTION_ARGUMENTS));
   }
+
+  @Override
+  protected Schema getTableSchema(Connection connection, String database,
+                                  String schema, String table) throws SQLException {
+
+    return loadTableSchema(getConnection(),  getTableQuery(database, schema, table));
```
itsankit-google, line 152: Instead of changing definition of method `getTableQuery()`, can we do something like this?
`return loadTableSchema(getConnection(), getTableQuery(new DBSpecificPath(supportSchema(), database, schema, table));`

itsankit-google:
Please ensure these changes are forwarded to develop. So that both the branches are consistent.

database-commons/src/main/java/io/cdap/plugin/db/connector/AbstractDBSpecificConnector.java
seanzhougoogle: I meant to do this to de-couple the db path and table query, mainly they are independent logically.

itsankit-google: makes sense.

itsankit-google (approved):

PR #191, opened 2021-12-06
[PLUGIN-994] Adding setFetchSize to the database plugins

itsankit-google:
[PLUGIN-994](https://cdap.atlassian.net/browse/PLUGIN-994)


mysql-plugin/src/main/java/io/cdap/plugin/mysql/MysqlSource.java
```diff
@@ -173,6 +174,15 @@ public void validate(FailureCollector collector) {
     public boolean canConnect() {
       return super.canConnect() && !containsMacro(DATABASE);
     }
+
+    @Override
+    public Map<String, String> getConnectionArguments() {
+      Map<String, String> arguments = new HashMap<>(super.getConnectionArguments());
+      // If connected to MySQL > 5.0.2, and setFetchSize() > 0 on a statement,
```
tivv: I think this should go into `getDBSpecificArguments`. Also Please add a check for fetch size. If for whatever reason it's <=0 we should not add this argument

itsankit-google: done

database-commons/src/main/java/io/cdap/plugin/db/batch/source/AbstractDBSource.java
```diff
@@ -471,6 +475,11 @@ public boolean canConnect() {
         !containsMacro(ConnectionConfig.USER) && !containsMacro(ConnectionConfig.PASSWORD) &&
         !containsMacro(DBSourceConfig.DATABASE) && !containsMacro(DBSourceConfig.IMPORT_QUERY);
     }
+
+    @Override
+    public Integer getFetchSize() {
+      return 0;
```
tivv: Why don't we add proper fetchSize here?

tivv (commented):
Does this cover CloudSQL plugins?

tivv:
I see that CloudSQL plugins descent from io.cdap.plugin.db.batch.source.AbstractDBSource.DBSourceConfig and that one has hard coded "0" as fetch size

database-commons/src/main/java/io/cdap/plugin/db/batch/source/AbstractDBSource.java
itsankit-google: done

itsankit-google:
Yes, it does cover CloudSQL plugins as discussed offline.

tivv (approved):

PR #224, opened 2022-02-24
Use timestamp in gregorian calendar format.

sagarkapare:
Java by default uses `October 15, 1582` as a Gregorian cutover date. (See details here - https://docs.oracle.com/javase/7/docs/api/java/util/GregorianCalendar.html) Any timestamp created with time less than this cutover date is treated as Julian date. 

As a result when in database we have timestamp values less than gregorian cutover date as `0001-01-01 01:00:00` (inserted with sample query as `insert into hr.ts6 values (5, to_timestamp('0001-01-01 01.00.00.000000 AM','YYYY-MM-DD HH.MI.SS.FF6 AM'));`), when creating java.sql.Timestamp object, java thinks this is julian date for which corresponding gregorian date is `0000-12-30` which is what we see in CDAP for actual date `0001-01-01 01:00:00` in database.

This PR explicitly sets the gregorian date cutover to min value i.e. all timestamps will be treated as gregorian timestamps. With this change we get same value in cdap as that in db without any conversion.

If anyone is using julian dates for some reason in db, workaround for them would be convert them into gregorian before using in cdap. Most db provides sql functions for such conversion.

### DIFF
--- a/database-commons/src/main/java/io/cdap/plugin/util/DBUtils.java
+++ b/database-commons/src/main/java/io/cdap/plugin/util/DBUtils.java
@@ -37,9 +37,13 @@ import java.sql.DriverManager;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Types;
+import java.util.Calendar;
+import java.util.Date;
+import java.util.GregorianCalendar;
 import java.util.Hashtable;
 import java.util.List;
 import java.util.Objects;
+import java.util.TimeZone;
 import javax.annotation.Nullable;
 import javax.management.InstanceNotFoundException;
 import javax.management.IntrospectionException;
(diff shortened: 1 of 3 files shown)


avirmani (approved):

sagarkapare:
Verified the change on Oracle and mysql databases. 

For mysql db 
```sql
create table sakila.tsdatadt (
   id INT NOT NULL,
   ts datetime
);

insert into tsdatadt values (1, current_time);
insert into tsdatadt values (2, '2016-01-01 00:00:01')

insert into tsdatadt values(3, '1001-01-01 00:00:01');
insert into tsdatadt values(4, '0001-01-01 00:00:01');
```

Without fix, for id 4, preview outputs `0000-12-30T00:00:01Z[UTC]`
image

With fix, for id 4, output is `0001-01-01T08:00:01Z[UTC]` which matches with db value
image

PR #230, opened 2022-03-10
oracle sink schema support: PLUGIN-1146

jster1357:
add support for a user to load data into multiple schemas. Today, the sink only works with the login users schema.

### DIFF
--- a/database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java
+++ b/database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java
@@ -66,6 +66,7 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.Properties;
 import java.util.stream.Collectors;
+import javax.annotation.Nullable;
 
 /**
  * Sink that can be configured to export data to a database table.
@@ -100,7 +101,8 @@ public abstract class AbstractDBSink extends ReferenceBatchSink<StructuredRecord
         pipelineConfigurer, dbSinkConfig, ConnectionConfig.JDBC_PLUGIN_TYPE);
       if (driverClass != null && !dbSinkConfig.connectionParamsContainsMacro()) {
         FailureCollector collector = configurer.getFailureCollector();
-        validateSchema(collector, driverClass, dbSinkConfig.tableName, inputSchema);
+        validateSchema(collector, driverClass, dbSinkConfig.getTableName(), inputSchema,
+                dbSinkConfig.getDBSchemaName());
       }
     }
   }
@@ -109,8 +111,9 @@ public abstract class AbstractDBSink extends ReferenceBatchSink<StructuredRecord
   public void prepareRun(BatchSinkContext context) {
     String connectionString = dbSinkConfig.getConnectionString();
 
-    LOG.debug("tableName = {}; pluginType = {}; pluginName = {}; connectionString = {};",
-              dbSinkConfig.tableName,
+    LOG.debug("tableName = {}; schemaName = {}, pluginType = {}; pluginName = {}; connectionString = {};",
+              dbSinkConfig.getTableName(),
+              dbSinkConfig.getDBSchemaName(),
               ConnectionConfig.JDBC_PLUGIN_TYPE,
               dbSinkConfig.getJdbcPluginName(),
               connectionString);
@@ -123,7 +126,8 @@ public abstract class AbstractDBSink extends ReferenceBatchSink<StructuredRecord
     try {
       if (Objects.nonNull(outputSchema)) {
         FailureCollector collector = context.getFailureCollector();
-        validateSchema(collector, driverClass, dbSinkConfig.tableName, outputSchema);
+        validateSchema(collector, driverClass, dbSinkConfig.getTableName(),
+                outputSchema, dbSinkConfig.getDBSchemaName());
         collector.getOrThrowException();
       } else {
         outputSchema = inferSchema(driverClass);
@@ -141,8 +145,9 @@ public abstract class AbstractDBSink extends ReferenceBatchSink<StructuredRecord
     configAccessor.setInitQueries(dbSinkConfig.getInitQueries());
     configAccessor.getConfiguration().set(DBConfiguration.DRIVER_CLASS_PROPERTY, driverClass.getName());
     configAccessor.getConfiguration().set(DBConfiguration.URL_PROPERTY, connectionString);
-    configAccessor.getConfiguration().set(DBConfiguration.OUTPUT_TABLE_NAME_PROPERTY,
-                                          dbSinkConfig.getEscapedTableName());
+    String fullyQualifiedTableName = dbSinkConfig.getDBSchemaName() == null ? dbSinkConfig.getTableName()
+            : dbSinkConfig.getDBSchemaName() + "." + dbSinkConfig.getTableName();
+    configAccessor.getConfiguration().set(DBConfiguration.OUTPUT_TABLE_NAME_PROPERTY, fullyQualifiedTableName);
     configAccessor.getConfiguration().set(DBConfiguration.OUTPUT_FIELD_NAMES_PROPERTY, dbColumns);
     if (dbSinkConfig.getUser() != null) {
       configAccessor.getConfiguration().set(DBConfiguration.USERNAME_PROPERTY, dbSinkConfig.getUser());
@@ -184,6 +189,8 @@ public abstract class AbstractDBSink extends ReferenceBatchSink<StructuredRecord
 
   private Schema inferSchema(Class<? extends Driver> driverClass) {
     List<Schema.Field> inferredFields = new ArrayList<>();
+    String fullyQualifiedTableName = dbSinkConfig.getDBSchemaName() == null ? dbSinkConfig.getTableName()
+            : dbSinkConfig.getDBSchemaName() + "." + dbSinkConfig.getTableName();
     try {
       DBUtils.ensureJDBCDriverIsAvailable(driverClass, dbSinkConfig.getConnectionString(),
                                           dbSinkConfig.getJdbcPluginName());
@@ -194,7 +201,7 @@ public abstract class AbstractDBSink extends ReferenceBatchSink<StructuredRecord
         executeInitQueries(connection, dbSinkConfig.getInitQueries());
 
         try (Statement statement = connection.createStatement();
-             ResultSet rs = statement.executeQuery("SELECT * FROM " + dbSinkConfig.getEscapedTableName()
+             ResultSet rs = statement.executeQuery("SELECT * FROM " + fullyQualifiedTableName
                                                      + " WHERE 1 = 0")) {
           inferredFields.addAll(getSchemaReader().getSchemaFields(rs));
         }
@@ -232,6 +239,8 @@ public abstract class AbstractDBSink extends ReferenceBatchSink<StructuredRecord
   private void setResultSetMetadata() throws Exception {
     List<ColumnType> columnTypes = new ArrayList<>(columns.size());
     String connectionString = dbSinkConfig.getConnectionString();
+    String fullyQualifiedTableName = dbSinkConfig.getDBSchemaName() == null ? dbSinkConfig.getTableName()
+            : dbSinkConfig.getDBSchemaName() + "." + dbSinkConfig.getTableName();
 
     driverCleanup = DBUtils
       .ensureJDBCDriverIsAvailable(driverClass, connectionString, dbSinkConfig.getJdbcPluginName());
@@ -245,7 +254,7 @@ public abstract class AbstractDBSink extends ReferenceBatchSink<StructuredRecord
            // that can be used to construct DBRecord objects to sink to the database table.
            ResultSet rs = statement.executeQuery(String.format("SELECT %s FROM %s WHERE 1 = 0",
                                                                dbColumns,
-                                                               dbSinkConfig.getEscapedTableName()))
+                                                               fullyQualifiedTableName))
       ) {
         ResultSetMetaData resultSetMetadata = rs.getMetaData();
         columnTypes.addAll(getMatchedColumnTypeList(resultSetMetadata, columns));
@@ -279,15 +288,18 @@ public abstract class AbstractDBSink extends ReferenceBatchSink<StructuredRecord
   }
 
   private void validateSchema(FailureCollector collector, Class<? extends Driver> jdbcDriverClass, String tableName,
-                              Schema inputSchema) {
+                              Schema inputSchema, String dbSchemaName) {
     String connectionString = dbSinkConfig.getConnectionString();
+    String fullyQualifiedTableName = dbSinkConfig.getDBSchemaName() == null ? dbSinkConfig.getTableName()
+            : dbSinkConfig.getDBSchemaName() + "." + dbSinkConfig.getTableName();
 
     try {
       DBUtils.ensureJDBCDriverIsAvailable(jdbcDriverClass, connectionString, dbSinkConfig.getJdbcPluginName());
     } catch (IllegalAccessException | InstantiationException | SQLException e) {
       collector.addFailure(String.format("Unable to load or register JDBC driver '%s' while checking for " +
                                            "the existence of the database table '%s'.",
-                                         jdbcDriverClass, tableName), null).withStacktrace(e.getStackTrace());
+                                         jdbcDriverClass, fullyQualifiedTableName),
+              null).withStacktrace(e.getStackTrace());
       throw collector.getOrThrowException();
     }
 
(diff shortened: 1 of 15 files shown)


sagarkapare:
Compilation is failing for me with checkstyle errors on the branch, when I run command `mvn clean package -DskipTests`. Can you help fix these errors?

image

database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java
```diff
@@ -346,13 +416,20 @@ private void executeInitQueries(Connection connection, List<String> initQueries)
...
+    public static final String DB_SCHEMA_NAME = "dbSchemaName";
     public static final String TRANSACTION_ISOLATION_LEVEL = "transactionIsolationLevel";
 
     @Name(TABLE_NAME)
     @Description("Name of the database table to write to.")
     @Macro
     public String tableName;
 
+    @Name(DB_SCHEMA_NAME)
+    @Description("Name of the database table to write to.")
```
sagarkapare: nit: Name of the database schema to write to.

sagarkapare: This property will be visible to all the databases even if they do not support the notion of Schema. As far as I know SQL server and oracle has notion of schema, but MySQL does not. Should we add it to Oracle for now and then incrementally add to other databases such as SQL server, Postgres etc?

jster1357: Correct. I figure that all sinks will always default to the users schema. In many customers I've worked with in the past there may be one ETL user that has write access across many schemas/databases. The way everything is written, it will always default to just that users default DB. I figure we can expose over Oracle and then if there's requests we can expose to other sinks.

oracle-plugin/widgets/Oracle-batchsink.json
```diff
@@ -72,7 +72,13 @@
           "widget-type": "textbox",
           "label": "Table Name",
           "name": "tableName"
+        },
+        {
+          "widget-type": "textbox",
+          "label": "Schema Name",
+          "name": "dbSchemaName"
         }
+
```
sagarkapare: minor nit: remove new line.

database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java
```diff
@@ -141,8 +143,14 @@ public void prepareRun(BatchSinkContext context) {
...
-    configAccessor.getConfiguration().set(DBConfiguration.OUTPUT_TABLE_NAME_PROPERTY,
-                                          dbSinkConfig.getEscapedTableName());
+    if (dbSinkConfig.dbSchemaName == null) {
+      configAccessor.getConfiguration().set(DBConfiguration.OUTPUT_TABLE_NAME_PROPERTY,
+            dbSinkConfig.getEscapedTableName());
+    }
+    if (dbSinkConfig.dbSchemaName != null) {
+      configAccessor.getConfiguration().set(DBConfiguration.OUTPUT_TABLE_NAME_PROPERTY,
+              dbSinkConfig.getEscapedSchemaName() + "." + dbSinkConfig.getEscapedTableName());
+    }
```
sagarkapare: minor nit:
you can remove these if conditions by doing something similar to 
```java
    String fullyQualifiedTableName = dbSinkConfig.dbSchemaName == null ? dbSinkConfig.getEscapedTableName() 
      : dbSinkConfig.getEscapedSchemaName() + "." + dbSinkConfig.getEscapedTableName();
    configAccessor.getConfiguration().set(DBConfiguration.OUTPUT_TABLE_NAME_PROPERTY, fullyQualifiedTableName);
```

database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java
```diff
@@ -189,18 +197,33 @@ private Schema inferSchema(Class<? extends Driver> driverClass) {
...
       connectionProperties.putAll(dbSinkConfig.getConnectionArguments());
-      try (Connection connection = DriverManager.getConnection(dbSinkConfig.getConnectionString(),
-                                                               connectionProperties)) {
-        executeInitQueries(connection, dbSinkConfig.getInitQueries());
-
-        try (Statement statement = connection.createStatement();
-             ResultSet rs = statement.executeQuery("SELECT * FROM " + dbSinkConfig.getEscapedTableName()
-                                                     + " WHERE 1 = 0")) {
-          inferredFields.addAll(getSchemaReader().getSchemaFields(rs));
+      if (dbSinkConfig.dbSchemaName == null) {
```
sagarkapare: similarly here you dont need big if-else blocks. You can construct `fullyQualifiedTableName` once using ternary operator and use it everywhere below while forming sql queries.

database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java
```diff
@@ -240,18 +263,32 @@ private void setResultSetMetadata() throws Exception {
...
-           // Run a query against the DB table that returns 0 records, but returns valid ResultSetMetadata
-           // that can be used to construct DBRecord objects to sink to the database table.
-           ResultSet rs = statement.executeQuery(String.format("SELECT %s FROM %s WHERE 1 = 0",
-                                                               dbColumns,
-                                                               dbSinkConfig.getEscapedTableName()))
-      ) {
-        ResultSetMetaData resultSetMetadata = rs.getMetaData();
-        columnTypes.addAll(getMatchedColumnTypeList(resultSetMetadata, columns));
+
+      if (dbSinkConfig.dbSchemaName  == null) {
```
sagarkapare: Similar comment here. Would be good to avoid if-else blocks

database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java
```diff
@@ -290,35 +327,68 @@ private void validateSchema(FailureCollector collector, Class<? extends Driver>
...
-      executeInitQueries(connection, dbSinkConfig.getInitQueries());
-      try (ResultSet tables = connection.getMetaData().getTables(null, null, tableName, null)) {
-        if (!tables.next()) {
-          collector.addFailure(
-            String.format("Table '%s' does not exist.", tableName),
-            String.format("Ensure table '%s' is set correctly and that the connection string '%s' points " +
-                            "to a valid database.", tableName, connectionString))
-            .withConfigProperty(DBSinkConfig.TABLE_NAME);
-          return;
+    if (dbSinkConfig.dbSchemaName == null) {
```
sagarkapare: here as well.

database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java
```diff
@@ -364,10 +441,14 @@ protected String getEscapedTableName() {
       return tableName;
     }
 
+    protected String getEscapedSchemaName() {
+      return dbSchemaName;
```
sagarkapare, line 390: This is not really escaped schema name right? but we are returning as it is. I see the problem exists for the tableName too. We should just call this method as getSchemaName and add javadoc saying subclasses can override method to perform any escaping required.

sagarkapare (commented):
Few comments about code refactoring.

database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java
sagarkapare: oh you mean since we only change the widget json for oracle sink this wont be visible to other sinks? It might show up for other sinks too but not just in neat format. better to validate that.

oracle-plugin/widgets/Oracle-batchsink.json
jster1357: fixed

database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java
jster1357: way better way of writing it. made change.

database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java
jster1357: fixed

database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java
jster1357: fixed

database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java
jster1357: fixed

database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java
jster1357: correct. The only place I'm using that specific widget is in the oracle sink. I spot checked some others and I didn't see it creep in there...at least with the 1.6 build

database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java
```diff
@@ -184,24 +187,27 @@ public void initialize(BatchRuntimeContext context) throws Exception {
...
       connectionProperties.putAll(dbSinkConfig.getConnectionArguments());
-      try (Connection connection = DriverManager.getConnection(dbSinkConfig.getConnectionString(),
-                                                               connectionProperties)) {
-        executeInitQueries(connection, dbSinkConfig.getInitQueries());
 
-        try (Statement statement = connection.createStatement();
-             ResultSet rs = statement.executeQuery("SELECT * FROM " + dbSinkConfig.getEscapedTableName()
-                                                     + " WHERE 1 = 0")) {
-          inferredFields.addAll(getSchemaReader().getSchemaFields(rs));
+        try (Connection connection = DriverManager.getConnection(dbSinkConfig.getConnectionString(),
```
sagarkapare: Indentation seems off here. Seems like unintended change.

database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java
```diff
@@ -240,18 +248,19 @@ private void setResultSetMetadata() throws Exception {
...
-                                                               dbColumns,
-                                                               dbSinkConfig.getEscapedTableName()))
-      ) {
-        ResultSetMetaData resultSetMetadata = rs.getMetaData();
-        columnTypes.addAll(getMatchedColumnTypeList(resultSetMetadata, columns));
-      }
-    }
 
+
+        try (Statement statement = connection.createStatement();
```
sagarkapare: Same here.. keep indentation as it was before everywhere in the PR. So only change would be of tablename.

sagarkapare (commented):
PR looks good! Minor comments about correcting indentations in the PR as it was before so only changes in the PR should be schema name related.

sagarkapare:
File improvement here - https://cdap.atlassian.net/browse/PLUGIN-1146 to track this change.

database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java
jster1357: got it. Thanks. I know know how your doing your comparisons and what to look for specifically.

database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java
```diff
@@ -346,13 +356,20 @@ private void executeInitQueries(Connection connection, List<String> initQueries)
...
     @Name(TABLE_NAME)
     @Description("Name of the database table to write to.")
     @Macro
     public String tableName;
 
+    @Name(DB_SCHEMA_NAME)
+    @Description("Name of the database schema table to write to.")
+    @Macro
+    @Nullable
+    public String dbSchemaName;
```
itsankit-google, line 373: We should not add this config here in the common base class because even if it is not present in the widget json it will be displayed as generic property in other plugins using this class.
For eg, even if MySQL plugin does not have this property in the widget json still it shows in the plugin config as generic.
image

Instead of this what we can do is add this property only in config class of oracle sink only and have a `getDBSchemaName()` returning `null` as default but will be overriden by oracle sink config class.

database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java
```diff
@@ -364,10 +381,14 @@ protected String getEscapedTableName() {
       return tableName;
     }
 
+    protected String getSchemaName() {
```
itsankit-google: Can we name it is a `getDBSchemaName()` to avoid confusion with input schema?
And add a javadoc please.

database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java
itsankit-google: @seanzhougoogle WDYT?

database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java
jster1357: this is done

database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java
jster1357: @seanzhougoogle any thoughts here? I certainly don't want these to show up in other plugins. I thought that by restricting them at the widget level they wouldn't show up but they do.

seanzhougoogle: sorry, missed this thread. Other database (Postgres , SQLServer ) also supports schema. I think schema is a general concept and mysql is corner case (correct me if I'm wrong). You can of course hide the schema properties using 
```
 "widget-type": "hidden"
```
for those who don't support schema

database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java
```diff
@@ -346,13 +356,20 @@ private void executeInitQueries(Connection connection, List<String> initQueries)
...
     @Name(TABLE_NAME)
     @Description("Name of the database table to write to.")
     @Macro
     public String tableName;
 
+    @Name(DB_SCHEMA_NAME)
+    @Description("Name of the database schema of table.")
+    @Macro
+    @Nullable
+    public String dbSchemaName;
```
itsankit-google, line 373: Can we make it private? It should only be accessed with the help of getter.

database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java
```diff
@@ -100,7 +101,7 @@ public void configurePipeline(PipelineConfigurer pipelineConfigurer) {
         pipelineConfigurer, dbSinkConfig, ConnectionConfig.JDBC_PLUGIN_TYPE);
       if (driverClass != null && !dbSinkConfig.connectionParamsContainsMacro()) {
         FailureCollector collector = configurer.getFailureCollector();
-        validateSchema(collector, driverClass, dbSinkConfig.tableName, inputSchema);
+        validateSchema(collector, driverClass, dbSinkConfig.tableName, inputSchema, dbSinkConfig.dbSchemaName);
```
itsankit-google: Is there any specific reason to pass `dbSchemaName` as a param to the method? `validateSchema` can always use `dbSinkConfig.getDBSchemaName()`. Looks like the existing code has the same problem.

itsankit-google (commented):
Overall looks good, a minor comment.
Thanks.

database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java
jster1357: done

database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java
jster1357: The DatabaseSinkConfig Interface is public so there's a conflict in setting it private

database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java
jster1357: The newer dev branch has this change. we use the geTableName and getSchemaName methods. Make the change here?

itsankit-google: yeah, let's only use getter to access the configs

database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java
```diff
@@ -279,44 +286,47 @@ private void setResultSetMetadata() throws Exception {
   }
 
   private void validateSchema(FailureCollector collector, Class<? extends Driver> jdbcDriverClass, String tableName,
-                              Schema inputSchema) {
+                              Schema inputSchema, String dbSchemaName) {
     String connectionString = dbSinkConfig.getConnectionString();
+    String fullyQualifiedTableName = dbSinkConfig.dbSchemaName == null ? dbSinkConfig.getEscapedTableName()
```
itsankit-google: similar comment here, already have it as param so don't need to again use config to access them.
```
String escapedTableName = dbSinkConfig.getEscapedTableName();
String fullyQualifiedTableName = dbSchemaName == null ? escapedTableName
          : dbSchemaName + "." + escapedTableName;
```

itsankit-google (approved):
Lgtm. Thanks Justin :)

database-commons/src/main/java/io/cdap/plugin/db/batch/sink/AbstractDBSink.java
jster1357: ok. Modified the code to use methods

PR #252, opened 2022-04-28
[PLUGIN-1156] Check schema type for bigint object

JessieYee:
Jira: https://cdap.atlassian.net/browse/PLUGIN-1156

Convert BigInteger object into Long/BigDecimal based on the schema type

### DIFF
--- a/database-commons/src/main/java/io/cdap/plugin/db/DBRecord.java
+++ b/database-commons/src/main/java/io/cdap/plugin/db/DBRecord.java
@@ -156,6 +156,7 @@ public class DBRecord implements Writable, DBWritable, Configurable {
       if (schema.getType() == Schema.Type.LONG) {
         recordBuilder.set(field.getName(), ((BigInteger) o).longValueExact());
       } else {
+        // BigInteger won't have any fraction part and scale is 0
         recordBuilder.setDecimal(field.getName(), new BigDecimal((BigInteger) o, 0));
       }
     } else {
--- a/database-commons/src/main/java/io/cdap/plugin/db/DBRecord.java
+++ b/database-commons/src/main/java/io/cdap/plugin/db/DBRecord.java
@@ -152,13 +152,10 @@ public class DBRecord implements Writable, DBWritable, Configurable {
     } else if (o instanceof BigDecimal) {
       recordBuilder.setDecimal(field.getName(), (BigDecimal) o);
     } else if (o instanceof BigInteger) {
-      if (sqlType == Types.BIGINT && (resultSet.getMetaData().isSigned(columnIndex) ||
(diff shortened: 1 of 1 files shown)


database-commons/src/main/java/io/cdap/plugin/db/DBRecord.java
```diff
@@ -152,13 +152,10 @@ protected void setField(ResultSet resultSet, StructuredRecord.Builder recordBuil
...
     } else if (o instanceof BigInteger) {
-      if (sqlType == Types.BIGINT && (resultSet.getMetaData().isSigned(columnIndex) ||
-          resultSet.getMetaData().getPrecision(columnIndex) < 19)) {
-        //SQL BIGINT type is 64-bit long thus signed should be able to convert to long without losing precisions
-        // or UNSIGNED type is within the scope of signed long
+      Schema schema = field.getSchema();
+      if (schema.getType() == Schema.Type.LONG) {
         recordBuilder.set(field.getName(), ((BigInteger) o).longValueExact());
       } else {
-        // BigInteger won't have any fraction part and scale is 0
```
seanzhougoogle, line 161: no need to delete this comment

seanzhougoogle (approved):
LGTM with minor comment

database-commons/src/main/java/io/cdap/plugin/db/DBRecord.java
JessieYee: Update the comment

database-commons/src/main/java/io/cdap/plugin/db/DBRecord.java
```diff
@@ -152,10 +152,8 @@ protected void setField(ResultSet resultSet, StructuredRecord.Builder recordBuil
     } else if (o instanceof BigDecimal) {
       recordBuilder.setDecimal(field.getName(), (BigDecimal) o);
     } else if (o instanceof BigInteger) {
-      if (sqlType == Types.BIGINT && (resultSet.getMetaData().isSigned(columnIndex) ||
-          resultSet.getMetaData().getPrecision(columnIndex) < 19)) {
-        //SQL BIGINT type is 64-bit long thus signed should be able to convert to long without losing precisions
-        // or UNSIGNED type is within the scope of signed long
+      Schema schema = field.getSchema();
+      if (schema.getType() == Schema.Type.LONG) {
```
albertshau, line 156: it's possible the Schema is a nullable long, which is represented as a union of long and null, which means this type here will be a union. So you need to do something like:
```
schema = schema.isNullable() ? schema.getNonNullable() : schema;
if (schema.getType() == Schema.Type.LONG) {
...
```

PR #258, opened 2022-04-29
add isolation level to Oracle Plugin: PLUGIN-1178

jster1357:
Oracle specific changes. I'm going to work on the broader set of databases. 

https://cdap.atlassian.net/browse/PLUGIN-1178

### DIFF
--- a/oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleConnectorConfig.java
+++ b/oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleConnectorConfig.java
@@ -79,6 +79,12 @@ public class OracleConnectorConfig extends AbstractDBSpecificConnectorConfig {
   @Macro
   private String database;
 
+  @Name(OracleConstants.TRANSACTION_ISOLATION_LEVEL)
+  @Description("The transaction isolation level for the database session.")
+  @Macro
+  @Nullable
+  private String transactionIsolationLevel;
+
   @Override
   protected int getDefaultPort() {
     return 1521;
@@ -105,10 +111,17 @@ public class OracleConnectorConfig extends AbstractDBSpecificConnectorConfig {
     prop.put(INTERNAL_LOGON_PROPERTY, getRole());
     return prop;
   }
-  
+
   public String getTransactionIsolationLevel() {
-    return ROLE_NORMAL.equals(getRole()) ? null :
-      TransactionIsolationLevel.Level.TRANSACTION_READ_COMMITTED.name();
+    //if null default to the highest isolation level possible
+    if (transactionIsolationLevel == null) {
+      transactionIsolationLevel = TransactionIsolationLevel.Level.TRANSACTION_SERIALIZABLE.name();
+    }
+    //To solve the problem of ORA-08178: illegal SERIALIZABLE clause specified for user INTERNAL
+    //This ensures that the role is mapped to the right serialization level, even w/ incorrect user input
+    //if role is SYSDBA or SYSOP it will map to read_committed. else serialized
+    return (!getRole().equals(ROLE_NORMAL)) ? TransactionIsolationLevel.Level.TRANSACTION_READ_COMMITTED.name() :
+            TransactionIsolationLevel.Level.valueOf(transactionIsolationLevel).name();
   }
 
   @Override
--- a/oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSource.java
+++ b/oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSource.java
@@ -78,6 +78,7 @@ public class OracleSource extends AbstractDBSource<OracleSource.OracleSourceConf
     @Nullable
     @Description("Whether to use an existing connection.")
     private Boolean useConnection;
+
     @Name(NAME_CONNECTION)
     @Macro
     @Nullable
@@ -130,7 +131,7 @@ public class OracleSource extends AbstractDBSource<OracleSource.OracleSourceConf
 
     @Override
     public String getTransactionIsolationLevel() {
-      return getConnection().getTransactionIsolationLevel();
+      return connection.getTransactionIsolationLevel();
     }
   }
 
--- a/oracle-plugin/widgets/Oracle-batchsink.json
+++ b/oracle-plugin/widgets/Oracle-batchsink.json
@@ -89,6 +89,18 @@
           }
         },
         {
+          "widget-type": "select",
+          "label": "Transaction Isolation Level",
+          "name": "transactionIsolationLevel",
+          "widget-attributes": {
+            "values": [
+              "TRANSACTION_READ_COMMITTED",
+              "TRANSACTION_SERIALIZABLE"
+            ],
+            "default": "TRANSACTION_SERIALIZABLE"
+          }
+        },
+        {
           "name": "connectionType",
           "label": "Connection Type",
           "widget-type": "radio-group",
@@ -180,6 +192,18 @@
   "outputs": [],
   "filters": [
     {
+      "name": "showIsolationLevels",
+      "condition": {
+        "expression": "role  == 'normal'"
+      },
+      "show": [
+        {
+          "type": "property",
+          "name": "transactionIsolationLevel"
+        }
+      ]
+    },
+    {
       "name": "showConnectionProperties ",
       "condition": {
         "expression": "useConnection == false"
@@ -220,6 +244,10 @@
         {
           "type": "property",
           "name": "database"
+        },
+        {
+          "type": "property",
+          "name": "transactionIsolationLevel"
         }
       ]
     },
--- a/oracle-plugin/widgets/Oracle-batchsource.json
+++ b/oracle-plugin/widgets/Oracle-batchsource.json
@@ -89,6 +89,18 @@
           }
         },
         {
+          "widget-type": "select",
+          "label": "Transaction Isolation Level",
+          "name": "transactionIsolationLevel",
+          "widget-attributes": {
+            "values": [
+              "TRANSACTION_READ_COMMITTED",
+              "TRANSACTION_SERIALIZABLE"
+            ],
+            "default": "TRANSACTION_SERIALIZABLE"
+          }
+        },
+        {
           "name": "connectionType",
           "label": "Connection Type",
           "widget-type": "radio-group",
@@ -238,6 +250,18 @@
   ],
   "filters": [
     {
+      "name": "showIsolationLevels",
+      "condition": {
+        "expression": "role == 'normal'"
+      },
+      "show": [
+        {
+          "type": "property",
+          "name": "transactionIsolationLevel"
+        }
+      ]
+    },
+    {
       "name": "showConnectionProperties ",
       "condition": {
         "expression": "useConnection == false"
@@ -278,6 +302,10 @@
         {
           "type": "property",
           "name": "database"
+        },
+        {
+          "type": "property",
+          "name": "transactionIsolationLevel"
         }
       ]
     },
--- a/oracle-plugin/widgets/Oracle-connector.json
+++ b/oracle-plugin/widgets/Oracle-connector.json
@@ -97,6 +97,18 @@
               }
             ]
           }
+        },
+        {
+          "widget-type": "select",
+          "label": "Transaction Isolation Level",
+          "name": "transactionIsolationLevel",
+          "widget-attributes": {
+            "values": [
+              "TRANSACTION_READ_COMMITTED",
+              "TRANSACTION_SERIALIZABLE"
+            ],
+            "default": "TRANSACTION_SERIALIZABLE"
+          }
         }
       ]
     },
@@ -114,6 +126,20 @@
             "kv-delimiter": "=",
             "delimiter": ";"
           }
+        }
+      ]
+    }
+  ],
+  "filters" : [
+    {
+      "name": "showIsolationLevels",
+      "condition": {
+        "expression": "role  == 'normal'"
+      },
+      "show": [
+        {
+          "type": "property",
+          "name": "transactionIsolationLevel"
         }
       ]
     }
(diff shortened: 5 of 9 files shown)


jster1357:
I wanted to get an opinion on if we even needed an isolation level setting for a source plugin. Furthermore, would it make sense for an action plugin. It certainly makes sense for a sink plugin but I'm on the fence on Action and wether we need it for source.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSink.java
```diff
@@ -91,6 +91,12 @@ public static class OracleSinkConfig extends AbstractDBSpecificSinkConfig {
     @Nullable
     public Integer defaultBatchValue;
 
+    @Name(OracleConstants.ISOLATION_LEVEL)
+    @Description("The transaction isolation level for the database session.")
+    @Nullable
+    @Macro
+    public String transactionIsolationLevel;
```
itsankit-google: I think we can move this property to `OracleConnectorConfig` instead of independently in sink and source.

itsankit-google (commented):
Is there any specific reason to leave out `TRANSACTION_READ_UNCOMMITTED` & `TRANSACTION_REPEATABLE_READ`? In generic DB plugin, we do allow configuring them as well.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSink.java
seanzhougoogle: +1, we'd better add more descriptions here suggesting what transaction level to set for corresponding role. As I recall, you can not set arbitrary isolation level for sysdba or sysopr role, that's why we have https://github.com/data-integrations/database-plugins/blob/develop/oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleConnectorConfig.java#L109

seanzhougoogle: also since isolation level is a common connection property , if we are only focusing on Oracle now, let's add a TODO to move it to AbstractDBSpecificConnectorConfig/AbstractDBConnectorConfig in the future

jster1357: @seanzhougoogle I kept the isolation level changes just to the Oracle Plugin. I'm currently working on the broader change across the other DB's and I'm using AbstractDBSpecificConnectorConfig/AbstractDBConnectorConfig for those

jster1357:
I only exposed those specific isolation levels because Oracle doesn't support the other 2 levels. I had it originally but had errors when using REPEATED and UNCOMMITTED levels. I thought this approach would help save the user from themselves so to speak.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSink.java
jster1357: @seanzhougoogle  I can confirm on the sysdba user failing w/ serialized isolation levels. I recieve ORA-08178: illegal SERIALIZABLE clause specified for user INTERNAL

We should document this in the docs (plugin) that any selection w/ an internal user like sysdba or sysop will result in the isolation level being changed to read_committed.

itsankit-google: @jster1357 Could you please check if its possible to not display transaction_serializable in UI if sysdba or sysop is selected?

jster1357: @itsankit-google got it. I'm looking into using a filter to hide the transaction leves when sysop or sysdba is selected

jster1357:
I made the change from having the transaction level code at the source/sink level to the ConnectorConfig Class. Things work with the exception of when the roles are set for SYSDBA or SYSOP. I can run fine using the Normal role and I can get the isolation level to change just fine. When I have a SYSDBA role selected, I always get ORA-08178: illegal SERIALIZABLE clause specified for user INTERNAL even though I have logic to remap the isolation level when I see that role being used. It always grabs the serialized...I must be missing something simple here. I tired overriding the getTransactionIsolationLevel method but it's still not picking up the changes. 

I also made the change to hide the isolation field when a user selects the SYSDBA or SYSOPS radio button.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleConnectorConfig.java
```diff
@@ -79,6 +79,12 @@ public String getConnectionString() {
   @Macro
   private String database;
 
+  @Name(OracleConstants.TRANSACTION_ISOLATION_LEVEL)
+  @Description("The transaction isolation level for the database session.")
+  @Macro
+  @Nullable
+  public String transactionIsolationLevel;
```
itsankit-google: nit: why public? it should be private and only accessible through getters.

oracle-plugin/widgets/Oracle-batchsink.json
```diff
@@ -150,6 +150,18 @@
             "label": "Browse Database"
           }
         },
+        {
+          "widget-type": "select",
+          "label": "Transaction Isolation Level",
```
itsankit-google: Since, this property is included in `ConnectionConfig`, I would suggest to move it under `Connection` subheading below `Role` instead of `Basic`.

oracle-plugin/widgets/Oracle-batchsource.json
```diff
@@ -170,6 +170,18 @@
             "rows": "4"
           }
         },
+        {
+          "widget-type": "select",
+          "label": "Transaction Isolation Level",
```
itsankit-google: Similar comment here.

oracle-plugin/widgets/Oracle-connector.json
```diff
@@ -59,6 +59,18 @@
           "label": "SID/Service Name/TNS Connect Descriptor",
           "description": "Oracle connection point (Database name, Service name, or TNS Connect Descriptor)",
           "name": "database"
+        },
+        {
+          "widget-type": "select",
+          "label": "Transaction Isolation Level",
```
itsankit-google: Similarly here, does it make sense to move the position of this widget below `Role` widget?

itsankit-google (commented):
Some comments.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSink.java
```diff
@@ -102,11 +102,6 @@ protected Map<String, String> getDBSpecificArguments() {
       return ImmutableMap.of(OracleConstants.DEFAULT_BATCH_VALUE, String.valueOf(defaultBatchValue));
     }
 
-    @Override
-    public String getTransactionIsolationLevel() {
```
itsankit-google: This method should not be removed.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSource.java
```diff
@@ -128,10 +129,6 @@ public void validate(FailureCollector collector) {
       super.validate(collector);
     }
 
-    @Override
-    public String getTransactionIsolationLevel() {
```
itsankit-google, line 132: Similarly here, removing this overriden method will make it return `null` always. https://github.com/data-integrations/database-plugins/blob/f2b0a7c0801b3ceedd9092a208d64b7fa7ffd03c/database-commons/src/main/java/io/cdap/plugin/db/batch/config/AbstractDBSpecificSourceConfig.java#L207

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSink.java
itsankit-google: Otherwise it will use the default implementation which returns `null` always.
https://github.com/data-integrations/database-plugins/blob/0c5d869af7fd7a13924de889003ad6b8ea1b5dc1/database-commons/src/main/java/io/cdap/plugin/db/batch/config/AbstractDBSpecificSinkConfig.java#L85

itsankit-google:
> I made the change from having the transaction level code at the source/sink level to the ConnectorConfig Class. Things work with the exception of when the roles are set for SYSDBA or SYSOP. I can run fine using the Normal role and I can get the isolation level to change just fine. When I have a SYSDBA role selected, I always get ORA-08178: illegal SERIALIZABLE clause specified for user INTERNAL even though I have logic to remap the isolation level when I see that role being used. It always grabs the serialized...I must be missing something simple here. I tired overriding the getTransactionIsolationLevel method but it's still not picking up the changes.
> 
> I also made the change to hide the isolation field when a user selects the SYSDBA or SYSOPS radio button.

I think it was because of removal of `getTransactionIsolationLevel()` overridden methods from `OracleSourceConfig` and `OracleSinkConfig` which made it always return `null` and setting the default `TRANSACTION_SERIALIZABLE` isolation level.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleConnectorConfig.java
jster1357: good point. No need to be public

oracle-plugin/widgets/Oracle-batchsource.json
jster1357: fixed

oracle-plugin/widgets/Oracle-batchsink.json
jster1357: so any configuration I add within the ConnectorConfig class would need to be applied to the connector section of a widget? still learning how these plugins and java classes mesh.

jster1357: fixed

itsankit-google: > so any configuration I add within the ConnectorConfig class would need to be applied to the connector section of a widget? still learning how these plugins and java classes mesh.

It is not mandatory but generally the pattern followed in all plugins. These fields should only be shown when `UseConnection` set to `False`.

jster1357:
I've made the changes and everything looks good on my end. I've tested the following situations:
- Normal Role with null isolation level, job runs with SERIALIZABLE
- Normal Role with SERIALIZABLE Level, job runs with SERIALIZABLE
- Normal Role with READ_COMMITTED level, job runs with READ_COMMITTED
- SYSDBA Role, UI Hides isolation level selection, job runs with READ_COMMITTED
- import pipeline with SYSDBA role and SERIALIZABLE level, job runs with READ_COMMITTED

I don't think we need to expose the isolation level w/ a predefined connector...that should be done on a table by table basis.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSource.java
jster1357: fixed

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSink.java
jster1357: fixed

oracle-plugin/widgets/Oracle-connector.json
jster1357: fixed

oracle-plugin/widgets/Oracle-batchsink.json
jster1357: got. this is now fixed

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleConnectorConfig.java
jster1357: fixed

itsankit-google:
> I don't think we need to expose the isolation level w/ a predefined connector...that should be done on a table by table basis.

Predefined connections also act like a source for sampling the tables through wrangler.
@seanzhougoogle WDYT?

oracle-plugin/widgets/Oracle-connector.json
```diff
@@ -118,5 +130,6 @@
       ]
     }
   ],
+  "Filters" : [],
```
itsankit-google: Instead of empty, we need to add a similar filter for isolation levels as in source and sink.

jster1357: I added the filter but I don't see the ISOLATION LEVEL showing up. I didn't even see it prior to adding the filter, which was why I originally left it off.

itsankit-google: yeah looks like there is some issue in the UI not rendering it correctly? Is it only with `select` widget? Did you try changing the widget type?

oracle-plugin/widgets/Oracle-connector.json
```diff
@@ -118,5 +130,19 @@
       ]
     }
   ],
+  "Filters" : [
```
itsankit-google: nit: `Filters`->`filters`

itsankit-google (approved):

itsankit-google:
LGTM, one minor comment. Thanks Justin :)

oracle-plugin/widgets/Oracle-connector.json
itsankit-google: I was able to verify your changes, looks like it is a UI bug when there are 2 connectors, one in system scope and other in user scope, UI only renders system scope one.

oracle-plugin/widgets/Oracle-connector.json
jster1357: fixed

oracle-plugin/widgets/Oracle-connector.json
jster1357: I tired the radio button but it didn't render. I thought the filter capitalization error would fix it but no dice.

itsankit-google: Opened a JIRA for UI team - https://cdap.atlassian.net/browse/PLUGIN-1248

itsankit-google:
> > I don't think we need to expose the isolation level w/ a predefined connector...that should be done on a table by table basis.
> 
> Predefined connections also act like a source for sampling the tables through wrangler. @seanzhougoogle WDYT?

Discussed offline with @seanzhougoogle, will need to live with this limitation because it is a connection property.

PR #273, opened 2022-06-21
add e2e test for mysql plugin

jockech:
Create the e2e test for mysql plugin.

snapshot:
image

result:
image

### DIFF
--- a/mysql-plugin/src/e2e-test/resources/errorMessage.properties
+++ b/mysql-plugin/src/e2e-test/resources/errorMessage.properties
@@ -1,0 +1,1 @@
+validationSuccessMessage=No errors found.
--- a/mysql-plugin/src/e2e-test/resources/pluginParameters.properties
+++ b/mysql-plugin/src/e2e-test/resources/pluginParameters.properties
@@ -1,0 +1,12 @@
+driver=mysql
+host=34.105.111.83
+username=dummy
+password=dummy
+database=sakila
+port=3306
+selectQuery=select * from ${table}
+sourceRef=source
+targetRef=target
+sourceTable=sourceTable
+targetTable=targetTable
+outputSchema=[{"key":"id","value":"int"},{"key":"lastName","value":"string"}]
--- a/pom.xml
+++ b/pom.xml
@@ -62,7 +62,7 @@
     <!-- version properties -->
     <cdap.version>6.8.0-SNAPSHOT</cdap.version>
     <cdap.plugin.version>2.10.0-SNAPSHOT</cdap.plugin.version>
-    <guava.version>27.0.1-jre</guava.version>
+    <guava.version>13.0.1</guava.version>
     <hadoop.version>2.3.0</hadoop.version>
     <hsql.version>2.2.4</hsql.version>
     <junit.version>4.13</junit.version>
--- a/pom.xml
+++ b/pom.xml
@@ -62,7 +62,7 @@
     <!-- version properties -->
     <cdap.version>6.8.0-SNAPSHOT</cdap.version>
     <cdap.plugin.version>2.10.0-SNAPSHOT</cdap.plugin.version>
-    <guava.version>13.0.1</guava.version>
+    <guava.version>27.0.1-jre</guava.version>
     <hadoop.version>2.3.0</hadoop.version>
     <hsql.version>2.2.4</hsql.version>
     <junit.version>4.13</junit.version>
--- a/mysql-plugin/pom.xml
+++ b/mysql-plugin/pom.xml
@@ -27,6 +27,11 @@
   <artifactId>mysql-plugin</artifactId>
   <modelVersion>4.0.0</modelVersion>
 
+  <properties>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <testSourceLocation>${project.basedir}/src/test/java/</testSourceLocation>
+  </properties>
+
   <dependencies>
     <dependency>
       <groupId>io.cdap.cdap</groupId>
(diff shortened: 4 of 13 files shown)


mysql-plugin/src/e2e-test/java/io/cdap/plugin/MysqlConnector.java
```diff
@@ -0,0 +1,98 @@
...
+    // Insert dummy data.
+    statement.executeUpdate("INSERT INTO " + sourceTable + " (id, lastName)" +
+                       "VALUES (1, 'Simpson')");
+    statement.executeUpdate("INSERT INTO " + sourceTable + " (id, lastName)" +
+                       "VALUES (2, 'McBeal')");
+    statement.executeUpdate("INSERT INTO " + sourceTable + " (id, lastName)" +
+                       "VALUES (3, 'Flinstone')");
+
+
+    String createTargetTableQuery = "CREATE TABLE IF NOT EXISTS " + targetTable +
```
itsankit-google: Can we separate out setting up of `targetTable` in another method?
If `sourceTable` and `targetTable` are same we might end up with unexpected results.

mysql-plugin/src/e2e-test/java/io/cdap/plugin/MysqlConnector.java
```diff
@@ -0,0 +1,96 @@
...
+    statement.executeUpdate(createTargetTableQuery);
+
+    // Truncate table to clean the data of last failure run.
+    String truncateTargetTableQuery = "TRUNCATE TABLE " + targetTable;
+    statement.executeUpdate(truncateTargetTableQuery);
+    statement.close();
+    connect.close();
+  }
+
+  public static void dropTables(String sourceTable, String targetTable) throws SQLException, ClassNotFoundException {
```
itsankit-google: Can we pass the input parameter as list of tables to drop instead of individual tables?

pom.xml
```diff
@@ -62,7 +62,7 @@
     <!-- version properties -->
     <cdap.version>6.8.0-SNAPSHOT</cdap.version>
     <cdap.plugin.version>2.10.0-SNAPSHOT</cdap.plugin.version>
-    <guava.version>13.0.1</guava.version>
+    <guava.version>27.0.1-jre</guava.version>
```
itsankit-google: why is this change needed?

itsankit-google (commented):
Few comments,

Are we expecting jdbc drivers to be already installed?

pom.xml
jockech: I have got `java.lang.NoSuchMethodError: 'java.util.stream.Collectorcom.google.common.collect.ImmutableList.toImmutableList()'`
and checked the [stackoverflow](https://stackoverflow.com/questions/65973443/java-lang-nosuchmethoderror-java-util-stream-collector-com-google-common-colle) and decided to syncup with [goolge-cloud](https://github.com/data-integrations/google-cloud/blob/de29c557c38a2322c807dbdf7061342c72b333f7/pom.xml#L90)

mysql-plugin/src/e2e-test/java/io/cdap/plugin/MysqlConnector.java
```diff
@@ -0,0 +1,96 @@
...
+import java.sql.SQLException;
+import java.sql.Statement;
+
+/**
+ *  MySQL connector
+ */
+@Plugin(type = Connector.PLUGIN_TYPE)
+@Name(MysqlConnector.NAME)
+@Category("MySQL")
+@Description("Connection to access data in MySQL.")
```
itsankit-google: Why are these annotations needed? it is not a CDAP Connector plugin.

mysql-plugin/src/e2e-test/java/io/cdap/plugin/MysqlConnector.java
jockech: > Few comments,
> 
> Are we expecting jdbc drivers to be already installed?

Yes. Don't think we could install the driver in the test.

mysql-plugin/src/e2e-test/resources/errorMessage.properties
```diff
@@ -0,0 +1 @@
+validationSuccessMessage=No errors found.
```
rmstar, line 1: where is this used?

rmstar (commented):
Do we need [pluginDataCyAttributes.properties](https://github.com/data-integrations/database-plugins/pull/273/files#diff-078674b9e17146b5b1fef0e32bde49c6d88c12943af69ee3f1d64ef051e03fab). Looks like it's empty?

mysql-plugin/src/e2e-test/java/io/cdap/plugin/MysqlConnector.java
```diff
@@ -0,0 +1,77 @@
...
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+
+/**
+ *  MySQL connector
+ */
+public class MysqlConnector {
```
itsankit-google: Can we rename it to `MysqlClient` or `MysqlHelper`? 
Want to avoid conflict with CDAP connector terminology.

itsankit-google (approved):
One minor comment, otherwise LGTM

mysql-plugin/src/e2e-test/resources/pluginParameters.properties
```diff
@@ -0,0 +1,12 @@
+driver=mysql
+host=34.105.111.83
+username=dummy
+password=dummy
+database=sakila
+port=3306
+selectQuery=select * from
```
itsankit-google: nit: `selectQuery=select * from ${table}` looks more readable.
Can replace `${table}` with `sourceTable` in test setup hook.

itsankit-google (approved):

PR #277, opened 2022-07-28
Refactor getTableQuery() in DB connectors to use getTableName()

ellliao:
# Refactor getTableQuery() in DB connectors to use getTableName()

## Description
Fixed `getTableQuery()` function in all database connectors within this repository to refer to an added `getTableName()` function instead of using the previous hardcoded table name within each connector's `getTableQuery()` SQL query. The way in which this was refactored should be backwards compatible (so shouldn't affect the behavior of any other functions/classes that rely on `getTableQuery()` or extend any of the connectors changed).

*This PR has been merged into PR #285*

## PR Type
- [ ] Bug Fix
- [ ] Feature
- [ ] Build Fix
- [ ] Testing
- [x] General Improvement
- [ ] Cherry Pick

## Links
Jira: none


adinilfeld (commented):
Looks great to me!

adinilfeld (approved):

PR #281, opened 2022-08-09
Change e2e tests run condition

sawantpritam:
@itsankit-google @rmstar 

Changed e2e-tests execute condition (Same as hydrator repo)

If there are changes inside tests of a particular module we will execute all tests of that module & required tests of other module.

If there are no changes in e2e-test execute required tests of all module.


sawantpritam:
Added below changes as well in this PR:
1. Removed e2e-test profile from mysql-plugin module's pom.xml (not required as we have profile in main pom.xml)
2. Added Required runner for mysql-plugin

itsankit-google (approved):

PR #282, opened 2022-08-12
[CDAP-19417] Fix classloading conflicts for DB Plugins

snehalakshmisha:
Jira: https://cdap.atlassian.net/browse/CDAP-19417

If MySQL plugin is used with other DB plugins in the same pipeline, it fails with an error `io.cdap.plugin.db.DBRecord cannot be cast to io.cdap.plugin.db.DBRecord`. This PR fixes the class loading error by making each individual plugin export its own contents

### DIFF
--- a/aurora-mysql-plugin/pom.xml
+++ b/aurora-mysql-plugin/pom.xml
@@ -90,6 +90,35 @@
         <groupId>io.cdap</groupId>
         <artifactId>cdap-maven-plugin</artifactId>
       </plugin>
+      <plugin>
+        <groupId>org.apache.felix</groupId>
+        <artifactId>maven-bundle-plugin</artifactId>
+        <version>5.1.2</version>
+        <extensions>true</extensions>
+        <configuration>
+          <instructions>
+            <_exportcontents>
+              io.cdap.plugin.auroradb.mysql.*;
+              io.cdap.plugin.db.batch.source.*;
+              io.cdap.plugin.db.batch.sink.*;
+              org.apache.commons.lang;
+              org.apache.commons.logging.*;
+              org.codehaus.jackson.*
+            </_exportcontents>
+            <Embed-Dependency>*;inline=false;scope=compile</Embed-Dependency>
+            <Embed-Transitive>true</Embed-Transitive>
+            <Embed-Directory>lib</Embed-Directory>
+          </instructions>
+        </configuration>
+        <executions>
+          <execution>
+            <phase>package</phase>
+            <goals>
+              <goal>bundle</goal>
+            </goals>
+          </execution>
+        </executions>
+      </plugin>
     </plugins>
   </build>
 </project>
--- a/mssql-plugin/pom.xml
+++ b/mssql-plugin/pom.xml
@@ -100,20 +100,28 @@
       <plugin>
         <groupId>org.apache.felix</groupId>
         <artifactId>maven-bundle-plugin</artifactId>
+        <version>5.1.2</version>
+        <extensions>true</extensions>
         <configuration>
           <instructions>
             <_exportcontents>
-              io.cdap.plugin.*;
-              org.apache.commons.lang;
-              org.apache.commons.logging.*;
-              org.codehaus.jackson.*;
-              com.microsoft.aad.*
+              io.cdap.plugin.mssql.*;
+              io.cdap.plugin.db.batch.source.*;
+              io.cdap.plugin.db.batch.sink.*;
             </_exportcontents>
             <Embed-Dependency>*;inline=false;scope=compile</Embed-Dependency>
             <Embed-Transitive>true</Embed-Transitive>
             <Embed-Directory>lib</Embed-Directory>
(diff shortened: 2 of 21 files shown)


mysql-plugin/src/main/java/io/cdap/plugin/mysql/MysqlDBRecord.java
```diff
@@ -0,0 +1,35 @@
...
+import io.cdap.cdap.api.data.format.StructuredRecord;
+import io.cdap.plugin.db.ColumnType;
+import io.cdap.plugin.db.DBRecord;
+
+import java.util.List;
+
+/**
+ * Writable class for MySQL Source/Sink.
+ * TODO: Adding this class is part of a quick fix for CDAP-19417.
+ * The correct way to fix the jira is to use the top level input format to return StructuredRecord.
```
snehalakshmisha: @chtyim Please correct this comment

seanzhougoogle: The correct fix is :
1. in AbstractDBSource , instead of using DataDrivenETLDBInputFormat, we should use a new input format which directly implements InputFormat<LongWritable,  StructuredRecord> and delegates its functions to DataDrivenETLDBInputFormat
2. Each db specific plugin should export its own packages without export the dependent packages from database-common project

seanzhougoogle (approved):

mysql-plugin/src/main/java/io/cdap/plugin/mysql/MysqlDBRecord.java
seanzhougoogle: two more thing
1. implement DB specific input format to extend the new input format ( so that we don't export the same new input format in every db specific plugin)
2. do similar for output format

seanzhougoogle: Meanwhile please add a jira for the TODO and put the Jira in the comments.

chtyim: Since we are fixing it. I think it is better to also make the effort to move the `exportcontent` in the pom.xml to individual children modules and have a more specific package to export.

snehalakshmisha: Fixed the export in this PR. Created a jira for the other part - https://cdap.atlassian.net/browse/PLUGIN-1346

aurora-mysql-plugin/pom.xml
```diff
@@ -90,6 +90,35 @@
...
+        <version>5.1.2</version>
+        <extensions>true</extensions>
+        <configuration>
+          <instructions>
+            <_exportcontents>
+              io.cdap.plugin.auroradb.mysql.*;
+              io.cdap.plugin.db.batch.source.*;
+              io.cdap.plugin.db.batch.sink.*;
+              org.apache.commons.lang;
+              org.apache.commons.logging.*;
```
chtyim, line 105: What need to export Apache common and Jackson?

snehalakshmisha: I simply retained exports from the parent pom. Let me look into this

snehalakshmisha: I removed it for mysql, mssql and postgres after verifying but it seems to be coming from the very first PR for each of these plugins - https://github.com/data-integrations/database-plugins/commit/4fb0a7e4469de89bcf25f1cd27a64a31e31c6c09#diff-852a38ca0def089168225bcc77b964c0ec801a99184863eb1b1bc651fa6bc4b0R94

mssql-plugin/pom.xml
```diff
@@ -100,20 +100,29 @@
...
           <instructions>
             <_exportcontents>
-              io.cdap.plugin.*;
-              org.apache.commons.lang;
-              org.apache.commons.logging.*;
-              org.codehaus.jackson.*;
+              io.cdap.plugin.mssql.*;
+              io.cdap.plugin.db.batch.source.*;
+              io.cdap.plugin.db.batch.sink.*;
               com.microsoft.aad.*
```
chtyim: Do you know why need this is exported?

chtyim (commented):
Most likely the old code was doing all the export wrong. It might worth to re-test them with a more specific set of exports.

mssql-plugin/pom.xml
snehalakshmisha: Yet another unwanted export. Combing through all plugins to confirm all exports

aurora-mysql-plugin/pom.xml
snehalakshmisha: I tested 4 plugins, created a jira to test the same for remaining 12 plugins - https://cdap.atlassian.net/browse/PLUGIN-1347

chtyim (approved):

PR #285, opened 2022-08-22
Add Sampling Options for Multiple Database Connectors

ellliao:
# Add Sampling Options for Multiple Database Connectors

## Description
Added `random` sampling option and unit tests to PostgreSQL, MySQL, OracleDB, and SQLServer connections in Wrangler, checking for `sampleType` in sample properties to determine which option (first [default] or random) to use.

Also refactored all DB connectors in `database-plugins` to use a `getTableName()`function within `getTableQuery()` instead of the previous hardcoded SQL queries (merged from #277). The refactor should be backwards-compatible.

Related PRs:
- [cdap #14495](https://github.com/cdapio/cdap/pull/14495)
- [google-cloud #1084](https://github.com/data-integrations/google-cloud/pull/1084)
- [wrangler #573](https://github.com/data-integrations/wrangler/pull/573)
- [cdap-ui #603](https://github.com/cdapio/cdap-ui/pull/603)

## PR Type
- [ ] Bug Fix
- [x] Feature
- [ ] Build Fix
- [ ] Testing
- [ ] General Improvement
- [ ] Cherry Pick

## Links
Jira: none

### DIFF
--- a/database-commons/src/main/java/io/cdap/plugin/db/ConnectionConfigAccessor.java
+++ b/database-commons/src/main/java/io/cdap/plugin/db/ConnectionConfigAccessor.java
@@ -111,5 +111,4 @@ public class ConnectionConfigAccessor {
   public Configuration getConfiguration() {
     return configuration;
   }
-
 }
--- a/database-commons/src/main/java/io/cdap/plugin/db/connector/AbstractDBSpecificConnector.java
+++ b/database-commons/src/main/java/io/cdap/plugin/db/connector/AbstractDBSpecificConnector.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2021 Cask Data, Inc.
+ * Copyright © 2021-2022 Cask Data, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not
  * use this file except in compliance with the License. You may obtain a copy of
@@ -23,13 +23,13 @@ import io.cdap.cdap.etl.api.batch.BatchConnector;
 import io.cdap.cdap.etl.api.connector.ConnectorContext;
 import io.cdap.cdap.etl.api.connector.ConnectorSpecRequest;
 import io.cdap.cdap.etl.api.connector.SampleRequest;
+import io.cdap.cdap.etl.api.connector.SampleType;
 import io.cdap.plugin.common.ConfigUtil;
 import io.cdap.plugin.common.SourceInputFormatProvider;
 import io.cdap.plugin.common.db.AbstractDBConnector;
 import io.cdap.plugin.common.db.DBConnectorPath;
 import io.cdap.plugin.common.util.ExceptionUtils;
 import io.cdap.plugin.db.CommonSchemaReader;
-import io.cdap.plugin.db.ConnectionConfig;
 import io.cdap.plugin.db.ConnectionConfigAccessor;
 import io.cdap.plugin.db.SchemaReader;
 import io.cdap.plugin.db.batch.source.DataDrivenETLDBInputFormat;
@@ -44,9 +44,12 @@ import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Statement;
 import java.util.Map;
+import java.util.UUID;
+import javax.annotation.Nullable;
 
 /**
  * An Abstract DB Specific Connector those specific DB connectors can inherits
+ *
  * @param <T> the Record type that specific DB Record Reader may return while sample the data with InputFormat
  */
 public abstract class AbstractDBSpecificConnector<T extends DBWritable> extends AbstractDBConnector
@@ -63,7 +66,7 @@ public abstract class AbstractDBSpecificConnector<T extends DBWritable> extends 
 
   protected abstract Class<? extends DBWritable> getDBRecordType();
 
-  protected SchemaReader getSchemaReader() {
+  protected SchemaReader getSchemaReader(String sessionID) {
     return new CommonSchemaReader();
   }
 
@@ -84,14 +87,16 @@ public abstract class AbstractDBSpecificConnector<T extends DBWritable> extends 
     ConnectionConfigAccessor connectionConfigAccessor = new ConnectionConfigAccessor();
     if (config.getUser() == null && config.getPassword() == null) {
       DBConfiguration.configureDB(connectionConfigAccessor.getConfiguration(), driverClass.getName(),
-                     getConnectionString(path.getDatabase()));
+                                  getConnectionString(path.getDatabase()));
     } else {
       DBConfiguration.configureDB(connectionConfigAccessor.getConfiguration(), driverClass.getName(),
-                     getConnectionString(path.getDatabase()), config.getUser(), config.getPassword());
+                                  getConnectionString(path.getDatabase()), config.getUser(), config.getPassword());
     }
-    String tableQuery = getTableQuery(path.getDatabase(), path.getSchema(), path.getTable(), request.getLimit());
+    String sessionID = generateSessionID();
+    String tableQuery = getTableQuery(path.getDatabase(), path.getSchema(), path.getTable(), request.getLimit(),
+      request.getProperties().get("sampleType"), request.getProperties().get("strata"), sessionID);
     DataDrivenETLDBInputFormat.setInput(connectionConfigAccessor.getConfiguration(), getDBRecordType(),
-                                        tableQuery, null, false);
+      tableQuery, null, false);
     connectionConfigAccessor.setConnectionArguments(Maps.fromProperties(config.getConnectionArgumentsProperties()));
     connectionConfigAccessor.getConfiguration().setInt(MRJobConfig.NUM_MAPS, 1);
     Map<String, String> additionalArguments = config.getAdditionalArguments();
@@ -99,41 +104,81 @@ public abstract class AbstractDBSpecificConnector<T extends DBWritable> extends 
       connectionConfigAccessor.getConfiguration().set(argument.getKey(), argument.getValue());
     }
     try {
-      connectionConfigAccessor.setSchema(loadTableSchema(getConnection(path),  tableQuery).toString());
+      Long timeoutMs = request.getTimeoutMs();
+      Integer timeoutSec = timeoutMs != null ? (int) (timeoutMs / 1000) : null;
+      connectionConfigAccessor
+        .setSchema(loadTableSchema(getConnection(path), tableQuery, timeoutSec, sessionID).toString());
     } catch (SQLException e) {
       throw new IOException(String.format("Failed to get table schema due to: %s.",
-                                          ExceptionUtils.getRootCauseMessage(e)), e);
+        ExceptionUtils.getRootCauseMessage(e)), e);
     }
-
 
     return new SourceInputFormatProvider(DataDrivenETLDBInputFormat.class, connectionConfigAccessor.getConfiguration());
   }
 
   protected Connection getConnection(DBConnectorPath path) {
-    return getConnection(getConnectionString(path.getDatabase()) , config.getConnectionArgumentsProperties());
+    return getConnection(getConnectionString(path.getDatabase()), config.getConnectionArgumentsProperties());
   }
 
   protected String getConnectionString(String database) {
     return config.getConnectionString();
   }
 
+  protected String getTableName(String database, String schema, String table) {
+    return schema == null ? String.format("\"%s\".\"%s\"", database, table)
+      : String.format("\"%s\".\"%s\".\"%s\"", database, schema, table);
+  }
+
   protected String getTableQuery(String database, String schema, String table) {
-    return schema == null ? String.format("SELECT * FROM \"%s\".\"%s\"", database, table)
-      : String.format("SELECT * FROM \"%s\".\"%s\".\"%s\"", database, schema, table);
+    String tableName = getTableName(database, schema, table);
+    return String.format("SELECT * FROM %s", tableName);
   }
 
   protected String getTableQuery(String database, String schema, String table, int limit) {
-    return schema == null ?
-      String.format("SELECT * FROM \"%s\".\"%s\" LIMIT %d", database, table, limit) :
-      String.format(
-        "SELECT * FROM \"%s\".\"%s\".\"%s\" LIMIT %d", database, schema, table, limit);
+    String tableName = getTableName(database, schema, table);
+    return String.format("SELECT * FROM %s LIMIT %d", tableName, limit);
   }
 
-  protected Schema loadTableSchema(Connection connection, String query) throws SQLException {
+  protected String getTableQuery(String database, String schema, String table, int limit, String sampleType,
+                                 String strata, String sessionID) throws IOException {
+    if (sampleType == null) {
+      return getTableQuery(database, schema, table, limit);
+    }
+    String tableName = getTableName(database, schema, table);
+    switch (SampleType.fromString(sampleType)) {
+      case RANDOM:
+        return getRandomQuery(tableName, limit);
+      case STRATIFIED:
+        if (strata == null) {
+          throw new IllegalArgumentException("No strata column given.");
+        }
+        return getStratifiedQuery(tableName, limit, strata, sessionID);
+      default:
+        return getTableQuery(database, schema, table, limit);
+    }
+  }
+
+  // Get the query to use for randomized sampling.
+  // By default, databases don't support randomized sampling; this method must be overridden
+  protected String getRandomQuery(String tableName, int limit) throws IOException {
+    throw new IOException("Connection does not support random sampling.");
+  }
+
+  // Get the query to use for stratified sampling.
+  // By default, databases don't support stratified sampling; this method must be overridden
+  protected String getStratifiedQuery(String tableName, int limit, String strata, String sessionID) throws IOException {
+    throw new IOException("Connection does not support stratified sampling.");
+  }
+
+  protected Schema loadTableSchema(Connection connection, String query, @Nullable Integer timeoutSec, String sessionID)
+    throws SQLException {
     Statement statement = connection.createStatement();
     statement.setMaxRows(1);
+    if (timeoutSec != null) {
+      statement.setQueryTimeout(timeoutSec);
+    }
     ResultSet resultSet = statement.executeQuery(query);
-    return Schema.recordOf("outputSchema", getSchemaReader().getSchemaFields(resultSet));
+    return Schema.recordOf("outputSchema", getSchemaReader(sessionID).getSchemaFields(resultSet));
   }
 
   protected void setConnectionProperties(Map<String, String> properties, ConnectorSpecRequest request) {
@@ -144,7 +189,12 @@ public abstract class AbstractDBSpecificConnector<T extends DBWritable> extends 
   @Override
   protected Schema getTableSchema(Connection connection, String database,
                                   String schema, String table) throws SQLException {
+    String sessionID = generateSessionID();
+    return loadTableSchema(getConnection(), getTableQuery(database, schema, table),
+      null, sessionID);
+  }
 
-    return loadTableSchema(getConnection(),  getTableQuery(database, schema, table));
+  protected String generateSessionID() {
+    return UUID.randomUUID().toString().replace('-', '_');
   }
 }
--- a/mssql-plugin/src/main/java/io/cdap/plugin/mssql/SqlServerSourceSchemaReader.java
+++ b/mssql-plugin/src/main/java/io/cdap/plugin/mssql/SqlServerSourceSchemaReader.java
@@ -34,6 +34,17 @@ public class SqlServerSourceSchemaReader extends CommonSchemaReader {
   public static final int SQL_VARIANT = -156;
   public static final String DATETIME_TYPE_PREFIX = "datetime";
 
+  private final String sessionID;
+
+  public SqlServerSourceSchemaReader() {
+    this(null);
+  }
+
+  public SqlServerSourceSchemaReader(String sessionID) {
+    super();
+    this.sessionID = sessionID;
+  }
+
   @Override
   public Schema getSchema(ResultSetMetaData metadata, int index) throws SQLException {
     int columnSqlType = metadata.getColumnType(index);
@@ -71,5 +82,14 @@ public class SqlServerSourceSchemaReader extends CommonSchemaReader {
    */
   public static boolean shouldConvertToDatetime(String typeName) {
     return typeName.startsWith(DATETIME_TYPE_PREFIX);
+  }
+
+  @Override
+  public boolean shouldIgnoreColumn(ResultSetMetaData metadata, int index) throws SQLException {
+    if (sessionID == null) {
+      return false;
+    }
+    return metadata.getColumnName(index).equals("c_" + sessionID) ||
+      metadata.getColumnName(index).equals("sqn_" + sessionID);
   }
 }
--- a/mysql-plugin/src/main/java/io/cdap/plugin/mysql/MysqlConnector.java
+++ b/mysql-plugin/src/main/java/io/cdap/plugin/mysql/MysqlConnector.java
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2021 Cask Data, Inc.
+ * Copyright © 2021-2022 Cask Data, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not
  * use this file except in compliance with the License. You may obtain a copy of
@@ -27,10 +27,11 @@ import io.cdap.cdap.etl.api.connector.Connector;
 import io.cdap.cdap.etl.api.connector.ConnectorSpec;
 import io.cdap.cdap.etl.api.connector.ConnectorSpecRequest;
 import io.cdap.cdap.etl.api.connector.PluginSpec;
+import io.cdap.cdap.etl.api.connector.SampleType;
 import io.cdap.plugin.common.Constants;
 import io.cdap.plugin.common.ReferenceNames;
 import io.cdap.plugin.common.db.DBConnectorPath;
-import io.cdap.plugin.db.DBRecord;
+import io.cdap.plugin.db.SchemaReader;
 import io.cdap.plugin.db.connector.AbstractDBSpecificConnector;
 import org.apache.hadoop.io.LongWritable;
 import org.apache.hadoop.mapreduce.lib.db.DBWritable;
@@ -60,6 +61,11 @@ public class MysqlConnector extends AbstractDBSpecificConnector<MysqlDBRecord> {
   }
 
   @Override
+  protected SchemaReader getSchemaReader(String sessionID) {
+    return new MysqlSchemaReader(sessionID);
+  }
+
+  @Override
   protected Class<? extends DBWritable> getDBRecordType() {
     return MysqlDBRecord.class;
   }
@@ -70,7 +76,9 @@ public class MysqlConnector extends AbstractDBSpecificConnector<MysqlDBRecord> {
     setConnectionProperties(properties, request);
     builder
       .addRelatedPlugin(new PluginSpec(MysqlConstants.PLUGIN_NAME, BatchSource.PLUGIN_TYPE, properties))
-      .addRelatedPlugin(new PluginSpec(MysqlConstants.PLUGIN_NAME, BatchSink.PLUGIN_TYPE, properties));
+      .addRelatedPlugin(new PluginSpec(MysqlConstants.PLUGIN_NAME, BatchSink.PLUGIN_TYPE, properties))
+      .addSupportedSampleType(SampleType.RANDOM)
+      .addSupportedSampleType(SampleType.STRATIFIED);
 
     String table = path.getTable();
     if (table == null) {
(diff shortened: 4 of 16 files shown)


database-commons/src/main/java/io/cdap/plugin/db/connector/AbstractDBSpecificConnector.java
```diff
@@ -117,16 +118,24 @@ protected String getConnectionString(String database) {
     return config.getConnectionString();
   }
 
+  protected String getTableName(String database, String schema, String table) {
+    return schema == null ? String.format("\"%s\".\"%s\"", database, table)
```
tivv, line 128: Hm. I see that it's preexisting code, but which DB is it targeted for? Usually A.B. is schema.table, not database.table

tivv: Actually it's better to make it abstract. As far as I can see all subclasses override it. And `SqlServerConnector` override just one method, so it's even more weird

tivv: And I don't think database.table would work for SqlServer at all. https://docs.microsoft.com/en-us/sql/t-sql/queries/from-transact-sql?view=sql-server-ver16 gives `[ database_name . [ schema_name ] . | schema_name . ] table_or_view_name [ AS ] table_or_view_alias` format, so if one want to go schemaless it should be `database..table`, not `database.table`

ellliao: Hi Vitalii, thanks for the feedback!

These changes were part of the `getTableName()` refactor, and thus tried to preserve the previously implemented behavior as much as possible. The existence of this part of the code was kept because, while I could not find any subclasses of AbstractDBConnector.java in this repository, I was not sure if there were some elsewhere that depended on this part's existence (as I assumed it was originally implemented for a reason). If you believe it is fine to make this abstract, I can make that change.

As for SqlServer, I have not looked into whether this format works -- if not, then it was likely also broken previously. We can fix it, if so.

ellliao: I have just fixed SqlServer's table name to have the same behavior as previously. I am leaving `AbstractDBSpecificConnector`'s `getTableName()` function as is, for now, for the reasons listed.

database-commons/src/main/java/io/cdap/plugin/db/batch/source/DataDrivenETLDBInputFormat.java
```diff
@@ -130,52 +151,92 @@ public Connection createConnection() {
...
-        dbRecordReader.close();
-        try {
-          DriverManager.deregisterDriver(driverShim);
-        } catch (SQLException e) {
-          throw new IOException(e);
-        }
-      }
-    };
-  }
+    DBRecordReaderWithTimeout(DBInputSplit split, Class inputClass, Configuration conf, Connection conn,
```
albertshau: after poking around the code some more, it looks like we actually have a very similar need with the fetch size, which we're doing by using the ConnectionWithFetchSize class in the getConnection() method of this class.

Instead of wrapping the record reader, I think we should follow a similar pattern as that one, by adding a ConnectionWithTimeout class that wraps another connection and sets the query timeout.

albertshau: It would actually be better to replace that ConnectionWithFetchSize wrapper class with something a bit more generic, having something like:

```
public class StatementModifyingConnection extends ForwardingConnection {
  private final Connection delegate;
  private final Collection<StatementFunction> funcs;  

  public StatementModifyingConnection(Connection delegate, Collection<StatementFunction> funcs) {
    super(delegate);
    this.delegate = delegate;
    this.funcs = funcs;
  }

  private <T extends Statement> T modifyStatement(T stmt) throws SQLException {
    for (StatementFunction f : funcs) {
      f.apply(stmt);
    }
    return stmt;
  }

  @Override
  public Statement createStatement() throws SQLException {
    return modifyStatement(super.createStatement());
  }

  ...
}

public interface StatementFunction<T extends Statement> {

  void apply(T statement) throws SQLException;

}
```

then the getConnection() method can pass in list of functions with one to set the fetch size and another to set the query timeout if both are configured.

database-commons/src/main/java/io/cdap/plugin/db/batch/source/DataDrivenETLDBInputFormat.java
```diff
@@ -54,14 +62,27 @@ public static void setInput(Configuration conf,
                               Class<? extends DBWritable> inputClass,
                               String inputQuery,
                               String inputBoundingQuery,
-                              boolean enableAutoCommit) {
+                              boolean enableAutoCommit,
+                              @Nullable Long timeoutMs) {
```
albertshau: Statement.setQueryTimeout() takes an argument in seconds. This should be sent as seconds instead of millis

database-commons/src/main/java/io/cdap/plugin/db/batch/source/DataDrivenETLDBInputFormat.java
albertshau: there also seem to be various things being set in the ConnectionConfigAccessor, we should re-use the same class instead of passing the timeout in through setInput()

mysql-plugin/src/test/java/io/cdap/plugin/mysql/MysqlConnectorUnitTest.java
```diff
@@ -0,0 +1,71 @@
...
+                                                100, "random", null));
+    // random query, strata input
+    Assert.assertEquals(String.format("SELECT * FROM %s\n" +
+                                        "WHERE rand() < %d.0 / (SELECT COUNT(*) FROM %s)",
+                                      tableName, 100, tableName),
+                        CONNECTOR.getTableQuery("db", "schema", "table",
+                                                100, "random", "strata"));
+  }
+
+}
```
albertshau: files should end in a newline. This is normally caught be checkstyle, I guess it must not be setup for this project?

postgresql-plugin/src/main/java/io/cdap/plugin/postgres/PostgresConnector.java
```diff
@@ -112,13 +114,25 @@ public StructuredRecord transform(LongWritable longWritable, PostgresDBRecord re
...
-  protected String getTableQuery(String database, String schema, String table) {
-    return String.format("SELECT * FROM \"%s\".\"%s\"", schema, table);
+  protected String getTableName(String database, String schema, String table) {
+    return String.format("\"%s\".\"%s\"", schema, table);
   }
 
   @Override
-  protected String getTableQuery(String database, String schema, String table, int limit) {
-    return String.format("SELECT * FROM \"%s\".\"%s\" LIMIT %d", schema, table, limit);
+  protected String getTableQuery(String database, String schema, String table, int limit, String sampleType,
```
albertshau: looks like this general structure is repeated a bunch. It would be better to refactor things to remove the duplicate code. Seems like we only need to get the query string from each individual plugin

postgresql-plugin/src/main/java/io/cdap/plugin/postgres/PostgresConnector.java
```diff
@@ -112,13 +114,25 @@ public StructuredRecord transform(LongWritable longWritable, PostgresDBRecord re
...
+                                 String strata) {
+    if (sampleType == null) {
+      return getTableQuery(database, schema, table, limit);
+    }
+    String tableName = getTableName(database, schema, table);
+    switch (SampleType.fromString(sampleType)) {
+      case RANDOM:
+        // This query doesn't guarantee exactly "limit" number of rows
+        return String.format("SELECT * FROM %s\n" +
```
albertshau: when building queries, variables that come from user input need to be set using the PreparedStatement API (https://docs.oracle.com/javase/7/docs/api/java/sql/PreparedStatement.html) to prevent sql injection.

It's probably not practically needed here since things will likely fail earlier if somebody tries to inject sql in the table name and limit is likely validated to be a number somewhere earlier, but it is still better practice to do this using a PreparedStatement.

adinilfeld: Since the previous code already used Strings, and Ellen and I only have a week left, I'm thinking this change might be best in a separate PR. Do you think I could file a bug for it and leave this PR as-is?

mysql-plugin/src/test/java/io/cdap/plugin/mysql/MysqlConnectorUnitTest.java
ellliao: Fixed. Thanks!

postgresql-plugin/src/main/java/io/cdap/plugin/postgres/PostgresConnector.java
ellliao: Adin has refactored this. Thank you for the feedback!

postgresql-plugin/src/main/java/io/cdap/plugin/postgres/PostgresConnector.java
albertshau: yes you can leave it as is. On second thought this probably isn't straightforward, as you need to pass the query to the InputFormat as a string. Don't worry about filing a bug.

mysql-plugin/src/main/java/io/cdap/plugin/mysql/MysqlConnector.java
```diff
@@ -87,17 +100,50 @@ protected void setConnectorSpec(ConnectorSpecRequest request, DBConnectorPath pa
...
+                           tableName, limit, tableName);
+  }
+
+  @Override
+  protected String getStratifiedQuery(String tableName, int limit, String strata) {
+    if (strata == null) {
+      throw new IllegalArgumentException("No strata column given.");
+    }
+    return String.format("WITH t_%s AS (\n" +
+                    "    SELECT *,\n" +
```
albertshau: nit: fix indentation to be consistent with the rest of the project. The string itself should be indented 2 spaces and new arguments should line up with the previous argument:

```
a.call("abc " + 
         "def",
       nextArg);
```

mysql-plugin/src/test/java/io/cdap/plugin/mysql/MysqlConnectorUnitTest.java
```diff
@@ -0,0 +1,55 @@
...
+  /**
+   * Unit tests for getTableQuery()
+   */
+  @Test
+  public void getTableQueryTest() {
+    String tableName = CONNECTOR.getTableName("db", "schema", "table");
+
+    // random query
+    Assert.assertEquals(String.format("SELECT * FROM %s\n" +
+                            "WHERE rand() < %d.0 / (SELECT COUNT(*) FROM %s)",
```
albertshau: indentation

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleConnector.java
```diff
@@ -134,13 +136,25 @@ protected String getConnectionString(@Nullable String database) {
...
+    String tableName = getTableName(database, schema, table);
+    return String.format("SELECT * FROM %s WHERE ROWNUM <= %d", tableName, limit);
+  }
+
+  @Override
+  protected String getRandomQuery(String tableName, int limit) {
+    // This query guarantees exactly "limit" rows.
+    // Note that it is very slow on large tables, since it assigns _every_ row a number and then sorts them
+    return String.format("SELECT * FROM (\n" +
+                    "SELECT * FROM %s ORDER BY DBMS_RANDOM.RANDOM\n" +
```
albertshau: indentation

oracle-plugin/src/test/java/io/cdap/plugin/oracle/OracleConnectorUnitTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+package io.cdap.plugin.oracle;
+
+import org.junit.Assert;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.ExpectedException;
+
+public class OracleConnectorUnitTest {
+    @Rule
```
albertshau: indentation should be 2 spaces

oracle-plugin/src/test/java/io/cdap/plugin/oracle/OracleConnectorUnitTest.java
```diff
@@ -0,0 +1,59 @@
...
+  /**
+   * Unit tests for getTableQuery()
+   */
+  @Test
+  public void getTableQueryTest() {
+    String tableName = CONNECTOR.getTableName("db", "schema", "table");
+
+      // default query
+      Assert.assertEquals(String.format("SELECT * FROM %s WHERE ROWNUM <= %d", tableName, 100),
+              CONNECTOR.getTableQuery("db", "schema", "table",
```
albertshau: indentation

oracle-plugin/src/test/java/io/cdap/plugin/oracle/OracleConnectorUnitTest.java
```diff
@@ -0,0 +1,59 @@
...
+                      100));
+
+      // random query
+      Assert.assertEquals(String.format("SELECT * FROM (\n" +
+                              "SELECT * FROM %s ORDER BY DBMS_RANDOM.RANDOM\n" +
```
albertshau: indentation

postgresql-plugin/src/main/java/io/cdap/plugin/postgres/PostgresConnector.java
```diff
@@ -112,13 +120,44 @@ public StructuredRecord transform(LongWritable longWritable, PostgresDBRecord re
...
+  protected String getTableName(String database, String schema, String table) {
+    return String.format("\"%s\".\"%s\"", schema, table);
   }
 
   @Override
-  protected String getTableQuery(String database, String schema, String table, int limit) {
-    return String.format("SELECT * FROM \"%s\".\"%s\" LIMIT %d", schema, table, limit);
+  protected String getRandomQuery(String tableName, int limit) {
+    return String.format("SELECT * FROM %s\n" +
+                    "TABLESAMPLE BERNOULLI (100.0 * %d / (SELECT COUNT(*) FROM %s))",
```
albertshau: indentation

postgresql-plugin/src/test/java/io/cdap/plugin/postgres/PostgresConnectorUnitTest.java
```diff
@@ -0,0 +1,55 @@
...
+  /**
+   * Unit tests for getTableQuery()
+   */
+  @Test
+  public void getTableQueryTest() {
+    String tableName = CONNECTOR.getTableName("db", "schema", "table");
+
+    // random query
+    Assert.assertEquals(String.format("SELECT * FROM %s\n" +
+                            "TABLESAMPLE BERNOULLI (100.0 * %d / (SELECT COUNT(*) FROM %s))",
```
albertshau: indentation

database-commons/src/main/java/io/cdap/plugin/db/batch/source/StatementModifyingConnection.java
```diff
@@ -0,0 +1,111 @@
...
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.Collection;
+
+/**
+ * A JDBC {@link Connection} that delegates all methods to another {@link Connection}, but automatically call
+ * the {@link StatementFunction}s for all {@link Statement} created via this class.
+ */
+public class StatementModifyingConnection extends ForwardingConnection {
+    private final Collection<StatementFunction> funcs;
```
albertshau: indentation should be 2 spaces

database-commons/src/main/java/io/cdap/plugin/db/batch/source/StatementFunction.java
```diff
@@ -0,0 +1,30 @@
...
+import java.sql.SQLException;
+import java.sql.Statement;
+
+/**
+ * Used for {@link StatementModifyingConnection} to represent a function which can be applied to a statement.
+ * @param <T>
+ */
+public interface StatementFunction<T extends Statement> {
+
+    void apply(T statement) throws SQLException;
```
albertshau: indentation

database-commons/src/main/java/io/cdap/plugin/db/batch/source/DataDrivenETLDBInputFormat.java
```diff
@@ -102,8 +107,17 @@ public Connection getConnection() {
...
         if (connectionConfigAccessor.getFetchSize() > 0) {
-          this.connection = new ConnectionWithFetchSize(connection, connectionConfigAccessor.getFetchSize());
+          statementFuncs.add(statement -> statement.setFetchSize(connectionConfigAccessor.getFetchSize()));
+        }
+
+        if (connectionConfigAccessor.getQueryTimeout() > 0) {
+          statementFuncs.add(statement -> statement.setQueryTimeout(connectionConfigAccessor.getQueryTimeout()));
+        }
+
+        if (statementFuncs.size() > 0) {
```
albertshau: nit: use `!statementFunctions.isEmpty()` instead, it's slightly more readable

database-commons/src/main/java/io/cdap/plugin/db/batch/source/StatementFunction.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+package io.cdap.plugin.db.batch.source;
+
+import java.sql.SQLException;
+import java.sql.Statement;
+
+/**
+ * Used for {@link StatementModifyingConnection} to represent a function which can be applied to a statement.
+ * @param <T>
```
albertshau: nit: should have a non-empty javadoc "type of statement to modify" or something like that

albertshau (commented):
mostly style comments, otherwise looks good. Please fix, then squash into a single commit and I'll do a final review before merging.

mysql-plugin/src/main/java/io/cdap/plugin/mysql/MysqlConnector.java
ellliao: Fixed indentation in connectors and unit tests.

mysql-plugin/src/main/java/io/cdap/plugin/mysql/MysqlConnector.java
```diff
@@ -87,17 +101,51 @@ protected void setConnectorSpec(ConnectorSpecRequest request, DBConnectorPath pa
...
+  @Override
+  protected String getTableName(String database, String schema, String table) {
+    return String.format("`%s`.`%s`", database, table);
   }
 
   @Override
-  protected String getTableQuery(String database, String schema, String table, int limit) {
-    return String.format("SELECT * FROM `%s`.`%s` LIMIT %d", database, table, limit);
+  protected String getRandomQuery(String tableName, int limit) {
+      // This query doesn't guarantee exactly "limit" number of rows
```
albertshau: indentation

database-commons/src/main/java/io/cdap/plugin/db/ConnectionConfigAccessor.java
```diff
@@ -112,4 +113,11 @@ public Configuration getConfiguration() {
     return configuration;
   }
 
+  public void setQueryTimeout(int timeout) {
```
albertshau: these changes don't look like they're used anymore

mysql-plugin/src/main/java/io/cdap/plugin/mysql/MysqlConnector.java
```diff
@@ -87,17 +101,51 @@ protected void setConnectorSpec(ConnectorSpecRequest request, DBConnectorPath pa
...
+  protected Schema loadTableSchema(Connection connection, String query, @Nullable Integer timeoutSec)
+    throws SQLException {
+    Statement statement = connection.createStatement();
+    statement.setMaxRows(1);
+    if (timeoutSec != null) {
+      statement.setQueryTimeout(timeoutSec);
+    }
+    ResultSet resultSet = statement.executeQuery(query);
+    MysqlSchemaReader schemaReader = (MysqlSchemaReader) getSchemaReader();
+    schemaReader.setSessionID(sessionID);
```
albertshau: if you set the sessionId in the getSchemaReader() method, you don't need to do a cast above. By removing the cast, it also seems like it would also allow all the logic to be pulled up to the parent class instead of duplicated in each connector

albertshau (commented):
minor comments. I also thought the timeout was wanted to limit the sampling query? These changes would only apply a timeout on getting the schema.

It's also ok to separate the timeout stuff into a separate PR and just try to get this piece merged first.

mysql-plugin/src/main/java/io/cdap/plugin/mysql/MysqlConnector.java
ellliao: Changed. Thanks!

database-commons/src/main/java/io/cdap/plugin/db/connector/AbstractDBSpecificConnector.java
```diff
@@ -84,54 +87,102 @@ public InputFormatProvider getInputFormatProvider(ConnectorContext context, Samp
...
     } else {
       DBConfiguration.configureDB(connectionConfigAccessor.getConfiguration(), driverClass.getName(),
-                     getConnectionString(path.getDatabase()), config.getUser(), config.getPassword());
+                                  getConnectionString(path.getDatabase()), config.getUser(), config.getPassword());
     }
-    String tableQuery = getTableQuery(path.getDatabase(), path.getSchema(), path.getTable(), request.getLimit());
+    // SessionID cannot be greater than 28 characters, as it is used as a suffix for variable names in the SQL queries
+    // in OracleConnector, and Oracle does not allow variable names over 30 characters.
+    // Thus sessionID is not a "true" UUID, but the chance of collision is still negligible.
+    sessionID = java.util.UUID.randomUUID()
```
albertshau: nit: change to `UUID.randomUUID()` and add the import for java.util.UUID

database-commons/src/main/java/io/cdap/plugin/db/connector/AbstractDBSpecificConnector.java
```diff
@@ -84,54 +87,102 @@ public InputFormatProvider getInputFormatProvider(ConnectorContext context, Samp
...
-                     getConnectionString(path.getDatabase()));
+                                  getConnectionString(path.getDatabase()));
     } else {
       DBConfiguration.configureDB(connectionConfigAccessor.getConfiguration(), driverClass.getName(),
-                     getConnectionString(path.getDatabase()), config.getUser(), config.getPassword());
+                                  getConnectionString(path.getDatabase()), config.getUser(), config.getPassword());
     }
-    String tableQuery = getTableQuery(path.getDatabase(), path.getSchema(), path.getTable(), request.getLimit());
+    // SessionID cannot be greater than 28 characters, as it is used as a suffix for variable names in the SQL queries
+    // in OracleConnector, and Oracle does not allow variable names over 30 characters.
```
albertshau: is this only a restriction for oracle and not other dbs? If so, it would be better to make a generateSessionID() method which has a default implementation, then have just the oracle plugin override it to some different logic.

It would also be a bit better to strip the '-' characters instead of replacing with '_'. This will provide 4 extra characters.

database-commons/src/main/java/io/cdap/plugin/db/connector/AbstractDBSpecificConnector.java
```diff
@@ -84,54 +87,102 @@ public InputFormatProvider getInputFormatProvider(ConnectorContext context, Samp
...
     connectionConfigAccessor.setConnectionArguments(Maps.fromProperties(config.getConnectionArgumentsProperties()));
     connectionConfigAccessor.getConfiguration().setInt(MRJobConfig.NUM_MAPS, 1);
     Map<String, String> additionalArguments = config.getAdditionalArguments();
     for (Map.Entry<String, String> argument : additionalArguments.entrySet()) {
       connectionConfigAccessor.getConfiguration().set(argument.getKey(), argument.getValue());
     }
     try {
-      connectionConfigAccessor.setSchema(loadTableSchema(getConnection(path),  tableQuery).toString());
+      Long timeoutMs = request.getTimeoutMs();
+      Integer timeoutSec = timeoutMs != null ? (int) (timeoutMs /  1000) : null;
```
albertshau: nit: remove extra space after `/`

database-commons/src/main/java/io/cdap/plugin/db/connector/AbstractDBSpecificConnector.java
```diff
@@ -44,15 +44,18 @@
...
 /**
  * An Abstract DB Specific Connector those specific DB connectors can inherits
+ *
  * @param <T> the Record type that specific DB Record Reader may return while sample the data with InputFormat
  */
 public abstract class AbstractDBSpecificConnector<T extends DBWritable> extends AbstractDBConnector
   implements BatchConnector<LongWritable, T> {
 
   private final AbstractDBConnectorConfig config;
+  protected String sessionID;
```
albertshau: it usually a bad idea to have stateful variable like this, especially when it is not private, as there is no clearly defined contract around when it is set, how it should be used, etc. For example, right now it gets set in getInputFormatProvider() method. If a subclass tries to use this before that, it will be null, which may not be expected. If somebody changes the order of when methods are called, they can unintentionally break the logic.

Instead, it is better to pass this into the methods that require it. In this particular case, it looks like it only needs to be passed into the getSchemaReader() method and getStratifiedQuery() method.

database-commons/src/main/java/io/cdap/plugin/db/CommonSchemaReader.java
```diff
@@ -61,4 +63,8 @@ public Schema getSchema(ResultSetMetaData metadata, int index) throws SQLExcepti
   public boolean shouldIgnoreColumn(ResultSetMetaData metadata, int index) throws SQLException {
     return false;
   }
+
+  public void setSessionID(String sessionID) {
```
albertshau: try to avoid setters when possible. They often violate encapsulation, as outside classes are now able to change the internal state of this class. Normally, it's better to pass in the value into the constructor as save it as a final variable.

database-commons/src/main/java/io/cdap/plugin/db/ConnectionConfigAccessor.java
```diff
@@ -37,6 +37,7 @@ public class ConnectionConfigAccessor {
   private static final String INIT_QUERIES = "io.cdap.plugin.db.init.queries";
   public static final String AUTO_COMMIT_ENABLED = "io.cdap.plugin.db.output.autocommit.enabled";
   public static final String FETCH_SIZE = "io.cdap.plugin.db.fetch.size";
+  public static final String QUERY_TIMEOUT = "io.cdap.plugin.db.timeout";
```
albertshau: this can be removed

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceSchemaReader.java
```diff
@@ -56,6 +56,17 @@ public class OracleSourceSchemaReader extends CommonSchemaReader {
     Types.DECIMAL
   );
 
+  private String sessionID;
+
+  public OracleSourceSchemaReader() {
+    super();
```
albertshau: nit: call the other constructor `OracleSourceSchemaReader(null)`

This is functionally the same today, but this is slightly better in case the other constructor is later modified to have any other logic.

postgresql-plugin/src/main/java/io/cdap/plugin/postgres/PostgresSchemaReader.java
```diff
@@ -38,6 +38,17 @@ public class PostgresSchemaReader extends CommonSchemaReader {
     "bit", "timetz", "money"
   );
 
+  private String sessionID;
+
+  public PostgresSchemaReader() {
+    super();
```
albertshau: similar comment, call the other constructor

mssql-plugin/src/main/java/io/cdap/plugin/mssql/SqlServerSourceSchemaReader.java
```diff
@@ -34,6 +34,17 @@ public class SqlServerSourceSchemaReader extends CommonSchemaReader {
   public static final int SQL_VARIANT = -156;
   public static final String DATETIME_TYPE_PREFIX = "datetime";
 
+  private String sessionID;
+
+  public SqlServerSourceSchemaReader() {
+    super();
```
albertshau: similar comment, call the other constructor

albertshau (approved):
minor comments, otherwise lgtm. Please fix and squash into one commit before merging.

database-commons/src/main/java/io/cdap/plugin/db/ConnectionConfigAccessor.java
adinilfeld: Done!

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceSchemaReader.java
adinilfeld: Good point -- and then the `sessionID` variable can be `final` as well. Done!

postgresql-plugin/src/main/java/io/cdap/plugin/postgres/PostgresSchemaReader.java
adinilfeld: Fixed, thanks!

mssql-plugin/src/main/java/io/cdap/plugin/mssql/SqlServerSourceSchemaReader.java
adinilfeld: Fixed!

sechegaray (approved):

PR #293, opened 2022-09-20
add macros for Connection Name and Database for CloudSQL MySQL & Clou…

jster1357:
add macros for Connection Name and Database for CloudSQL MySQL & CloudSQL Postgres for Source,Sink, and Action


itsankit-google:
Was added in this PR: https://github.com/data-integrations/database-plugins/pull/414.

PR #308, opened 2022-10-11
e2e tests secrets manager for credentials

sawantpritam:
@itsankit-google @rmstar Please review.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -23,7 +23,7 @@ on:
   workflow_dispatch:
 
 jobs:
-  build:
+  secretsmanager:
     runs-on: k8s-runner-e2e
     # We allow builds:
     # 1) When triggered manually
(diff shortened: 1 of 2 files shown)


mysql-plugin/src/e2e-test/java/io/cdap/plugin/common/stepsdesign/TestSetupHooks.java
```diff
@@ -30,11 +30,11 @@ public class TestSetupHooks {
 
   @Before(order = 1)
   public static void overrideUserAndPasswordIfProvided() {
-    String username = System.getenv("username");
+    String username = System.getenv("MYSQL_USERNAME");
```
itsankit-google, line 33: why `MYSQL`? This should be configurable right depending on the test??

.github/workflows/e2e.yml
```diff
@@ -23,7 +23,7 @@ on:
   workflow_dispatch:
 
 jobs:
-  build:
+  secretsmanager:
```
itsankit-google, line 26: why a separate job is needed??

.github/workflows/e2e.yml
```diff
@@ -38,6 +38,35 @@ jobs:
       || (contains(github.event.pull_request.labels.*.name, 'build')
           && (github.event.action != 'labeled' || github.event.label.name == 'build')
           )
+    steps:
+      - name: Get Secrets from GCP Secret Manager
+        id: 'secrets'
+        uses: 'google-github-actions/get-secretmanager-secrets@v0'
+        with:
+          secrets: |-
```
itsankit-google, line 46: It would be good to fetch only those secrets depending on the module rather than all.

mysql-plugin/src/e2e-test/java/io/cdap/plugin/common/stepsdesign/TestSetupHooks.java
sawantpritam: This file is inside mysql module so expecting MYSQL credentials only

.github/workflows/e2e.yml
sawantpritam: this is required to set env variable on `build` job level. If we add secret manager step under `build` job then we will need to set env variables at 2 steps (run all tests and run required tests) separately

PR #312, opened 2022-10-31
Feature/plugin 1433 oracle numeric scale changes

MrRahulSharma:
- Showing Warning in case any precision less Numeric type is detected when reading from Oracle DB.
- Allowing the change in Scale while handling Numeric types in Oracle plugins.

Validations performed using Oracle DB Plugin:
1. Number(Precision, Scale) type on GetSchema detects as Number(Precision, Scale), and pipeline runs without overriding the scale. Expected: Pipeline should run using Scale detected. Actual: Same as Expectation
2. Number(Precision, Scale) type on GetSchema detects as Number(Precision, Scale), and pipeline runs after overriding the scale to scale'. Expected : Pipeline should run using scale' set by the user. Actual: Same as Expectation
3. Number(Precision, Scale) type and direct pipeline run without GetSchema. Expected: Pipeline should run using the scale. Actual: Same as Expectation
4. Number with undefined precision and scale detected and pipeline run without any user override for scale. Expected: A warning message in the logs regarding to the Number detected with undefined precision and scale and the type gets default to use Number(38, 0). Actual: Same as Expectation
5. Number with undefined precision and scale detected and pipeline run with the user changing the scale to say scale'. Expected: A warning message in the logs regarding to the Number detected with undefined precision and scale and the user provided scale' is used. Actual: Same as Expectation

### DIFF
--- a/oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceDBRecord.java
+++ b/oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceDBRecord.java
@@ -218,16 +218,16 @@ public class OracleSourceDBRecord extends DBRecord {
           // scale set in the logical schema. For example for value '77.12' if the scale set in the logical schema is
           // set to 4 then the number will change to '77.1200'. Also if the value is '22.1274' and the logical schema
           // scale is set to 2 then the decimal value used will be '22.13' after rounding.
-          BigDecimal decimal = resultSet.getBigDecimal(columnIndex, getSchemaScale(field.getSchema()));
+          BigDecimal decimal = resultSet.getBigDecimal(columnIndex, getScale(field.getSchema()));
           recordBuilder.setDecimal(field.getName(), decimal);
         }
     }
   }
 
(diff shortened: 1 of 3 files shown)


itsankit-google (approved):
LGTM, 

one question - will the optimisation of treating precision less numbers to string be taken as a separate task? If yes, can we create a separate JIRA for that??

itsankit-google:
Please squash commits before merging the PR.

MrRahulSharma:
> will the optimisation of treating precision less numbers to string be taken as a separate task? If yes, can we create a separate JIRA for that??

Yes it will be treated as a separate task and I believe https://cdap.atlassian.net/browse/PLUGIN-1119 is to track the same.

itsankit-google:
> > will the optimisation of treating precision less numbers to string be taken as a separate task? If yes, can we create a separate JIRA for that??
> 
> Yes it will be treated as a separate task and I believe https://cdap.atlassian.net/browse/PLUGIN-1119 is to track the same.

Got it, thanks for the info.

PR #316, opened 2022-11-04
Using the scale from logical schema instead of the number value.

MrRahulSharma:
Cherry-pick of the commits from the PR : https://github.com/data-integrations/database-plugins/pull/312

Using the scale from logical schema instead of the number value.
Function rename and doc correction
Added Warning logs when detecting precision less Oracle numbers.
Added Unit Tests for precision less Number.


itsankit-google (approved):
Please add the link of original pr in description.

PR #334, opened 2023-01-13
Added E2E test for Oracle plugin

Aryan-Verma:
Created the E2E test for Oracle Plugin

snapshot:
Screenshot 2023-01-12 at 2 17 17 PM

result:
Screenshot 2023-01-12 at 2 18 51 PM

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -40,8 +40,9 @@ jobs:
           )
     strategy:
       matrix:
-        module: [mysql-plugin, postgresql-plugin]
+        module: [mysql-plugin, postgresql-plugin, oracle-plugin]
       fail-fast: false
+
     steps:
       # Pinned 1.0.0 version
       - uses: actions/checkout@v3
@@ -73,13 +74,33 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-maven-${{ github.workflow }}
 
+      - name: Get Secrets from GCP Secret Manager
+        id: secrets
+        uses: 'google-github-actions/get-secretmanager-secrets@v0'
+        with:
+          secrets: |-
+            ORACLE_HOST:cdapio-github-builds/ORACLE_HOST
+            ORACLE_USERNAME:cdapio-github-builds/ORACLE_USERNAME
+            ORACLE_PASSWORD:cdapio-github-builds/ORACLE_PASSWORD
+            ORACLE_PORT:cdapio-github-builds/ORACLE_PORT
+
       - name: Run required e2e tests
         if: github.event_name != 'workflow_dispatch' && github.event_name != 'push' && steps.filter.outputs.e2e-test == 'false'
         run: python3 e2e/src/main/scripts/run_e2e_test.py --module ${{ matrix.module }} --testRunner TestRunnerRequired.java
+        env:
+          ORACLE_HOST: ${{ steps.secrets.outputs.ORACLE_HOST }}
+          ORACLE_USERNAME: ${{ steps.secrets.outputs.ORACLE_USERNAME }}
+          ORACLE_PASSWORD: ${{ steps.secrets.outputs.ORACLE_PASSWORD }}
+          ORACLE_PORT: ${{ steps.secrets.outputs.ORACLE_PORT }}
 
       - name: Run all e2e tests
         if: github.event_name == 'workflow_dispatch' || github.event_name == 'push' || steps.filter.outputs.e2e-test == 'true'
         run: python3 e2e/src/main/scripts/run_e2e_test.py --module ${{ matrix.module }}
+        env:
+          ORACLE_HOST: ${{ steps.secrets.outputs.ORACLE_HOST }}
+          ORACLE_USERNAME: ${{ steps.secrets.outputs.ORACLE_USERNAME }}
+          ORACLE_PASSWORD: ${{ steps.secrets.outputs.ORACLE_PASSWORD }}
+          ORACLE_PORT: ${{ steps.secrets.outputs.ORACLE_PORT }}
 
       - name: Upload report
         uses: actions/upload-artifact@v3
--- a/oracle-plugin/src/e2e-test/features/oracle/Oracle.feature
+++ b/oracle-plugin/src/e2e-test/features/oracle/Oracle.feature
@@ -1,0 +1,68 @@
+#
+# Copyright © 2023 Cask Data, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not
+# use this file except in compliance with the License. You may obtain a copy of
+# the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations under
+# the License.
+#
+
+@Oracle
+Feature: Oracle - Verify Oracle source data transfer
+  @ORACLE_SOURCE_TEST @ORACLE_SINK_TEST @Oracle_Required
+  Scenario: To verify data is getting transferred from Oracle to Oracle successfully
+    Given Open Datafusion Project to configure pipeline
+    When Expand Plugin group in the LHS plugins list: "Source"
+    When Select plugin: "Oracle" from the plugins list as: "Source"
+    When Expand Plugin group in the LHS plugins list: "Sink"
+    When Select plugin: "Oracle" from the plugins list as: "Sink"
+    Then Connect plugins: "Oracle" and "Oracle2" to establish connection
+    Then Navigate to the properties page of plugin: "Oracle"
+    Then Select dropdown plugin property: "select-jdbcPluginName" with option value: "driverName"
+    Then Replace input plugin property: "host" with value: "host" for Credentials and Authorization related fields
+    Then Replace input plugin property: "port" with value: "port" for Credentials and Authorization related fields
+    Then Replace input plugin property: "user" with value: "username" for Credentials and Authorization related fields
+    Then Replace input plugin property: "password" with value: "password" for Credentials and Authorization related fields
+    Then Select radio button plugin property: "connectionType" with value: "service"
+    Then Select radio button plugin property: "role" with value: "sysdba"
+    Then Enter input plugin property: "referenceName" with value: "sourceRef"
+    Then Replace input plugin property: "database" with value: "databaseName"
+    Then Enter textarea plugin property: "importQuery" with value: "selectQuery"
+    Then Click on the Get Schema button
+    Then Verify the Output Schema matches the Expected Schema: "outputSchema"
+    Then Validate "Oracle" plugin properties
+    Then Close the Plugin Properties page
+    Then Navigate to the properties page of plugin: "Oracle2"
+    Then Select dropdown plugin property: "select-jdbcPluginName" with option value: "driverName"
+    Then Replace input plugin property: "host" with value: "host" for Credentials and Authorization related fields
+    Then Replace input plugin property: "port" with value: "port" for Credentials and Authorization related fields
+    Then Replace input plugin property: "database" with value: "databaseName"
+    Then Replace input plugin property: "tableName" with value: "targetTable"
+    Then Replace input plugin property: "dbSchemaName" with value: "schema"
+    Then Replace input plugin property: "user" with value: "username" for Credentials and Authorization related fields
+    Then Replace input plugin property: "password" with value: "password" for Credentials and Authorization related fields
+    Then Enter input plugin property: "referenceName" with value: "targetRef"
+    Then Select radio button plugin property: "connectionType" with value: "service"
+    Then Select radio button plugin property: "role" with value: "sysdba"
+    Then Validate "Oracle2" plugin properties
+    Then Close the Plugin Properties page
+    Then Save the pipeline
+    Then Preview and run the pipeline
+    Then Verify the preview of pipeline is "success"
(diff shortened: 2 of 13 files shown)


.github/workflows/e2e.yml
```diff
@@ -40,10 +60,19 @@ jobs:
...
     steps:
       # Pinned 1.0.0 version
+      - name: Run Tests
+        if: github.event_name == 'workflow_dispatch' || github.event_name == 'push' || steps.filter.outputs.e2e-test == 'true'
+        run: python3 e2e/src/main/scripts/run_e2e_test.py --module ${{ matrix.module }}
+        env:
+          ORACLE_HOST: ${{ needs.secretsmanager.outputs.ORACLE_HOST }}
+          ORACLE_USERNAME: ${{ needs.secretsmanager.outputs.ORACLE_USERNAME }}
+          ORACLE_PASSWORD: ${{ needs.secretsmanager.outputs.ORACLE_PASSWORD }}
+          ORACLE_PORT: ${{ needs.secretsmanager.outputs.ORACLE_PORT }}
```
itsankit-google: There is already `Run required e2e tests` and `Run all e2e tests` steps below. Why do we need this extra step?

We should be able to add these env variables in both of these steps.

oracle-plugin/src/e2e-test/java/io.cdap.plugin/OracleClient.java
```diff
@@ -0,0 +1,67 @@
...
+import java.sql.Statement;
+import java.util.TimeZone;
+
+/**
+ *  Oracle client.
+ */
+public class OracleClient {
+
+  private static Connection getOracleConnection() throws SQLException, ClassNotFoundException {
+    TimeZone timezone = TimeZone.getTimeZone("Asia/Kolkata");
```
itsankit-google: nit: can we use `UTC` instead of `IST`?

oracle-plugin/src/e2e-test/java/io.cdap.plugin/common.stepsdesign/TestSetupHooks.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+    String host = System.getenv("ORACLE_HOST");
+    if (host != null && !host.isEmpty()) {
+      PluginPropertyUtils.addPluginProp("ORACLE_HOST", host);
+    }
+    String port = System.getenv("ORACLE_PORT");
+    if (port != null && !port.isEmpty()) {
+      PluginPropertyUtils.addPluginProp("ORACLE_PORT", port);
+    }
+  }
```
itsankit-google: why is this hook needed?

oracle-plugin/src/e2e-test/java/io.cdap.plugin/OracleClient.java
```diff
@@ -0,0 +1,67 @@
...
+    String username = PluginPropertyUtils.pluginProp("ORACLE_USERNAME");
+    String password = PluginPropertyUtils.pluginProp("ORACLE_PASSWORD");
+    int port = Integer.parseInt(PluginPropertyUtils.pluginProp("ORACLE_PORT"));
+    String database = PluginPropertyUtils.pluginProp("database");
+    return DriverManager.getConnection("jdbc:oracle:thin:@//" + host + ":" + port + "/" + database, username, password);
+  }
+
+  public static int countRecord(String table, String schema) throws SQLException, ClassNotFoundException {
+//    String countQuery = "SELECT COUNT(*) as total FROM " + schema + "." + table;
+    String countQuery = PluginPropertyUtils.pluginProp("selectQuery");
```
itsankit-google: why using select query for counting records?

oracle-plugin/src/e2e-test/resources/pluginParameters.properties
```diff
@@ -0,0 +1,13 @@
+driver=oracle
+database=xe
```
itsankit-google: nit: `driver->driverName` & `database->databaseName`

oracle-plugin/src/e2e-test/java/io.cdap.plugin/OracleClient.java
Aryan-Verma: Done

oracle-plugin/src/e2e-test/java/io.cdap.plugin/common.stepsdesign/TestSetupHooks.java
Aryan-Verma: Removed the hook. Using these credentials directly from the environment now.

oracle-plugin/src/e2e-test/java/io.cdap.plugin/OracleClient.java
Aryan-Verma: Changed it to count query

oracle-plugin/src/e2e-test/resources/pluginParameters.properties
Aryan-Verma: Done

itsankit-google (commented):
One question, are we depending on source & target table to be already present in the database?? 

If yes, this is not good, tomorrow if I point the test to another database, the test should not fail. We should be creating and deleting the required tables using test setup hooks.

oracle-plugin/src/e2e-test/java/io.cdap.plugin/OracleClient.java
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  public static int countRecord(String table, String schema) throws SQLException, ClassNotFoundException {
+    String countQuery = "SELECT COUNT(*) as total FROM " + schema + "." + table;
+    try (Connection connect = getOracleConnection(); Statement statement = connect.createStatement();
+         ResultSet rs = statement.executeQuery(countQuery)) {
+      int num = 0;
+      while (rs.next()) {
+        num = (rs.getInt(1));
+      }
```
itsankit-google, line 50: Why do we need a while loop here?
The count query should return only one row and that will be the count, isn't it??

oracle-plugin/src/e2e-test/features/oracle/Oracle.feature
```diff
@@ -0,0 +1,69 @@
...
+    Then Verify preview output schema matches the outputSchema captured in properties
+    Then Close the preview data
+    Then Deploy the pipeline
+    Then Run the Pipeline in Runtime
+    Then Wait till pipeline is in running state
+    Then Open and capture logs
+    Then Verify the pipeline status is "Succeeded"
+    Then Get count of no of records transferred to target Oracle Table
+    Then Validate records transferred to target table is equal to number of records from source table
+    Then Truncate target table
```
MrRahulSharma: Ideally, the truncate table step should be at the start as well to make sure that there is no repercussion in case of some E2E test failures for subsequent runs.

oracle-plugin/src/e2e-test/java/io.cdap.plugin/oracle/stepsdesign/Oracle.java
```diff
@@ -0,0 +1,64 @@
...
+ */
+public class Oracle implements CdfHelper {
+
+  @Then("Click on preview data for Oracle sink")
+  public void clickOnPreviewDataForOracleSink() {
+    openSinkPluginPreviewData("Oracle");
+  }
+
+
+  @Then("Get count of no of records transferred to target Oracle Table")
```
MrRahulSharma, line 40: Ideally this should named as "Validate number of records in target table matches with the number of records written by the pipeline."

oracle-plugin/src/e2e-test/resources/pluginParameters.properties
```diff
@@ -0,0 +1,13 @@
+driverName=oracle
+databaseName=xe
+selectQuery=select * from ${schema}.${table}
+sourceRef=source
+targetRef=target
+sourceTable=ST2
```
MrRahulSharma: Please use the full name such as "SourceTable" instead of "ST2". Same for TT2 as well.

oracle-plugin/src/e2e-test/java/io.cdap.plugin/oracle/stepsdesign/Oracle.java
Aryan-Verma: We have a function below this for validating if the number of records match. This function is just to get the count of records now available at the sink side.

oracle-plugin/src/e2e-test/java/io.cdap.plugin/OracleClient.java
Aryan-Verma: The `statement.executeQuery` returns an interface `ResultSet` which has a cursor initially pointed before its first row and needs to be moved forward using loop

oracle-plugin/src/e2e-test/resources/pluginParameters.properties
Aryan-Verma: Done

oracle-plugin/src/e2e-test/features/oracle/Oracle.feature
Aryan-Verma: Added delete table function. So truncating won't be needed

Aryan-Verma:
> One question, are we depending on source & target table to be already present in the database??
> 
> If yes, this is not good, tomorrow if I point the test to another database, the test should not fail. We should be creating and deleting the required tables using test setup hooks.

Added functions to create and delete tables, instead of using already existing tables

oracle-plugin/src/e2e-test/features/oracle/Oracle.feature
```diff
@@ -0,0 +1,70 @@
...
+    Then Verify preview output schema matches the outputSchema captured in properties
+    Then Close the preview data
+    Then Deploy the pipeline
+    Then Run the Pipeline in Runtime
+    Then Wait till pipeline is in running state
+    Then Open and capture logs
+    Then Verify the pipeline status is "Succeeded"
+    Then Get count of no of records transferred to target Oracle Table
+    Then Validate records transferred to target table is equal to number of records from source table
+    Then Delete tables
```
itsankit-google: If the test fails in between, this step won't be executed and the resources will be left behind.
This is why we use cucumber step hooks so that they will be performed irrespective of test results.

Aryan-Verma: Added in cucumber step hooks

oracle-plugin/src/e2e-test/java/io.cdap.plugin/common.stepsdesign/TestSetupHooks.java
```diff
@@ -0,0 +1,55 @@
...
+
+  @Before(order = 2, value = "@ORACLE_SOURCE_TEST")
+  public static void createTables() throws SQLException, ClassNotFoundException {
+    OracleClient.createSourceTable(PluginPropertyUtils.pluginProp("sourceTable"),
+                                   PluginPropertyUtils.pluginProp("schema"));
+    OracleClient.createTargetTable(PluginPropertyUtils.pluginProp("targetTable"),
+                                   PluginPropertyUtils.pluginProp("schema"));
+  }
+
+  @After(order = 2, value = "@ORACLE_SINK_TEST")
```
itsankit-google: order of execution of `after` step hooks is reverse than the `before` hooks, we can keep `order=1` here.

itsankit-google (approved):
1 comment otherwise LGTM, please squash the commits before merge.

oracle-plugin/src/e2e-test/java/io.cdap.plugin/common.stepsdesign/TestSetupHooks.java
Aryan-Verma: done

oracle-plugin/src/e2e-test/java/io.cdap.plugin/OracleClient.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+  }
+
+  public static void createSourceTable(String sourceTable, String schema) throws SQLException, ClassNotFoundException {
+    try (Connection connect = getOracleConnection(); Statement statement = connect.createStatement()) {
+      String createSourceTableQuery = "CREATE TABLE " + schema + "." + sourceTable +
+        "(ID number(38), LASTNAME varchar2(100))";
+      statement.executeUpdate(createSourceTableQuery);
+
+      // Truncate table to clean the data of last failure run.
```
MrRahulSharma: This might not be required now

oracle-plugin/src/e2e-test/java/io.cdap.plugin/OracleClient.java
```diff
@@ -0,0 +1,90 @@
...
+      String createTargetTableQuery = "CREATE TABLE " + schema + "." + targetTable +
+        "(ID number(38), LASTNAME varchar2(100))";
+      statement.executeUpdate(createTargetTableQuery);
+    }
+  }
+
+  public static void deleteTables(String schema, String sourceTable, String targetTable)
+    throws SQLException, ClassNotFoundException {
+    try (Connection connect = getOracleConnection(); Statement statement = connect.createStatement()) {
+      String dropSourceTableQuery = "DROP TABLE " + schema + "." + sourceTable;
```
MrRahulSharma: Can we use String[] instead of sourceTable and targetTable and run the delete query in loop.

oracle-plugin/src/e2e-test/java/io.cdap.plugin/OracleClient.java
Aryan-Verma: Removed

oracle-plugin/src/e2e-test/java/io.cdap.plugin/OracleClient.java
Aryan-Verma: Changed it to String[]

PR #357, opened 2023-02-24
[PLUGIN-1503] Package conflict resolution between database-commons and database-plugins

MrRahulSharma:
JIRA : https://cdap.atlassian.net/browse/PLUGIN-1503

Issue: 
The class ETLDBOutputFormat is present in both the generic database plugin and Oracle plugin at the same package io.cdap.plugin.db.batch.sink.ETLDBOutputFormat. At run time wrong class is used in Oracle plugin sink write process, due to which the provided connection arguments are not respected during connection to the Oracle plugin. 
This issue occurs with any sink plugin present in the database-plugins artifact when used along with Database-plugin bundled with the CDF.

Resolution:
Changed the package name in the database-commons to io.cdap.plugin.db.sink


aurora-mysql-plugin/pom.xml
```diff
@@ -99,8 +99,8 @@
           <instructions>
             <_exportcontents>
               io.cdap.plugin.auroradb.mysql.*;
-              io.cdap.plugin.db.batch.source.*;
-              io.cdap.plugin.db.batch.sink.*;
+              io.cdap.plugin.db.source.*;
+              io.cdap.plugin.db.sink.*;
               org.apache.commons.lang;
```
albertshau, line 104: Not related to this PR, but these are some interesting exports. I don't see why these should be needed, but they can definitely cause issues with other plugins. Normally we should only need to export packages from the project itself.

Can you open a jira to investigate whether we can remove these exports?

aurora-mysql-plugin/pom.xml
```diff
@@ -99,8 +99,8 @@
           <instructions>
             <_exportcontents>
               io.cdap.plugin.auroradb.mysql.*;
-              io.cdap.plugin.db.batch.source.*;
-              io.cdap.plugin.db.batch.sink.*;
+              io.cdap.plugin.db.source.*;
```
albertshau, line 102: this looks like an existing issue, but we don't want multiple plugins to export the same package. This can cause classloading conflicts among the plugins and result in confusing errors like `cannot cast DBRecord to DBRecord` if multiple plugins with the same exports are used in the same pipeline. Normally we want to have io.cdap.plugin.[something unique].* as the only export for each plugin jar.

Is there something breaking if we don't include these?

aurora-postgresql-plugin/pom.xml
```diff
@@ -98,12 +98,12 @@
...
-              io.cdap.plugin.auroradb.postgres.*;
-              io.cdap.plugin.db.batch.source.*;
-              io.cdap.plugin.db.batch.sink.*;
-              org.apache.commons.lang;
-              org.apache.commons.logging.*;
-              org.codehaus.jackson.*
+                io.cdap.plugin.auroradb.postgres.*;
+                io.cdap.plugin.db.source.*;
+                io.cdap.plugin.db.sink.*;
+                org.apache.commons.lang;
```
albertshau, line 104: similar comments everywhere. With these exports it seems like things would break if we used multiple of them in the same pipeline.

albertshau (approved):
lgtm, though please open a jira to follow up on these exports. I'm pretty sure they would cause issues.

aurora-mysql-plugin/pom.xml
MrRahulSharma: Created JIRA : https://cdap.atlassian.net/browse/PLUGIN-1521 to track the said issue.

PR #361, opened 2023-03-02
Converting timestamptz type to String type

MrRahulSharma:
Related JIRA : https://cdap.atlassian.net/browse/PLUGIN-1531

### DIFF
--- a/postgresql-plugin/src/main/java/io/cdap/plugin/postgres/PostgresSchemaReader.java
+++ b/postgresql-plugin/src/main/java/io/cdap/plugin/postgres/PostgresSchemaReader.java
@@ -35,7 +35,7 @@ public class PostgresSchemaReader extends CommonSchemaReader {
   );
 
   public static final Set<String> STRING_MAPPED_POSTGRES_TYPES_NAMES = ImmutableSet.of(
-    "bit", "timetz", "money"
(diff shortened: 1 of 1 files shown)


albertshau (commented):
this is a backwards incompatible change, where existing pipelines that were expecting it to be the previous data type will start failing after upgrade.

In the past, we've handled this type of situation by implementing this such that if a schema is given, and the type is the previous type, we'll handle it the same way as before. Otherwise, we handle it the new way.

Also, why do we what to convert to string vs datetime?

MrRahulSharma:
> this is a backwards incompatible change, where existing pipelines that were expecting it to be the previous data type will start failing after upgrade.
> 
> In the past, we've handled this type of situation by implementing this such that if a schema is given, and the type is the previous type, we'll handle it the same way as before. Otherwise, we handle it the new way.
> 
> Also, why do we what to convert to string vs datetime?

This change works with both String and timestamptz type fields in the Sink table. Hence this change should be backward compatible.

albertshau:
as discussed offline, let's try to compile a list of all the changes we're planning to make and whether they are backwards compatible or not. If there's a whole bunch it may make sense to bump the major version and rename the plugins so that they cannot be upgraded from the old versions to the new.

PR #365, opened 2023-03-11
PLUGIN-1535 : Oracle decimal type Backward Compatibility Fix

MrRahulSharma:
Backward compatibility fix for Oracle number type with undefined precision and scale.
JIRA : https://cdap.atlassian.net/browse/PLUGIN-1535

### DIFF
--- a/oracle-plugin/docs/Oracle-batchsink.md
+++ b/oracle-plugin/docs/Oracle-batchsink.md
@@ -60,37 +60,38 @@ will be passed to the JDBC driver as connection arguments for JDBC drivers that 
 Data Types Mapping
 ----------
 
-    | Oracle Data Type               | CDAP Schema Data Type | Comment                                                |
-    | ------------------------------ | --------------------- | ------------------------------------------------------ |
-    | VARCHAR2                       | string                |                                                        |
-    | NVARCHAR2                      | string                |                                                        |
-    | VARCHAR                        | string                |                                                        |
-    | NUMBER                         | string                | For NUMBER types defined without a precision and scale |
-    | NUMBER                         | decimal               | For NUMBER types with a defined precision and scale    |
-    | FLOAT                          | double                |                                                        |
-    | LONG                           | string                |                                                        |
-    | DATE                           | timestamp             |                                                        |
-    | BINARY_FLOAT                   | float                 |                                                        |
-    | BINARY_DOUBLE                  | double                |                                                        |
-    | TIMESTAMP                      | timestamp             |                                                        |
-    | TIMESTAMP WITH TIME ZONE       | string                | Timestamp string in the following format:              |
-    |                                |                       | "2019-07-15 15:57:46.65 GMT"                           |
-    | TIMESTAMP WITH LOCAL TIME ZONE | timestamp             |                                                        |
-    | INTERVAL YEAR TO MONTH         | string                | Oracle's 'INTERVAL YEAR TO MONTH' literal in the       |
-    |                                |                       | standard format: "year[-month]"                        |
-    | INTERVAL DAY TO SECOND         | string                | Oracle's 'INTERVAL DAY TO SECOND' literal in the       |
-    |                                |                       | standard format:                                       |
-    |                                |                       | "[day] [hour][:minutes][:seconds[.milliseconds]"       |
-    | RAW                            | bytes                 |                                                        |
-    | LONG RAW                       | bytes                 |                                                        |
-    | ROWID                          | string                |                                                        |
-    | UROWID                         | string                |                                                        |
-    | CHAR                           | string                |                                                        |
-    | NCHAR                          | string                |                                                        |
-    | CLOB                           | string                |                                                        |
-    | NCLOB                          | string                |                                                        |
-    | BLOB                           | bytes                 |                                                        |
-    | BFILE                          |                       | BFILE data type is not supported for the sink          |
+    | Oracle Data Type               | CDAP Schema Data Type | Comment                                                    |
+    | ------------------------------ | --------------------- | -----------------------------------------------------------|
+    | VARCHAR2                       | string                |                                                            |
+    | NVARCHAR2                      | string                |                                                            |
+    | VARCHAR                        | string                |                                                            |
+    | NUMBER                         | string                | For NUMBER types defined without a precision and scale.    |
+    |                                |                       | Users can manually set output schema to map it to Decimal. |
+    | NUMBER                         | decimal               | For NUMBER types defined with a precision and scale.       |
+    | FLOAT                          | double                |                                                            |
+    | LONG                           | string                |                                                            |
+    | DATE                           | timestamp             |                                                            |
+    | BINARY_FLOAT                   | float                 |                                                            |
+    | BINARY_DOUBLE                  | double                |                                                            |
+    | TIMESTAMP                      | timestamp             |                                                            |
+    | TIMESTAMP WITH TIME ZONE       | string                | Timestamp string in the following format:                  |
+    |                                |                       | "2019-07-15 15:57:46.65 GMT"                               |
+    | TIMESTAMP WITH LOCAL TIME ZONE | timestamp             |                                                            |
+    | INTERVAL YEAR TO MONTH         | string                | Oracle's 'INTERVAL YEAR TO MONTH' literal in the           |
+    |                                |                       | standard format: "year[-month]"                            |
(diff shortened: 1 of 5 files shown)


tivv (commented):
Please add a unit test for the fallback and refer respective JIRA in the PR title

MrRahulSharma:
> Please add a unit test for the fallback and refer respective JIRA in the PR title

Added Unit Test to validate if the schema is Decimal and does not match with the expected String type.
Updated JIRA in the description.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSource.java
```diff
@@ -153,6 +154,17 @@ public void validate(FailureCollector collector) {
...
       return connection.getTransactionIsolationLevel();
     }
+
+    @Override
+    protected void validateField(FailureCollector collector,
+                                 Schema.Field field, Schema actualFieldSchema, Schema expectedFieldSchema) {
+      // Validate if the field schema is of type Decimal and the expected is String
+      if (Schema.LogicalType.DECIMAL.equals(expectedFieldSchema.getLogicalType())
+          && actualFieldSchema.getType().equals(Schema.Type.STRING)) {
+        return;
```
itsankit-google: The comment and the code does not match here: Comment says `expectedSchema` to be `String` and `actualSchema` but we are validating `actualSchema` to be `String`??

MrRahulSharma: Updated

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSource.java
```diff
@@ -153,6 +154,17 @@ public void validate(FailureCollector collector) {
...
     }
+
+    @Override
+    protected void validateField(FailureCollector collector,
+                                 Schema.Field field, Schema actualFieldSchema, Schema expectedFieldSchema) {
+      // Validate if the field output schema is set to Decimal and the actual evaluated schema is String.
+      if (Schema.LogicalType.DECIMAL.equals(expectedFieldSchema.getLogicalType())
+          && actualFieldSchema.getType().equals(Schema.Type.STRING)) {
+        return;
+      }
```
itsankit-google, line 169: Why is this change needed? When do we expect actual schema (schema in user's database) as `STRING` and output schema to be `DECIMAL`?

MrRahulSharma: This change is needed to make sure that the pipeline upgrade continues to work post upgrade. Since the older handling of the precision less used to convert to the decimal type, and the new version would try to convert to the String type. In that case the output schema would contain Decimal(38, 0) (or something similar), and the code internally would try to identify the schema of the field(without precision and scale) as String.

albertshau: would be good to put this explanation in the code comment

oracle-plugin/src/test/java/io/cdap/plugin/oracle/OracleSourceDBRecordUnitTest.java
```diff
@@ -95,6 +95,55 @@ public void validatePrecisionLessDecimalParsing() throws Exception {
     Assert.assertEquals(record.get("ID4"), "123.4568");
   }
 
+  /**
+   * Validate the precision less Numbers handling against following use cases.
+   * 1. Ensure that for Number(0,-127) non nullable type a String is returned by default.
+   * 2. Ensure that for Number(0,-127) non nullable type a String is returned,
```
albertshau: this javadoc seems incorrect, should be returning a BigDecimal in the cases where the schema is defined as a Decimal.

oracle-plugin/src/test/java/io/cdap/plugin/oracle/OracleSourceDBRecordUnitTest.java
```diff
@@ -95,6 +95,55 @@ public void validatePrecisionLessDecimalParsing() throws Exception {
     Assert.assertEquals(record.get("ID4"), "123.4568");
   }
 
+  /**
+   * Validate the precision less Numbers handling against following use cases.
+   * 1. Ensure that for Number(0,-127) non nullable type a String is returned by default.
```
albertshau: I don't see this behavior in the test, there are only asserts for BigDecimal values returned.

oracle-plugin/src/test/java/io/cdap/plugin/oracle/OracleSourceDBRecordUnitTest.java
```diff
@@ -95,6 +95,55 @@ public void validatePrecisionLessDecimalParsing() throws Exception {
...
+    Schema schema = Schema.recordOf(
+      "dbRecord",
+      field1,
+      field2,
+      field3,
+      field4
+    );
+
+    when(resultSet.getMetaData()).thenReturn(resultSetMetaData);
+    when(resultSet.getBigDecimal(eq(1), eq(0))).thenReturn(new BigDecimal("123"));
```
albertshau, line 120: do we also have any tests where we actually read from an oracle DB? That would give more confidence than mocking the return value. For example, I see an OracleSourceTestRun, is that being executed?

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSource.java
MrRahulSharma: Added.

oracle-plugin/src/test/java/io/cdap/plugin/oracle/OracleSourceDBRecordUnitTest.java
MrRahulSharma: Updated the comments.

oracle-plugin/src/test/java/io/cdap/plugin/oracle/OracleSourceDBRecordUnitTest.java
MrRahulSharma: Updated the comments

oracle-plugin/src/test/java/io/cdap/plugin/oracle/OracleSourceDBRecordUnitTest.java
MrRahulSharma: OracleSourceTestRun does not execute.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceDBRecord.java
```diff
@@ -223,8 +223,16 @@ private void handleOracleSpecificType(ResultSet resultSet, StructuredRecord.Buil
...
-            // In case of precision less decimal convert the field to String type
-            recordBuilder.set(field.getName(), resultSet.getString(columnIndex));
+            Schema nonNullableSchema = field.getSchema().isNullable() ?
+                                        field.getSchema().getNonNullable() : field.getSchema();
+            if (Schema.LogicalType.DECIMAL.equals(nonNullableSchema.getLogicalType())) {
+              // Handle the field using the schema set in the output schema
+              BigDecimal decimal = resultSet.getBigDecimal(columnIndex, getScale(field.getSchema()));
+              recordBuilder.setDecimal(field.getName(), decimal);
+            } else {
+              // In case of precision less decimal convert the field to String type
```
fernst: nit: "In case of a decimal with no precision/scale set, convert..."

fernst (commented):
Small nitpick

albertshau (approved):

oracle-plugin/docs/Oracle-batchsink.md
```diff
@@ -60,37 +60,38 @@ will be passed to the JDBC driver as connection arguments for JDBC drivers that
...
-    | NCLOB                          | string                |                                                        |
-    | BLOB                           | bytes                 |                                                        |
-    | BFILE                          |                       | BFILE data type is not supported for the sink          |
+    | Oracle Data Type               | CDAP Schema Data Type | Comment                                                    |
+    | ------------------------------ | --------------------- | -----------------------------------------------------------|
+    | VARCHAR2                       | string                |                                                            |
+    | NVARCHAR2                      | string                |                                                            |
+    | VARCHAR                        | string                |                                                            |
+    | NUMBER                         | string                | For NUMBER types defined without a precision and scale     |
```
robinrielley, line 68: add "defined" For NUMBER types without a defined precision and scale

itsankit-google (approved):

PR #386, opened 2023-04-24
[PLUGIN-1540][Plugin-1595] Added Support for proper handling of Timestamp, TimestampLTZ, TimestampTZ types

MrRahulSharma:
Changes to support following changes for Oracle types.

Timestamp: CDAP DateTime since it does not contain any timezone information.
TimestampLTZ : CDAP DateTime since it does not contain any timezone information.
TimestampTZ : CDAP Timestamp since it contains timezone information.

Issues getting fixed with this change:
1. TimestampTZ is currently mapped to CDAP String type, which also maps to BQ String type. After this PR the TimestampTZ will default map to CDAP Timestamp type which maps to BQ Timestamp type as well. This change also ensures backward compatible, and is done in order to ensure a consistent datatype handling in the Oracle plugin.
3. TimestampTZ type does not work in Oracle Sink node, due to format mismatch. After this change the pipeline will detect TimestampTZ type to CDAP Timestamp type which works as intended in Oracle Sink node. JIRA: https://cdap.atlassian.net/browse/PLUGIN-1540
4. For TimestampLTZ type if it has value like '0001-01-01 01:00:00.000', then it does not gets inserted into the BQ Timestamp type with an invalid instant value. This PR changes the mapping of TimestampLTZ type to CDAP Datetime type which fixes the BQ insert failure.
5. For TimestampLTZ type if the source value is '0001-01-01 09:00:00.000' then in preview the value becomes '0000-12-30T09:00:00Z[UTC]'. After this change it will show as '0001-01-01T09:00:00' which is as per the documentation of Oracle TimestampLTZ datatype. https://cdap.atlassian.net/browse/PLUGIN-1595

### DIFF
--- a/oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceDBRecord.java
+++ b/oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceDBRecord.java
@@ -16,6 +16,7 @@
 
 package io.cdap.plugin.oracle;
 
+import com.google.common.base.Strings;
 import com.google.common.io.ByteStreams;
 import io.cdap.cdap.api.common.Bytes;
 import io.cdap.cdap.api.data.format.StructuredRecord;
@@ -37,9 +38,11 @@ import java.sql.ResultSetMetaData;
 import java.sql.SQLException;
 import java.sql.Timestamp;
 import java.sql.Types;
-import java.time.Instant;
+import java.time.LocalDateTime;
+import java.time.OffsetDateTime;
 import java.time.ZoneId;
 import java.time.ZoneOffset;
+import java.time.ZonedDateTime;
 import java.util.List;
 import javax.annotation.Nullable;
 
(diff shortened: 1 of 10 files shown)


oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceDBRecord.java
```diff
@@ -116,20 +118,70 @@ protected void handleField(ResultSet resultSet, StructuredRecord.Builder recordB
   protected void writeToDB(PreparedStatement stmt, @Nullable Schema.Field field, int fieldIndex) throws SQLException {
     int sqlType = columnTypes.get(fieldIndex).getType();
     int sqlIndex = fieldIndex + 1;
-    switch (sqlType) {
-      case OracleSourceSchemaReader.TIMESTAMP_TZ:
-        if (field != null && record.get(field.getName()) != null) {
```
itsankit-google, line 121: why did we removed the `null` checks?

MrRahulSharma: Thats a miss, allow me to fix it.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceDBRecord.java
```diff
@@ -195,12 +294,53 @@ private void handleOracleSpecificType(ResultSet resultSet, StructuredRecord.Buil
         recordBuilder.set(field.getName(), resultSet.getString(columnIndex));
         break;
       case OracleSourceSchemaReader.TIMESTAMP_TZ:
-        recordBuilder.set(field.getName(), resultSet.getString(columnIndex));
+        if (Schema.Type.STRING.equals(nonNullSchema.getType())) {
+          recordBuilder.set(field.getName(), resultSet.getString(columnIndex));
+        } else {
+          // In case of TimestampTZ and TimestampLTZ datatypes the getTimestamp(index, Calendar) method call does not
```
MrRahulSharma: Remove TimestampLTZ

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceDBRecord.java
```diff
@@ -195,12 +294,53 @@ private void handleOracleSpecificType(ResultSet resultSet, StructuredRecord.Buil
...
+          // return a correct value for any year which is less than the gregorian cutover date. Due to which
+          // 0001-01-01 01:00:00 changes to 0000-12-30. In this case classes oracle.sql.TIMESTAMPTZ type
+          // need to be used to identify the correct timestamp instances.
+          Object timeStampObj = resultSet.getObject(columnIndex);
```
MrRahulSharma, line 311: Add driver documentation link here : https://docs.oracle.com/cd/E13222_01/wls/docs91/jdbc_drivers/oracle.html

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceDBRecord.java
```diff
@@ -195,12 +294,53 @@ private void handleOracleSpecificType(ResultSet resultSet, StructuredRecord.Buil
...
+        }
         break;
       case OracleSourceSchemaReader.TIMESTAMP_LTZ:
-        Timestamp timestamp = resultSet.getTimestamp(columnIndex);
-        recordBuilder.setTimestamp(field.getName(), (timestamp != null) ?
-                timestamp.toInstant().atZone(ZoneId.ofOffset("UTC", ZoneOffset.UTC)) : null);
+        if (Schema.LogicalType.DATETIME.equals(nonNullSchema.getLogicalType())) {
+          Timestamp timestampLTZ = resultSet.getTimestamp(columnIndex);
+          if (timestampLTZ != null) {
+            recordBuilder.setDateTime(field.getName(),
```
MrRahulSharma, line 351: Add an example here with explanation for the change.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceDBRecord.java
```diff
@@ -151,6 +217,43 @@ private Object createOracleTimestampWithTimeZone(Connection connection, String t
...
+      return timestampLTZClass.getConstructor(Connection.class, String.class).newInstance(connection, timestampString);
+    } catch (ClassNotFoundException e) {
+      throw new InvalidStageException("Unable to load 'oracle.sql.TIMESTAMPLTZ'.", e);
+    } catch (InstantiationException | InvocationTargetException | NoSuchMethodException | IllegalAccessException e) {
+      throw new InvalidStageException("Unable to instantiate 'oracle.sql.TIMESTAMPLTZ'.", e);
+    }
+  }
+
+  /**
+   * Creates an instance of 'oracle.sql.TIMESTAMP' which corresponds to the specified timestamp with local time zone.
```
itsankit-google: Is this comment correct? Looks like oracle TIMESTAMP does not contain timezone information.

```
Using the example format, specify TIMESTAMP as a literal as follows:
TIMESTAMP '1997-01-31 09:26:50.12'
```

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceDBRecord.java
```diff
@@ -151,6 +217,43 @@ private Object createOracleTimestampWithTimeZone(Connection connection, String t
...
+   * @param connection sql connection.
+   * @param timestampString timestamp with local time zone string, such as "2019-07-15 15:57:46.65 GMT".
```
itsankit-google: similarly the example here.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceDBRecord.java
MrRahulSharma: Agreed.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceDBRecord.java
MrRahulSharma: Agreed.

itsankit-google (approved):

PR #390, opened 2023-05-09
[Plugin-1614][Plugin-1507][Plugin-1615][Plugin-1611] Int datatype changes

MrRahulSharma:
JIRA
https://cdap.atlassian.net/browse/PLUGIN-1614
https://cdap.atlassian.net/browse/PLUGIN-1507
https://cdap.atlassian.net/browse/PLUGIN-1615
https://cdap.atlassian.net/browse/PLUGIN-1611

### DIFF
--- a/database-commons/src/main/java/io/cdap/plugin/db/DBRecord.java
+++ b/database-commons/src/main/java/io/cdap/plugin/db/DBRecord.java
@@ -212,10 +212,10 @@ public class DBRecord implements Writable, DBWritable, Configurable {
    * @param stmt the {@link PreparedStatement} to write the {@link StructuredRecord} to
    */
   public void write(PreparedStatement stmt) throws SQLException {
-    for (int i = 0; i < columnTypes.size(); i++) {
-      ColumnType columnType = columnTypes.get(i);
+    for (int fieldIndex = 0; fieldIndex < columnTypes.size(); fieldIndex++) {
+      ColumnType columnType = columnTypes.get(fieldIndex);
       Schema.Field field = record.getSchema().getField(columnType.getName());
-      writeToDB(stmt, field, i);
+      writeToDB(stmt, field, fieldIndex);
     }
   }
 
@@ -274,26 +274,54 @@ public class DBRecord implements Writable, DBWritable, Configurable {
     }
   }
 
-  protected void writeToDB(PreparedStatement stmt, @Nullable Schema.Field field, int fieldIndex) throws SQLException {
+  private void writeToDB(PreparedStatement stmt, @Nullable Schema.Field field, int fieldIndex) throws SQLException {
+    if (shouldWriteNullField(field)) {
+      writeNullToDB(stmt, fieldIndex);
+    } else {
+      Schema nonNullableSchema = getNonNullableSchema(field);
+      writeNonNullToDB(stmt, nonNullableSchema, field.getName(), fieldIndex);
+    }
+  }
 
+  /**
+   * This method returns true in case a field can support writeNullToDB for the current field.
+   * By default, this method returns true when field or field value is set to null.
+   *
+   * @param field Field
+   * @return true if null value of the field can be written to DB
+   */
+  protected boolean shouldWriteNullField(Schema.Field field) {
+    return (field == null || record.get(field.getName()) == null);
+  }
+
+  /**
+   * This method handle the null field and null field values, by internally using the PreparedStatement.setNull
+   * method. Any class requiring a custom handling to write null value for any type should override this method.
+   *
+   * @param stmt       PreparedStatement object for writing to db
+   * @param fieldIndex Field index in the columnTypes
+   * @throws SQLException Exception while calling PreparedStatement.setNull
+   */
+  protected void writeNullToDB(PreparedStatement stmt, int fieldIndex) throws SQLException {
     int sqlIndex = fieldIndex + 1;
     int sqlType = columnTypes.get(fieldIndex).getType();
-    if (field == null) {
-      // Some of the fields can be absent in the record
-      stmt.setNull(sqlIndex, sqlType);
-      return;
-    }
+    stmt.setNull(sqlIndex, sqlType);
+  }
 
-    String fieldName = field.getName();
-    Schema fieldSchema = getNonNullableSchema(field);
-    Schema.Type fieldType = fieldSchema.getType();
+  /**
+   * Write Non null values to DB.
+   *
+   * @param stmt          PreparedStatement object for writing to db
+   * @param fieldSchema   Non-Nullable schema of the field
+   * @param fieldName     Current Field from record's schema
+   * @param fieldIndex    Field index in the columnTypes
+   * @throws SQLException Exception while calling PreparedStatement.set... calls
+   */
+  protected void writeNonNullToDB(PreparedStatement stmt, Schema fieldSchema,
+                                  String fieldName, int fieldIndex) throws SQLException {
+
+    int sqlIndex = fieldIndex + 1;
     Schema.LogicalType fieldLogicalType = fieldSchema.getLogicalType();
-    Object fieldValue = record.get(fieldName);
-
-    if (fieldValue == null) {
-      stmt.setNull(sqlIndex, columnTypes.get(fieldIndex).getType());
-      return;
-    }
 
     if (fieldLogicalType != null) {
       switch (fieldLogicalType) {
@@ -318,6 +346,8 @@ public class DBRecord implements Writable, DBWritable, Configurable {
       return;
     }
 
+    Schema.Type fieldType = fieldSchema.getType();
+    Object fieldValue = record.get(fieldName);
     switch (fieldType) {
       case NULL:
         stmt.setNull(sqlIndex, columnTypes.get(fieldIndex).getType());
@@ -330,8 +360,7 @@ public class DBRecord implements Writable, DBWritable, Configurable {
         stmt.setBoolean(sqlIndex, (Boolean) fieldValue);
         break;
       case INT:
-        // write short or int appropriately
-        writeInt(stmt, fieldIndex, sqlIndex, fieldValue);
+        stmt.setInt(sqlIndex, (Integer) fieldValue);
         break;
       case LONG:
         long fieldValueLong = ((Number) fieldValue).longValue();
@@ -362,16 +391,6 @@ public class DBRecord implements Writable, DBWritable, Configurable {
     }
     // handles BINARY, VARBINARY and LOGVARBINARY
     stmt.setBytes(sqlIndex, byteValue);
-  }
-
-  protected void writeInt(PreparedStatement stmt, int fieldIndex, int sqlIndex, Object fieldValue) throws SQLException {
-    Integer intValue = (Integer) fieldValue;
-    int parameterType = columnTypes.get(fieldIndex).getType();
-    if (Types.TINYINT == parameterType || Types.SMALLINT == parameterType) {
-      stmt.setShort(sqlIndex, intValue.shortValue());
-      return;
-    }
-    stmt.setInt(sqlIndex, intValue);
   }
 
   @Override
--- a/mysql-plugin/docs/Mysql-batchsink.md
+++ b/mysql-plugin/docs/Mysql-batchsink.md
@@ -66,37 +66,40 @@ connections.
 
 Data Types Mapping
 ----------
-| MySQL Data Type                | CDAP Schema Data Type | Comment                                                 |
-| ------------------------------ |-----------------------|---------------------------------------------------------|
-| BIT                            | boolean               |                                                         |
-| TINYINT                        | int                   |                                                         |
-| BOOL, BOOLEAN                  | boolean               |                                                         |
-| SMALLINT                       | int                   |                                                         |
-| MEDIUMINT                      | double                |                                                         |
-| INT,INTEGER                    | int                   |                                                         |
-| BIGINT                         | long                  |                                                         |
-| FLOAT                          | float                 |                                                         |
-| DOUBLE                         | double                |                                                         |
-| DECIMAL                        | decimal               |                                                         |
-| DATE                           | date                  |                                                         |
-| DATETIME                       | timestamp             |                                                         |
-| TIMESTAMP                      | timestamp             |                                                         |
-| TIME                           | time                  |                                                         |
-| YEAR                           | int                   | Users can manually set output schema to map it to Date. |
-| CHAR                           | string                |                                                         |
-| VARCHAR                        | string                |                                                         |
-| BINARY                         | bytes                 |                                                         |
-| VARBINARY                      | bytes                 |                                                         |
-| TINYBLOB                       | bytes                 |                                                         |
-| TINYTEXT                       | string                |                                                         |
-| BLOB                           | bytes                 |                                                         |
-| TEXT                           | string                |                                                         |
-| MEDIUMBLOB                     | bytes                 |                                                         |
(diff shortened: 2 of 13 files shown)


Aryan-Verma (approved):

Aryan-Verma (approved):

saimukkamala (approved):

mysql-plugin/src/main/java/io/cdap/plugin/mysql/MysqlConnectorConfig.java
```diff
@@ -58,6 +60,8 @@ public Properties getConnectionArgumentsProperties() {
     prop.put(JDBC_PROPERTY_CONNECT_TIMEOUT, "20000");
     prop.put(JDBC_PROPERTY_SOCKET_TIMEOUT, "20000");
     prop.put(JDBC_REWRITE_BATCHED_STATEMENTS, "true");
+    // MySQL property to ensure that TINYINT(1) type data is not converted to MySQL Bit/Boolean type in the ResultSet.
+    prop.put(MYSQL_TINYINT1_IS_BIT, "false");
```
itsankit-google, line 64: This is a backward incompatible change right?

itsankit-google: On a second thought, we already support adding connection arguments whenever needed why do we need to change the default behaviour?

MrRahulSharma: Without this connection argument, for Tinyint(1) type MySQL driver code internally converts these type into boolean type, Due to which any value between [-128, 127] becomes true/false(for 0).

itsankit-google: But this is the issue with mysql driver not the plugin what if the existing pipelines are depending on the boolean conversion?

Any customer want to change that behaviour they can always set `MYSQL_TINYINT1_IS_BIT, "false"` in connection arguments.

@albertshau @tivv wdyt??

MrRahulSharma: Driver details about this is present here : https://dev.mysql.com/doc/connector-j/8.0/en/connector-j-reference-type-conversions.html#:~:text=TINYINT(1)%20SIGNED,java.lang.Integer

mysql-plugin/docs/Mysql-batchsource.md
```diff
@@ -85,14 +85,14 @@ with the tradeoff of higher memory usage.
...
+| ------------------------------ |-----------------------|---------------------------------------------------------|
 | BIT                            | boolean               |                                                         |
 | TINYINT                        | int                   |                                                         |
 | BOOL, BOOLEAN                  | boolean               |                                                         |
 | SMALLINT                       | int                   |                                                         |
-| MEDIUMINT                      | double                |                                                         |
+| MEDIUMINT                      | int                   | Users can manually set output schema to map it to Long. |
 | INT,INTEGER                    | int                   |                                                         |
-| BIGINT                         | long                  |                                                         |
+| BIGINT                         | long, decimal         | BIGINT UNSIGNED are converted to Decimal type.          |
```
albertshau: think it would be more clear to 2 rows -- BIGINT (signed) and BIGINT (unsigned)

mysql-plugin/docs/Mysql-batchsink.md
```diff
@@ -72,9 +72,9 @@ Data Types Mapping
 | TINYINT                        | int                   |                                                         |
 | BOOL, BOOLEAN                  | boolean               |                                                         |
 | SMALLINT                       | int                   |                                                         |
-| MEDIUMINT                      | double                |                                                         |
+| MEDIUMINT                      | int                   | Users can manually set output schema to map it to Long. |
 | INT,INTEGER                    | int                   |                                                         |
-| BIGINT                         | long                  |                                                         |
+| BIGINT                         | long, decimal         | BIGINT UNSIGNED are converted to Decimal type.                                                        |
```
albertshau: would be more clear to have 2 rows, one for signed and one for unsigned

mysql-plugin/docs/Mysql-batchsink.md
```diff
@@ -72,9 +72,9 @@ Data Types Mapping
 | TINYINT                        | int                   |                                                         |
 | BOOL, BOOLEAN                  | boolean               |                                                         |
 | SMALLINT                       | int                   |                                                         |
-| MEDIUMINT                      | double                |                                                         |
+| MEDIUMINT                      | int                   | Users can manually set output schema to map it to Long. |
```
albertshau: what does this comment mean? There is no output schema for a sink.

database-commons/src/main/java/io/cdap/plugin/db/DBRecord.java
```diff
@@ -364,16 +363,6 @@ protected void writeBytes(PreparedStatement stmt, int fieldIndex, int sqlIndex,
     stmt.setBytes(sqlIndex, byteValue);
   }
 
-  protected void writeInt(PreparedStatement stmt, int fieldIndex, int sqlIndex, Object fieldValue) throws SQLException {
```
albertshau, line 367: why are we removing this? Does the driver do the right thing if we do setInt() for tinyint/smallint columns?

mysql-plugin/docs/Mysql-batchsink.md
albertshau: There is similar existing comment for YEAR. Seems like this table was just copied from the source? For a sink I would expect to find what CDAP types can be written to which MySQL types

mysql-plugin/docs/Mysql-batchsource.md
```diff
@@ -85,14 +85,14 @@ with the tradeoff of higher memory usage.
...
 ----------
 | MySQL Data Type                | CDAP Schema Data Type | Comment                                                 |
-| ------------------------------ | --------------------- |---------------------------------------------------------|
+| ------------------------------ |-----------------------|---------------------------------------------------------|
 | BIT                            | boolean               |                                                         |
 | TINYINT                        | int                   |                                                         |
 | BOOL, BOOLEAN                  | boolean               |                                                         |
 | SMALLINT                       | int                   |                                                         |
-| MEDIUMINT                      | double                |                                                         |
+| MEDIUMINT                      | int                   | Users can manually set output schema to map it to Long. |
```
albertshau: should Long in the comment be double? Or was the previous 'double' type a documentation bug?

mysql-plugin/src/main/java/io/cdap/plugin/mysql/MysqlConnectorConfig.java
albertshau: what was the previous behavior? Would we detect a tinyint(1) column as a CDAP int or boolean?

If we detected it as an CDAP int, then I imagine we were setting a boolean for an int field in the StructuredRecord? If so that was a bug and we can just do this fix without worrying about compatibility.

If we detected it and read it as a CDAP boolean, then this is a backwards incompatible change and shouldn't be done.

mysql-plugin/src/main/java/io/cdap/plugin/mysql/MysqlDBRecord.java
```diff
@@ -74,7 +74,12 @@ protected void writeToDB(PreparedStatement stmt, @Nullable Schema.Field field, i
       case Types.DATE:
         Schema schema = field.getSchema().isNullable() ? field.getSchema().getNonNullable() : field.getSchema();
         if (schema.getType().equals(Schema.Type.INT) && !Schema.LogicalType.DATE.equals(schema.getLogicalType())) {
-          stmt.setInt(sqlIndex, record.get(field.getName()));
+          Object fieldValue = record.get(field.getName());
+          if (fieldValue == null) {
```
albertshau: it doesn't let us pass a null value with setInt()? 

It seems error-prone to force each implementation to check for a null value, can this be done in the parent class?

mysql-plugin/src/main/java/io/cdap/plugin/mysql/MysqlSchemaReader.java
```diff
@@ -48,12 +48,21 @@ public Schema getSchema(ResultSetMetaData metadata, int index) throws SQLExcepti
...
+    switch(sqlType) {
+      case Types.DATE:
+        // YEAR type in MySQL should get converted to integer to avoid truncation
+        // failures in the MySQL Sink plugin due to missing date and month details.
+        if (YEAR_TYPE_NAME.equalsIgnoreCase(sqlTypeName)) {
+          return Schema.of(Schema.Type.INT);
+        }
+        break;
+      case Types.INTEGER:
+        // MEDIUMINT UNSIGNED should be mapped to int
```
albertshau, line 60: why is MEDIUMINT UNSIGNED a special case, shouldn't a signed mediumint also be an int?

mysql-plugin/src/main/java/io/cdap/plugin/mysql/MysqlConnectorConfig.java
MrRahulSharma: The previous behavior was that tinyint(1) gets detected as CDAP boolean. 
This change can be done in the backward compatible way. 
@albertshau Should the default behavior be that tinyint(1) be detected as int instead of a bool ?

database-commons/src/main/java/io/cdap/plugin/db/DBRecord.java
MrRahulSharma: setShort does not handle max value for smallint unsigned type. 
Yes, setInt internally does the right thing for tinyint/smallint types, hence intended to clear the logic a bit.

mysql-plugin/docs/Mysql-batchsource.md
MrRahulSharma: It was a documentation bug. MediumInt used to get detected as int and Mediumint unsigned as long.

mysql-plugin/src/main/java/io/cdap/plugin/mysql/MysqlDBRecord.java
MrRahulSharma: Yes, Statement is A SQL type, different drivers can have this differently. This code segment was a miss in the previous change and got detected as part of E2E tests addition in this PR.

mysql-plugin/src/main/java/io/cdap/plugin/mysql/MysqlSchemaReader.java
MrRahulSharma: In the DBUtils we have a central logic of converting all Int unsigned types to long. But since Mysql Mediumint unsigned type have same sql type as that of int, hence it also gets converted to long type. This cause a failure in BigQuery sink during the serialization to json type. Since int can contain the mediumint unsigned min and max value hence it should ideally be mapped to int.

mysql-plugin/docs/Mysql-batchsink.md
MrRahulSharma: Currently, all the Sink documents present in the database-plugins project contain a copy of mapping details from their source plugins. I believe it might be a little tricky in adding a CDAP to sink type mapping, given we have multiple types mapped to single sink types. I feel that might be a bigger documentation change, which might require some more thoughts.

mysql-plugin/src/main/java/io/cdap/plugin/mysql/MysqlConnectorConfig.java
albertshau: I think it makes more sense for it to be an int by default, since there is a separate boolean type.

mysql-plugin/docs/Mysql-batchsink.md
albertshau: ok, though can we remove the comments column in the sink doc? I think it's pretty confusing.

mysql-plugin/src/main/java/io/cdap/plugin/mysql/MysqlSchemaReader.java
albertshau: ok I see. It would be good to put that in the code comment.

mysql-plugin/src/main/java/io/cdap/plugin/mysql/MysqlDBRecord.java
albertshau: it looks like there is already logic for this in the parent (https://github.com/data-integrations/database-plugins/blob/f717eb2b1a0a0cc4726b365f11d8bbe19d3d6989/database-commons/src/main/java/io/cdap/plugin/db/DBRecord.java#L293). It's also doing the getNonNullable() logic for the schema.

To avoid duplicate code, it would be better to refactor the writeToDB() method in DBRecord into 2 methods. The first part does the null handling and then calls some new method (i.e writeNonNullToDB()). That way children don't need to duplicate the logic.

mysql-plugin/docs/Mysql-batchsink.md
MrRahulSharma: I will work on this comment in the next set of changes, proposing to take this after this PR submission.

database-commons/src/main/java/io/cdap/plugin/db/DBRecord.java
```diff
@@ -274,26 +274,57 @@ private void writeToDataOut(DataOutput out, Schema.Field field) throws IOExcepti
...
+   * @param field         Current Field from record's schema
+   * @param fieldIndex    Field index in the columnTypes
+   * @throws SQLException Exception while calling PreparedStatement.setNull
+   */
+  protected void writeNonNullToDB(PreparedStatement stmt,
+                                  @Nullable Schema.Field field, int fieldIndex) throws SQLException {
+
+    int sqlIndex = fieldIndex + 1;
     String fieldName = field.getName();
     Schema fieldSchema = getNonNullableSchema(field);
```
albertshau: it would be good to do this in the caller and then pass in the fieldName and Schema as arguments instead of the Schema.Field.

That way anything that implements this method doesn't need to do the getNonNullable logic, which can be pretty easy to forget to do.

albertshau (commented):
minor suggestion

albertshau (commented):
minor suggestion

database-commons/src/main/java/io/cdap/plugin/db/DBRecord.java
```diff
@@ -274,26 +274,54 @@ private void writeToDataOut(DataOutput out, Schema.Field field) throws IOExcepti
...
+   * Write Non null values to DB.
+   *
+   * @param stmt          PreparedStatement object for writing to db
+   * @param fieldSchema   Non-Nullable schema of the field
+   * @param fieldName     Current Field from record's schema
+   * @param fieldIndex    Field index in the columnTypes
+   * @throws SQLException Exception while calling PreparedStatement.set... calls
+   */
+  protected void writeNonNullToDB(PreparedStatement stmt, Schema fieldSchema,
+                                  @Nullable String fieldName, int fieldIndex) throws SQLException {
```
albertshau: field name should not be nullable

albertshau (approved):
minor comment, lgtm

mysql-plugin/src/main/java/io/cdap/plugin/mysql/MysqlConnectorConfig.java
itsankit-google, line 64: Does it makes sense to use `putIfAbsent()` so that if user has provided it in connection args in plugin config we don't end up overriding it?

PR #404, opened 2023-06-13
[Plugin 1522] Enabling support for the precision less numbers in PostgreSQL plugin.

MrRahulSharma:
Converting the precision less numbers into String type and ensuring that they are compatible with the Postgres Sink as well.
Related JIRA :
https://cdap.atlassian.net/browse/PLUGIN-1522

### DIFF
--- a/postgresql-plugin/pom.xml
+++ b/postgresql-plugin/pom.xml
@@ -54,10 +54,6 @@
       <scope>test</scope>
     </dependency>
     <dependency>
-      <groupId>org.mockito</groupId>
-      <artifactId>mockito-core</artifactId>
-    </dependency>
-    <dependency>
       <groupId>io.cdap.plugin</groupId>
       <artifactId>database-commons</artifactId>
       <version>${project.version}</version>
@@ -73,10 +69,6 @@
       <artifactId>cdap-data-pipeline3_2.12</artifactId>
     </dependency>
     <dependency>
-      <groupId>junit</groupId>
-      <artifactId>junit</artifactId>
-    </dependency>
-    <dependency>
       <groupId>io.cdap.cdap</groupId>
       <artifactId>cdap-api</artifactId>
       <scope>provided</scope>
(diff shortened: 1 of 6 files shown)


postgresql-plugin/src/main/java/io/cdap/plugin/postgres/PostgresDBRecord.java
```diff
@@ -51,22 +53,30 @@ public PostgresDBRecord() {
...
                              int columnIndex, int sqlType, int sqlPrecision, int sqlScale) throws SQLException {
-    if (isUseSchema(resultSet.getMetaData(), columnIndex)) {
+    ResultSetMetaData metadata = resultSet.getMetaData();
+    if (isUseSchema(metadata, columnIndex)) {
       setFieldAccordingToSchema(resultSet, recordBuilder, field, columnIndex);
     } else {
+      // For the precision less number use the string type
+      int columnType = metadata.getColumnType(columnIndex);
+      int precision = metadata.getPrecision(columnIndex);
+      if (columnType == Types.NUMERIC && precision == 0) {
```
itsankit-google: is this backward compatible like we did for oracle plugins, the user should be able to manually modify/cast the output schema to `decimal`?

MrRahulSharma: Added the code changes for backward compatibility.

postgresql-plugin/pom.xml
```diff
@@ -87,6 +79,16 @@
       <version>RELEASE</version>
       <scope>compile</scope>
     </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <scope>compile</scope>
```
itsankit-google: `scope` should be `test` for dependencies used only in unit tests.

postgresql-plugin/pom.xml
```diff
@@ -87,6 +79,16 @@
...
+    </dependency>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <scope>compile</scope>
```
itsankit-google: similarly here

itsankit-google (approved):
Minor comments, otherwise LGTM.

PR #414, opened 2023-06-30
PLUGIN-1640: add support for using different ports in cloud sql proxy VM

itsankit-google:
Why:
- The Cloud SQL Auth Proxy can listen on different ports except the default port when using private IP to connect to the Cloud SQL instance.

Tested in CDAP Sandbox.

### DIFF
--- a/cloudsql-mysql-plugin/src/main/java/io/cdap/plugin/cloudsql/mysql/CloudSQLMySQLAction.java
+++ b/cloudsql-mysql-plugin/src/main/java/io/cdap/plugin/cloudsql/mysql/CloudSQLMySQLAction.java
@@ -18,11 +18,13 @@ package io.cdap.plugin.cloudsql.mysql;
 
 import com.google.common.collect.ImmutableMap;
 import io.cdap.cdap.api.annotation.Description;
+import io.cdap.cdap.api.annotation.Macro;
 import io.cdap.cdap.api.annotation.Name;
 import io.cdap.cdap.api.annotation.Plugin;
 import io.cdap.cdap.etl.api.FailureCollector;
 import io.cdap.cdap.etl.api.PipelineConfigurer;
 import io.cdap.cdap.etl.api.action.Action;
+import io.cdap.plugin.db.ConnectionConfig;
 import io.cdap.plugin.db.action.AbstractDBAction;
 import io.cdap.plugin.db.action.QueryConfig;
 import io.cdap.plugin.util.CloudSQLUtil;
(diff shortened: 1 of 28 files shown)


cloudsql-mysql-plugin/widgets/CloudSQLMySQL-action.json
```diff
@@ -74,11 +74,11 @@
           }
         },
         {
-          "widget-type": "textbox",
-          "label": "Instance Name",
```
itsankit-google, line 78: Removed the duplicate widget.

cloudsql-mysql-plugin/src/main/java/io/cdap/plugin/cloudsql/mysql/CloudSQLMySQLAction.java
```diff
@@ -103,6 +112,10 @@ public String getConnectionString() {
           connectionName);
     }
 
+    public Integer getPort() {
```
albertshau: since this can't return null, Integer -> int

albertshau: need to be careful with how this is used though. If the port is a macro, it will be null at configure time and this will return 3306. If we do any connectivity validation at configure time, it will need to be skipped if the port is a macro, otherwise it will try to connect on the wrong port.

it may be good to actually make this nullable and have it return null if it contains a macro. That way the caller is forced to deal with null instead of potentially misusing.

cloudsql-postgresql-plugin/src/main/java/io/cdap/plugin/cloudsql/postgres/CloudSQLPostgreSQLAction.java
```diff
@@ -103,6 +112,10 @@ public String getConnectionString() {
           connectionName);
     }
 
+    public Integer getPort() {
```
albertshau: similar comment about int and being careful about macros

cloudsql-mysql-plugin/src/main/java/io/cdap/plugin/cloudsql/mysql/CloudSQLMySQLAction.java
itsankit-google: I have overridden the `canConnect()` method which return false if any of the connection property is macro.

Does it makes sense? wdyt?

albertshau (approved):

PR #415, opened 2023-07-04
Mssql batch source and sink documentation changes

Aryan-Verma:
Added documentation for recent changes in SmallDateTime and DateTimeOffset handling in SQL Server batch source and batch sink plugins.

Also removed incorrect mapping of datetime types to string output type.


MrRahulSharma:
What about the sink documentation ?

MrRahulSharma (approved):

PR #423, opened 2023-07-28
PLUGIN-1654

Shubhangi-cs:
**BugFix**
Issue summary: Customer has a pipeline that is trying to load a table from bigquery into a netezza database that is failing.
This pipeline has exactly 52000 records and will not load, nor will it provide any usable error message.
1. When customer loads half the table first, then the second half later, it works.
2. When customer loads 51999 records, it works.
3. When customer loads 52001 (added a null record), it works.
4. But the load for 52000 does not work.
CDAP : https://cdap.atlassian.net/browse/PLUGIN-1654

Solution: 
Added proper handling for empty data flag. In case of empty data, batch will not be executed in close method.

### DIFF
--- a/database-commons/src/main/java/io/cdap/plugin/db/sink/ETLDBOutputFormat.java
+++ b/database-commons/src/main/java/io/cdap/plugin/db/sink/ETLDBOutputFormat.java
@@ -97,8 +97,8 @@ public class ETLDBOutputFormat<K extends DBWritable, V> extends DBOutputFormat<K
           try {
             if (!emptyData) {
               getStatement().executeBatch();
-              getConnection().commit();
             }
+            getConnection().commit();
           } catch (SQLException e) {
             try {
               getConnection().rollback();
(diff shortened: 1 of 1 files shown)


Aryan-Verma (approved):
Please merge after atleast 2 approvals

database-commons/src/main/java/io/cdap/plugin/db/sink/ETLDBOutputFormat.java
```diff
@@ -97,8 +97,8 @@ public void close(TaskAttemptContext context) throws IOException {
           try {
             if (!emptyData) {
               getStatement().executeBatch();
-              getConnection().commit();
             }
+            getConnection().commit();
           } catch (SQLException e) {
             try {
               getConnection().rollback();
```
MrRahulSharma, line 104: We should ideally separate out the executeBatch and commit calls into two separate try-catches to ensure that we commit the previously executed batches and not rollback unnecessarily in case of exception(s) occurring at executeBatch.

Shubhangi-cs: Changes done, in this case we are not committing the last batch if error occurs and pipeline will get successful with wrong no of records, there will be a warning only.

MrRahulSharma: Ohh I did not realise that this is a shared code across different sinks. Please revert to the older state of the code.

Shubhangi-cs: Done

MrRahulSharma (approved):

MrRahulSharma:
Correct the indentation and squash the commits.

PR #434, opened 2023-08-28
Version bump, required only in develop

MrRahulSharma:
Version bump in database-plugins for hydrator-plugin dependency.

Changes in Hydrator plugin which is required here : https://github.com/cdapio/hydrator-plugins/pull/1806

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -60,7 +60,7 @@
     <surefire.redirectTestOutputToFile>true</surefire.redirectTestOutputToFile>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <!-- version properties -->
-    <cdap.version>6.10.0-SNAPSHOT</cdap.version>
+    <cdap.version>6.9.1</cdap.version>
     <cdap.plugin.version>2.12.0-SNAPSHOT</cdap.plugin.version>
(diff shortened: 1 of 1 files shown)


itsankit-google (commented):
Why do we require this change? 

It is not good to bump the dependency version to a snapshot unless the plugin is using a new feature added to the platform

MrRahulSharma:
> Why do we require this change?
> 
> It is not good to bump the dependency version to a snapshot unless the plugin is using a new feature added to the platform

Removed the version bump for CDAP, making bump for hydrator only to ingest new changes from it.

itsankit-google:
> > Why do we require this change?
> > It is not good to bump the dependency version to a snapshot unless the plugin is using a new feature added to the platform
> 
> Removed the version bump for CDAP, making bump for hydrator only to ingest new changes from it.

Can you plz add a link to the new changes which need to be used in the PR description?

itsankit-google (changes requested):
All `E2E tests` are failing after this change, we should make sure `E2E tests` build is green before merging.

MrRahulSharma:
> All `E2E tests` are failing after this change, we should make sure `E2E tests` build is green before merging.

The failures seems independent from the change.

PR #433, opened 2023-08-25
[PLUGIN-1522] Enabling support for the precision less numbers in PostgreSQL plugin for release/1.8

Shubhangi-cs:
Converting the precision less numbers into String type and ensuring that they are compatible with the Postgres Sink as well.
CherryPicked from the PR : https://github.com/data-integrations/database-plugins/pull/404
One Pager : https://docs.google.com/document/d/1JpD52aIQ6VxZvVt5NhnBtdftQhPehMdO6SBQt-yfccA/edit
Related JIRA :
https://cdap.atlassian.net/browse/PLUGIN-1522

PR #444, opened 2023-09-04
Changes in Postgres_Precision_Fix - Sync up with Releases

Shubhangi-cs:
Converting the precision less numbers into String type and ensuring that they are compatible with the Postgres Sink as well.
changes from the comments on PR : https://github.com/data-integrations/database-plugins/pull/433


MrRahulSharma (approved):
LGTM

PR #452, opened 2023-09-12
 [PLUGIN-1680] numeric_precision_fix

Shubhangi-cs:
The mapping of source to sink should happen in case of numeric data type in which it is being converted to string in case of 0 precision. Previously it was fetching nullable schema and schema type was getting as Union in place of string. Now it is fetching nonNullable schema with the expected data type and schema is mapped from source to sink.
https://cdap.atlassian.net/browse/PLUGIN-1680


MrRahulSharma (approved):
LGTM

PR #476, opened 2024-01-02
ConnectionString calculation refactoring

MrRahulSharma:
Code Refactor for ConnectionString evaluation logic and case insensitive comparison.

### DIFF
--- a/oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleConstants.java
+++ b/oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleConstants.java
@@ -16,6 +16,8 @@
 
 package io.cdap.plugin.oracle;
 
+import javax.annotation.Nullable;
+
 /**
  * Oracle Constants.
  */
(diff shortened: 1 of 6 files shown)


oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleConstants.java
```diff
@@ -36,4 +36,28 @@ private OracleConstants() {
...
+   * @param database Database to connect to
+   * @return Connection String based on the given ConnectionType
+   */
+  public static String getConnectionString(String connectionType,
+                                           String host,
+                                           int port,
+                                           String database) {
+    if (OracleConstants.TNS_CONNECTION_TYPE.equalsIgnoreCase(connectionType)) {
+      return String.format(OracleConstants.ORACLE_CONNECTION_STRING_TNS_FORMAT, database);
+    } else if (OracleConstants.SERVICE_CONNECTION_TYPE.equalsIgnoreCase(connectionType)) {
```
itsankit-google: I guess `else` is not required here since we are returning if `if-clause` is `true`.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleConstants.java
```diff
@@ -36,4 +36,28 @@ private OracleConstants() {
...
+      return String.format(OracleConstants.ORACLE_CONNECTION_STRING_SERVICE_NAME_FORMAT,
+          host, port, database);
+    } else {
```
itsankit-google: similarly `else` is not required.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleConstants.java
```diff
@@ -36,4 +36,28 @@ private OracleConstants() {
...
+   *
+   * @param connectionType SID/Service/SID
+   * @param host Host name of the oracle server
+   * @param port Port of the oracle server
+   * @param database Database to connect to
+   * @return Connection String based on the given ConnectionType
+   */
+  public static String getConnectionString(String connectionType,
+                                           String host,
+                                           int port,
```
itsankit-google: Are `host` and `port` nullable when connection is of type `TNS`?

MrRahulSharma: Good catch, they can be null in some cases.

itsankit-google (approved):

PR #485, opened 2024-02-15
Redshift source and connector plugin added.

Shubhangi-cs:
Redshift source and connector plugin added.
CherryPicked from the main PR https://github.com/data-integrations/database-plugins/pull/471

PR #492, opened 2024-03-19
[Cherrypicked] Redshift source and connector plugin added.

vikasrathee-cs:
Added Redshift Source plugin for 6.9.x versions.
Cherrypicked from https://github.com/data-integrations/database-plugins/pull/485


database-commons/src/main/java/io/cdap/plugin/util/DBUtils.java
```diff
@@ -59,7 +59,7 @@
 public final class DBUtils {
   private static final Logger LOG = LoggerFactory.getLogger(DBUtils.class);
 
-  private static final Calendar PURE_GREGORIAN_CALENDAR = createPureGregorianCalender();
+  public static final Calendar PURE_GREGORIAN_CALENDAR = createPureGregorianCalender();
```
MrRahulSharma, line 62: Why this change ?

vikasrathee-cs: this was getting used in one of the class and was public in 1.11 release

itsankit-google: If it is not getting used in any of the redshift class related to your change, let's revert. Otherwise LGTM.

itsankit-google (approved):

database-commons/src/main/java/io/cdap/plugin/util/DBUtils.java
vikasrathee-cs: It is getting used.

PR #499, opened 2024-06-17
[PLUGIN-1793] Oracle sink plugin issue with small case

vikasrathee-cs:
https://cdap.atlassian.net/browse/PLUGIN-1793
Oracle sink plugin failing with null error when input schema contains small case letters.

Note: This is not a cherry pick as develop branch contains different methods but fix is similar.

Develop PR: #497


itsankit-google (changes requested):
Please also add a PR with unit test in develop and cherry-pick the commit with this PR.

itsankit-google (approved):

PR #512, opened 2024-09-18
Add support for TCPS based connection string to support SSL/TLS connection.

Vipinofficial11:


itsankit-google (changes requested):
The changes also need to be done in `widgets.json` for oracle plugins.

Example: https://github.com/data-integrations/database-plugins/blob/a83eaac204fdff5f9a81fbacc938114d040dd844/oracle-plugin/widgets/Oracle-batchsource.json#L120

How was this change tested?

PR #514, opened 2024-10-07
[PLUGIN-1813] CloudSQLMySQL Escape column names

psainics:
## CloudSQLMySQL Escape column names

Jira : [PLUGIN-1813](https://cdap.atlassian.net/browse/PLUGIN-1813)

### Description

If column name is a MySQL reserved keyword say `insert` the query will fail due to incorrect sql syntax.
This PR fixes this issue by escaping the column names using back ticks.

Similar column escapes is used in other plugins. Ref [CloudSQLPostgreSQLSink](https://github.com/data-integrations/database-plugins/blob/a83eaac204fdff5f9a81fbacc938114d040dd844/cloudsql-postgresql-plugin/src/main/java/io/cdap/plugin/cloudsql/postgres/CloudSQLPostgreSQLSink.java#L107-L118) 

### Code change

- Modified `CloudSQLMySQLSink.java`
- Modified `MysqlSink.java`

### Unit Tests

- Added `CloudSQLMySQLSinkTest.java`
- image
- Added `MysqlSinkTest.java`
- image

[PLUGIN-1813]: https://cdap.atlassian.net/browse/PLUGIN-1813?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/mysql-plugin/src/main/java/io/cdap/plugin/mysql/MysqlDBRecord.java
+++ b/mysql-plugin/src/main/java/io/cdap/plugin/mysql/MysqlDBRecord.java
@@ -93,4 +93,13 @@ public class MysqlDBRecord extends DBRecord {
 
     super.writeNonNullToDB(stmt, fieldSchema, fieldName, fieldIndex);
   }
+
+  @Override
+  protected void insertOperation(PreparedStatement stmt) throws SQLException {
+    for (int fieldIndex = 0; fieldIndex < columnTypes.size(); fieldIndex++) {
+      ColumnType columnType = columnTypes.get(fieldIndex);
+      Schema.Field field = record.getSchema().getField(columnType.getName(), true);
+      writeToDB(stmt, field, fieldIndex);
+    }
+  }
 }
--- a/cloudsql-mysql-plugin/src/main/java/io/cdap/plugin/cloudsql/mysql/CloudSQLMySQLSink.java
+++ b/cloudsql-mysql-plugin/src/main/java/io/cdap/plugin/cloudsql/mysql/CloudSQLMySQLSink.java
@@ -16,6 +16,7 @@
 
 package io.cdap.plugin.cloudsql.mysql;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableMap;
 import io.cdap.cdap.api.annotation.Description;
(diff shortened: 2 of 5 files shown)


ritwiksahani (approved):

itsankit-google (changes requested):
I know this will fix one part of the problem but now this turns into a backward incompatible change because let's say if MySQL by default convert all identifiers to lowercase and if a cx has an existing pipeline which has any column name in uppercase, it will start failing.

The right thing to do would be as described in JIRA to introduce a new config named `Field Mapping` which can map any CDAP field in the schema to a different name and will be passed as quoted.

psainics:
> I know this will fix one part of the problem but now this turns into a backward incompatible change because let's say if MySQL by default convert all identifiers to lowercase and if a cx has an existing pipeline which has any column name in uppercase, it will start failing.
> 
> The right thing to do would be as described in JIRA to introduce a new config named `Field Mapping` which can map any CDAP field in the schema to a different name and will be passed as quoted.

Mysql Col names are not case sensitive.

itsankit-google (changes requested):
Please do the same changes for `MySQL` sink as well.

itsankit-google:
> > I know this will fix one part of the problem but now this turns into a backward incompatible change because let's say if MySQL by default convert all identifiers to lowercase and if a cx has an existing pipeline which has any column name in uppercase, it will start failing.
> > The right thing to do would be as described in JIRA to introduce a new config named `Field Mapping` which can map any CDAP field in the schema to a different name and will be passed as quoted.
> 
> Mysql Col names are not case sensitivity.

Discussed offline, to address the larger problem in future.

mysql-plugin/src/main/java/io/cdap/plugin/mysql/MysqlDBRecord.java
```diff
@@ -93,4 +93,13 @@ protected void writeNonNullToDB(PreparedStatement stmt, Schema fieldSchema,
...
   }
+
+  @Override
+  protected void insertOperation(PreparedStatement stmt) throws SQLException {
+    for (int fieldIndex = 0; fieldIndex < columnTypes.size(); fieldIndex++) {
+      ColumnType columnType = columnTypes.get(fieldIndex);
+      Schema.Field field = record.getSchema().getField(columnType.getName(), true);
+      writeToDB(stmt, field, fieldIndex);
+    }
+  }
```
itsankit-google, line 104: why this change?

psainics: Ref: https://github.com/data-integrations/database-plugins/blob/acd47fa18a8bcfbc382c0c9ba92e7395e0e17f8b/oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSinkDBRecord.java#L44-L52


This is done to get the cols values from schema with ignoring case.

If cols in CDAP schema are in different case from the cols in the mysql DB, ignoring case is required else we will get null values.

itsankit-google (approved):

PR #515, opened 2024-10-10
Enable TLS Support (DTS - Specific)

Vipinofficial11:
***Add SSL/TLS support in Oracle Batch Source Classes***

This PR introduces a new field, ```useSSL```, to enable SSL (TLS) connections for Oracle by allowing the specification of protocol (tcp or tcps) in the connection string format.

****Key Changes****

1. New field

useSSL (Allows to specify whether SSL (TLS) should be used for  connections or not )
 - If useSSL is set to yes, the protocol in the connection string is set to tcps. Otherwise, it defaults to tcp.
 
 2. Connection String Format Updated:
 
 - For SID-based connections: ```jdbc:oracle:thin:@tcps or @tcp ://<host>:<port>/<SID>```
 - For Service Name-based connections: : ```jdbc:oracle:thin:tcps or @tcp ://<host>:<port>/<ServiceName>```
 
 
****Testing Scope****

- Validated TCP protocol connectivity:
Verified that the connection works as expected when specifying protocol as tcp explicitly in the connection string to ensure backward compatibility with existing behavior.

- Validated TCPS protocol connectivity:
Conducted internal tests with tcps to ensure secure connections (SSL/TLS) are correctly established and that the updated connection string format handles this protocol appropriately.

### DIFF
--- a/oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleConnectorConfig.java
+++ b/oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleConnectorConfig.java
@@ -43,12 +43,12 @@ public class OracleConnectorConfig extends AbstractDBSpecificConnectorConfig {
 
   public OracleConnectorConfig(String host, int port, String user, String password, String jdbcPluginName,
                                String connectionArguments, String connectionType, String database) {
-    this(host, port, user, password, jdbcPluginName, connectionArguments, connectionType, database, null);
+    this(host, port, user, password, jdbcPluginName, connectionArguments, connectionType, database, null, null);
   }
 
   public OracleConnectorConfig(String host, int port, String user, String password, String jdbcPluginName,
                                String connectionArguments, String connectionType, String database,
-                               String role) {
+                               String role, Boolean useSSL) {
 
     this.host = host;
     this.port = port;
@@ -59,11 +59,12 @@ public class OracleConnectorConfig extends AbstractDBSpecificConnectorConfig {
     this.connectionType = connectionType;
     this.database = database;
     this.role = role;
+    this.useSSL = useSSL;
   }
 
   @Override
   public String getConnectionString() {
-    return OracleConstants.getConnectionString(connectionType, host, getPort(), database);
+    return OracleConstants.getConnectionString(connectionType, host, getPort(), database, useSSL);
   }
 
   @Name(OracleConstants.CONNECTION_TYPE)
@@ -86,6 +87,11 @@ public class OracleConnectorConfig extends AbstractDBSpecificConnectorConfig {
   @Nullable
   private String transactionIsolationLevel;
 
+  @Name(OracleConstants.USE_SSL)
+  @Description("Turns on SSL encryption. Connection will fail if SSL is not available")
+  @Nullable
+  public Boolean useSSL;
+
   @Override
   protected int getDefaultPort() {
     return 1521;
@@ -101,6 +107,11 @@ public class OracleConnectorConfig extends AbstractDBSpecificConnectorConfig {
 
   public String getDatabase() {
     return database;
+  }
+
+  public Boolean getSSlMode() {
+    // return false if useSSL is null, otherwise return its value
+    return useSSL != null && useSSL;
   }
 
   @Override
--- a/oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleConstants.java
+++ b/oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleConstants.java
@@ -29,6 +29,10 @@ public final class OracleConstants {
   public static final String PLUGIN_NAME = "Oracle";
   public static final String ORACLE_CONNECTION_STRING_SID_FORMAT = "jdbc:oracle:thin:@%s:%s:%s";
   public static final String ORACLE_CONNECTION_STRING_SERVICE_NAME_FORMAT = "jdbc:oracle:thin:@//%s:%s/%s";
+  // Connection formats to accept protocol (e.g., jdbc:oracle:thin:@<protocol>://<host>:<port>/<SID>)
+  public static final String ORACLE_CONNECTION_STRING_SID_FORMAT_WITH_PROTOCOL = "jdbc:oracle:thin:@%s:%s:%s/%s";
+  public static final String ORACLE_CONNECTION_STRING_SERVICE_NAME_FORMAT_WITH_PROTOCOL =
+      "jdbc:oracle:thin:@%s://%s:%s/%s";
   public static final String ORACLE_CONNECTION_STRING_TNS_FORMAT = "jdbc:oracle:thin:@%s";
   public static final String DEFAULT_BATCH_VALUE = "defaultBatchValue";
   public static final String DEFAULT_ROW_PREFETCH = "defaultRowPrefetch";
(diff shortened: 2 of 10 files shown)


oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleConnectorConfig.java
```diff
@@ -86,6 +87,11 @@ public String getConnectionString() {
   @Nullable
   private String transactionIsolationLevel;
 
+  @Name(OracleConstants.USE_SSL)
+  @Description("Turns on SSL encryption. Connection will fail if SSL is not available")
+  @Nullable
+  public String useSSL;
```
itsankit-google: Why not `boolean`?

When we add something in plugin config, if we don't add a widget json for this. It will show up as a textbox in CDAP UI as well. 

If we don't want it to be visible in CDAP UI, please add a config in widgets json which is hidden.

Vipinofficial11: ``` Why not boolean? ```
There was no specific reason, change it to Boolean.


``` When we add something in plugin config, if we don't add a widget json for this. It will show up as a textbox in CDAP UI as well. If we don't want it to be visible in CDAP UI, please add a config in widgets json which is hidden.```

got it, added it in the config of Batch Source and Batch Sink as only these are using OracleConnectorConfig. [8d62fb9cc4a75433d844f5e5285e0dcd690ea88b]

File: oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleConnectorConfig.java [hunk 2]
```diff
@@ -103,6 +109,10 @@ public String getDatabase() {
     return database;
   }
 
+  public Boolean getSSlMode() {
+    return useSSL;
```
itsankit-google: For existing pipelines this property will be null, please handle that case.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleConstants.java
```diff
@@ -57,9 +60,48 @@ public static String getConnectionString(String connectionType,
...
+   * If SSL is enabled, the connection protocol will be "tcps" instead of "tcp".
+   *
+   * @param connectionType TNS/Service/SID
+   * @param host Host name of the oracle server
+   * @param port Port of the oracle server
+   * @param database Database to connect to
+   * @param useSSL Whether SSL/TLS is required(YES/NO)
+   * @return Connection String based on the given parameters and connection type.
+   */
+  public static String getConnectionString(String connectionType,
```
itsankit-google, line 58: I cannot see any reason of this new method, why can't we reuse the existing one?

Vipinofficial11: This was actually overloaded to reduce the impact of these changes on CDAP connectors, this method will only be used by Source and Sink plugin.

Oracle Action, Post Action and Oracle Connector will still use the old method to generate connection string. When we enable TLS for CDAP Connectors we can reuse the code and remove this method.

itsankit-google: If we are using this method here: https://github.com/data-integrations/database-plugins/pull/515/files#diff-bd85394fe9ff353467fd45fd10ad329b508e0fbbc70145fe34de0d54f483a99bR129

It means it does impact CDAP connectors so it does not really help by overriding the method.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleConnectorConfig.java
itsankit-google: There will also be a widget json file for `OracleConnector` which also need to be modified the same way.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleConnectorConfig.java
Vipinofficial11: Done, please have a look if this change would be enough to avoid failures in case of existing pipelines. 

[https://github.com/data-integrations/database-plugins/pull/515/commits/b68ec933ff27ff2a1ade2c578727113604803367#diff-afe65e631b8b5b1d360af43ae84dc844e2c55ee83e9179cdb8bd087da954ae06R114]

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleConstants.java
Vipinofficial11: got it, removed. [b68ec933ff27ff2a1ade2c578727113604803367]

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleConnectorConfig.java
Vipinofficial11: Updated [b68ec933ff27ff2a1ade2c578727113604803367]

itsankit-google:
Please check e2e test failures.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleConstants.java
```diff
@@ -27,39 +27,56 @@ private OracleConstants() {
   }
 
   public static final String PLUGIN_NAME = "Oracle";
-  public static final String ORACLE_CONNECTION_STRING_SID_FORMAT = "jdbc:oracle:thin:@%s:%s:%s";
-  public static final String ORACLE_CONNECTION_STRING_SERVICE_NAME_FORMAT = "jdbc:oracle:thin:@//%s:%s/%s";
+  // Updating connection strings to accept protocol (e.g., jdbc:oracle:thin:@<protocol>://<host>:<port>/<SID>)
+  public static final String ORACLE_CONNECTION_STRING_SID_FORMAT = "jdbc:oracle:thin:@%s:%s:%s/%s";
```
itsankit-google: We can add a constant `ORACLE_CONNECTION_STRING_SID_FORMAT_WITH_PROTOCOL` instead of modifying the older one and use the new one when `useSSL` is `true` otherwise use the old connection string. 

This would make the change backward compatible and ensure CDAP connectors are not affected.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleConstants.java
```diff
@@ -27,39 +27,56 @@ private OracleConstants() {
...
+  public static final String ORACLE_CONNECTION_STRING_SERVICE_NAME_FORMAT = "jdbc:oracle:thin:@%s://%s:%s/%s";
```
itsankit-google: similar change here `ORACLE_CONNECTION_STRING_SERVICE_NAME_FORMAT_WITH_PROTOCOL`

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleConstants.java
Vipinofficial11: got it, updated [b31e53c08a5d67793ec914efcfefe270c5e3a434]

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleConstants.java
Vipinofficial11: updated [b31e53c08a5d67793ec914efcfefe270c5e3a434]

Vipinofficial11:
> Please check e2e test failures.

Sure, I was looking at the reports but looks like there is some issue. Can you check if you are able to access these  [link](https://storage.googleapis.com/e2e-tests-cucumber-reports/database-plugins/refs/pull/515/merge/plugin/oracle-plugin/target/cucumber-reports/advanced-reports/cucumber-html-reports/overview-features.html)

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleConstants.java
```diff
@@ -29,37 +29,66 @@ private OracleConstants() {
...
+      case OracleConstants.TNS_CONNECTION_TYPE:
+        // TNS connection doesn't require protocol
+        return String.format(OracleConstants.ORACLE_CONNECTION_STRING_TNS_FORMAT, database);
+      case OracleConstants.SERVICE_CONNECTION_TYPE:
+        if (isSSLEnabled) {
+          return String.format(OracleConstants.ORACLE_CONNECTION_STRING_SERVICE_NAME_FORMAT_WITH_PROTOCOL,
+              connectionProtocol, host, port, database);
+        }
+        return String.format(OracleConstants.ORACLE_CONNECTION_STRING_SERVICE_NAME_FORMAT,
+            host, port, database);
```
itsankit-google: can we have two methods which takes `host, port, database & connectionProtocol` `getConnectionStringWithServic` & `getConnectionStringWithSID`?

Vipinofficial11: Added separate methods to get the different type of connection strings. [7dc9a6c7f9d409c8f3debe71085df9d10ecaeb0a]

itsankit-google:
> > Please check e2e test failures.
> 
> Sure, I was looking at the reports but looks like there is some issue. Can you check if you are able to access these [link](https://storage.googleapis.com/e2e-tests-cucumber-reports/database-plugins/refs/pull/515/merge/plugin/oracle-plugin/target/cucumber-reports/advanced-reports/cucumber-html-reports/overview-features.html)

If you see a step before, I don't even see these files getting uploaded: 
image

Vipinofficial11:
> > > Please check e2e test failures.
> > 
> > 
> > Sure, I was looking at the reports but looks like there is some issue. Can you check if you are able to access these [link](https://storage.googleapis.com/e2e-tests-cucumber-reports/database-plugins/refs/pull/515/merge/plugin/oracle-plugin/target/cucumber-reports/advanced-reports/cucumber-html-reports/overview-features.html)
> 
> If you see a step before, I don't even see these files getting uploaded: image

Yes, looks like some issue with reporting plugin for oracle. I might have seen this before, will try to fix it.

I was looking at older reports, even for the plugins where files were uploaded we are still not able to access the public bucket.  [see this](https://github.com/data-integrations/database-plugins/actions/runs/10919832650/job/30308367536)
Screenshot 2024-10-15 at 10 35 08 PM

itsankit-google:
> > > > Please check e2e test failures.
> > > 
> > > 
> > > Sure, I was looking at the reports but looks like there is some issue. Can you check if you are able to access these [link](https://storage.googleapis.com/e2e-tests-cucumber-reports/database-plugins/refs/pull/515/merge/plugin/oracle-plugin/target/cucumber-reports/advanced-reports/cucumber-html-reports/overview-features.html)
> > 
> > 
> > If you see a step before, I don't even see these files getting uploaded: image
> 
> Yes, looks like some issue with reporting plugin for oracle. I might have seen this before, will try to fix it.
> 
> I was looking at older reports, even for the plugins where files were uploaded we are still not able to access the public bucket. [see this](https://github.com/data-integrations/database-plugins/actions/runs/10919832650/job/30308367536) Screenshot 2024-10-15 at 10 35 08 PM

If the runs are 7 days old, the reports get deleted automatically from the storage bucket.

itsankit-google (approved):

PR #516, opened 2024-10-15
[PLUGIN-1812] Added fix for date datatype in oracle sink

vikasrathee-cs:
Added fix for date datatype in oracle sink. Added handling for logical type Date.
https://cdap.atlassian.net/browse/PLUGIN-1812


oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceDBRecord.java
```diff
@@ -160,7 +160,7 @@ protected void writeNonNullToDB(PreparedStatement stmt, Schema fieldSchema,
         String timestampString = Timestamp.valueOf(localDateTime).toString();
         Object timestampWithTimeZone = createOracleTimestamp(stmt.getConnection(), timestampString);
         stmt.setObject(sqlIndex, timestampWithTimeZone);
-      } else if (Schema.LogicalType.TIMESTAMP_MICROS.equals(fieldSchema.getLogicalType())) {
+      } else {
         // Deprecated: Handle the case when the Timestamp is mapped to CDAP Timestamp type
```
MrRahulSharma: Please update the comment.

vikasrathee-cs: done

MrRahulSharma (approved):
LGTM

itsankit-google (dismissed):
Please add a jira with the description.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceDBRecord.java
```diff
@@ -160,8 +160,8 @@ protected void writeNonNullToDB(PreparedStatement stmt, Schema fieldSchema,
         String timestampString = Timestamp.valueOf(localDateTime).toString();
         Object timestampWithTimeZone = createOracleTimestamp(stmt.getConnection(), timestampString);
         stmt.setObject(sqlIndex, timestampWithTimeZone);
-      } else if (Schema.LogicalType.TIMESTAMP_MICROS.equals(fieldSchema.getLogicalType())) {
-        // Deprecated: Handle the case when the Timestamp is mapped to CDAP Timestamp type
+      } else {
```
itsankit-google, line 163: How are we ensuring that this condition will be executed in case of `Timestamp` or `Date` type only?

itsankit-google:
The JIRA contains almost no description like when can the issue occur, how was it identified and what change caused it?
@vikasrathee-cs please add these details.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceDBRecord.java
vikasrathee-cs: In prepareRun and configurePipeline class, we have validateSchema calls which will validate it.

itsankit-google: It should be `DateTime` according to https://github.com/data-integrations/database-plugins/blob/f0347b9572c7c4a1032a8e64186119d5adf1f448/oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSource.java#L189

vikasrathee-cs: Source plugin will never return Date logical type as Date is not there in oracle, but BQ source can have Date type that user wants to put into Oracle timestamp field. Thats why need to handle in Sink plugin

itsankit-google: Ah I see, makes sense.

itsankit-google (approved):

PR #549, opened 2025-01-28
Update SQLException to ProgramFailureException

psainics:
Update SQLException to ProgramFailureException

Breaking due to changes in https://github.com/cdapio/hydrator-plugins/pull/1918


itsankit-google (commented):
Can we please check some more details like `ErrorReason`, `ErrorCategory`, etc?

The point is we also throw generic `ProgramFailureException` at multiple places so it would be good to check specifics here.

itsankit-google (approved):
LGTM for now, comment can be addressed in a follow up PR.

PR #584, opened 2025-04-23
[PLUGIN-1815] Commit/Rollback added in Committer.

vikasrathee-cs:
[PLUGIN-1815] In DB plugins we should commit/rollback DB transaction in commiter as currently we commit transaction even if task failed.

[PLUGIN-1815]: https://cdap.atlassian.net/browse/PLUGIN-1815?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/database-commons/src/main/java/io/cdap/plugin/db/sink/ETLDBOutputFormat.java
+++ b/database-commons/src/main/java/io/cdap/plugin/db/sink/ETLDBOutputFormat.java
@@ -25,6 +25,8 @@ import io.cdap.plugin.db.Operation;
 import io.cdap.plugin.db.TransactionIsolationLevel;
 import io.cdap.plugin.util.DBUtils;
 import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.mapreduce.JobContext;
+import org.apache.hadoop.mapreduce.OutputCommitter;
 import org.apache.hadoop.mapreduce.RecordWriter;
 import org.apache.hadoop.mapreduce.TaskAttemptContext;
 import org.apache.hadoop.mapreduce.lib.db.DBConfiguration;
@@ -43,6 +45,7 @@ import java.sql.SQLException;
 import java.sql.Statement;
 import java.util.Map;
 import java.util.Properties;
+import java.util.concurrent.ConcurrentHashMap;
 
 import static io.cdap.plugin.db.ConnectionConfigAccessor.OPERATION_NAME;
 import static io.cdap.plugin.db.ConnectionConfigAccessor.RELATION_TABLE_KEY;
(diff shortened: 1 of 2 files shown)


database-commons/src/main/java/io/cdap/plugin/db/sink/ETLDBOutputFormat.java
```diff
@@ -56,15 +59,98 @@
...
   private static final Character ESCAPE_CHAR = '"';
-
+  private static final String CONNECTION_MAP_ID_REGEX = "%s_%s";
   private static final Logger LOG = LoggerFactory.getLogger(ETLDBOutputFormat.class);
 
   private Configuration conf;
   private Driver driver;
   private JDBCDriverShim driverShim;
 
+  private static Map<String, Connection> connectionMap = new HashMap<>();
```
tivv: Please put it into static section, use uppercase name and add a comment on why is it static. Also use a concurrent map

database-commons/src/main/java/io/cdap/plugin/db/sink/ETLDBOutputFormat.java
```diff
@@ -56,15 +59,98 @@
 public class ETLDBOutputFormat<K extends DBWritable, V> extends DBOutputFormat<K, V> {
   // Batch size before submitting a batch to the SQL engine. If set to 0, no batches will be submitted until commit.
   public static final String COMMIT_BATCH_SIZE = "io.cdap.plugin.db.output.commit.batch.size";
+  public static final String STAGE_NAME = "io.cdap.plugin.db.output.stage_name";
   public static final int DEFAULT_COMMIT_BATCH_SIZE = 1000;
   private static final Character ESCAPE_CHAR = '"';
-
+  private static final String CONNECTION_MAP_ID_REGEX = "%s_%s";
```
tivv: It's not REGEX

database-commons/src/main/java/io/cdap/plugin/db/sink/ETLDBOutputFormat.java
```diff
@@ -56,15 +59,98 @@
...
+      @Override
+      public boolean needsTaskCommit(TaskAttemptContext taskContext) throws IOException {
+        return true;
+      }
+
+      @Override
+      public void commitTask(TaskAttemptContext taskContext) throws IOException {
+        conf = context.getConfiguration();
+        String stageName = conf.get(STAGE_NAME);
+        String connectionId = String.format(CONNECTION_MAP_ID_REGEX, context.getTaskAttemptID().toString(), stageName);
```
tivv: Please make a small utility method for this. Not a lot of code, but method would ensure we never go out of sync

tivv: Please make a small utility method for this. Not a lot of code, but method would ensure we never go out of sync

database-commons/src/main/java/io/cdap/plugin/db/sink/ETLDBOutputFormat.java
```diff
@@ -56,15 +59,98 @@
...
+          } finally {
+            try {
+              connection.close();
+              LOG.debug("Connection Closed after committing the task with taskAttemptId {}", connectionId);
+            } catch (SQLException ex) {
+              LOG.warn(StringUtils.stringifyException(ex));
+            }
+          }
+          try {
+            DriverManager.deregisterDriver(driverShim);
```
tivv: I believe driverShim should also go through the map, should not it?

vikasrathee-cs: I am moving it to close method only, it will not have any impact on commitTask or abortTask as we are not making any new connections.

database-commons/src/main/java/io/cdap/plugin/db/sink/ETLDBOutputFormat.java
```diff
@@ -98,23 +183,15 @@ public void close(TaskAttemptContext context) throws IOException {
...
             try {
               getStatement().close();
-              getConnection().close();
             } catch (SQLException ex) {
               throw new IOException(ex);
             }
           }
-
           try {
             DriverManager.deregisterDriver(driverShim);
```
tivv, line 196: After this call, driver may invalidate the connection, see java.sql.DriverAction#deregister:
"it is implementation specific as to whether the connections are closed or allowed to continue."

tivv: Discussed offline, DriverAction is not used in CDAP

tivv (approved):

PR #323, opened 2022-12-06
[PLUGIN-1119] Converting precision less decimal/numbers into String type.

MrRahulSharma:
Jira ticket: https://cdap.atlassian.net/browse/PLUGIN-1119

Changes to ensure that, Precision less Number types in Oracle Source gets converted into String types to avoid any precision loss.

Validations performed:
1. The getSchema shows String type against precision less number types.
2. Pipeline execution without getSchema converts the Number type to String type in the sink.
3. Pipeline execution with getSchema converts the Number type to String type in the sink.
4. Any other type selected manually for the field in the output schema of Oracle plugin, results into failure of type mismatch during pipeline execution.

PR #601, opened 2025-06-02
[PLUGIN-1893] Adding fields in Oracle source and connector which acts like a flag for Backward compatibility issues for Timestamp and number (precisionless) 

sahusanket:
[PLUGIN-1893](https://cdap.atlassian.net/browse/PLUGIN-1893)

### Problem / Context 

There were changes in Oracle DB plugin, where had modified the Schema handling. i.e. 
- Oracle `Number ( without precision)` was treated as `CDAP Decimal`, but in [PR_NUM](https://github.com/data-integrations/database-plugins/pull/323) it was changed to `String`
- Oracle `Date` was treated as `CDAP Timestamp micros`, but in [PR_timestamp](https://github.com/data-integrations/database-plugins/pull/386) it was changed to `Datetime` 

This caused issue for users who were already on older version of Oracle plugin, and while upgrading the schema type changed which acted like a blocker for them to upgrade. This was particularly the case where Schema was auto-detected ( i.e. use of macros ) 

This is targeted for Users with existing pipelines mostly. 

### Fix in the PR 

In this PR , we are giving a way to revert back to old behavior by setting a flag either in Connections or Plugin. 

By default : The flags are NOT ENABLED. So, creating a new pipeline will have new schema behavior. 

#### Via Connections 

- If the existing pipelines use connection, then this is easiest way 
- Export the connection json , edit `treatPrecisionlessNumAsDeci` / `treatAsOldTimestamp` to true or false. 
- upload again ( delete old one if needed) 
- Now all pipelines using this connection will handle schema as per old behavior. 

#### Via pipelines (only if connection is not used)

- If connections are not used , then user can edit or add field `treatPrecisionlessNumAsDeci` / `treatAsOldTimestamp` to true or false to the pipeline's json's Oracle plugin. 
- deploy the pipeline and the pipeline will handle schema as per old behavior ( as per flag).

 ### Various Scenarios and Testing Results
 
 All testing are done using a Pipeline ( ORACLE --> BQ ) 

 
#### Pipeline Run : 

`Case 1` : In pipeline : `Conn's flag is False/Default` + `Query -> Macro (auto-detect - schema)`
- BQ Table's Schema : NEW BEHAVIOR (num as String , date as Datetime) 

`Case 2` : In pipeline : `Conn's flag is True` + `Query -> Macro (auto-detect - schema)`
- BQ Table's Schema :  OLD BEHAVIOR (num as Deci --> BIGNUMERIC , date as Timestamp)

`Case 3` : In pipeline : `No Conn` + `Plugin's flag is False` + `Query -> Macro (auto-detect - schema)`
- BQ Table's Schema : NEW BEHAVIOR (num as String , date as Datetime) 

`Case 4` : In pipeline : `No Conn` + `Plugin's flag is True` + `Query -> Macro (auto-detect - schema)`
- BQ Table's Schema : OLD BEHAVIOR (num as Deci --> BIGNUMERIC , date as Timestamp)

---
#### Studio : 

`Case 1` : In Studio : `Conn's flag is False/Default` + `Query provided`
- Get Schema :NEW BEHAVIOR (num as String , date as Datetime)

`Case 2` : In Studio : `Conn's flag is True` + `Query provided`
- Get Schema : OLD BEHAVIOR (num as Deci , date as Timestamp)

`Case 3` : In Studio : `No Conn` + `Query provided`
- Get Schema : NEW BEHAVIOR (num as String , date as Datetime)

`Case 4` : In Studio : `No Conn` + `Query provided` + `Edit oracle plugin's flag to true `
- Get Schema : OLD BEHAVIOR (num as Deci , date as Timestamp)

---
#### Wrangler / Connection : 

`Case 1` : `Conn's flag is False/Default` 
-  Schema : NEW BEHAVIOR (num as String , date as Datetime)

`Case 2` : `Conn's flag is True` 
-  Schema : OLD BEHAVIOR (num as Deci , date as Timestamp)

---
#### Upgrade / Import / export case 

- on 6.11 : Create a pipeline with an  connection + existing plugin + run ( this should give new bahavior )
- Upload new plugin + export connection json + delete the conn and import connection using TRUE with same name. 
- upgrade pipeline as well. Run , this should give OLD schema. 
- Result : The initial pipeline : NEW BEHAVIOR (num as String , date as Datetime) 
- Result : The upgraded conn + upgraded pipeline : OLD BEHAVIOR (num as Deci , date as Timestamp)

[PLUGIN-1893]: https://cdap.atlassian.net/browse/PLUGIN-1893?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleConnectorConfig.java
+++ b/oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleConnectorConfig.java
@@ -22,8 +22,6 @@ import io.cdap.cdap.api.annotation.Name;
 import io.cdap.plugin.db.TransactionIsolationLevel;
 import io.cdap.plugin.db.connector.AbstractDBSpecificConnectorConfig;
 
-import java.util.HashMap;
-import java.util.Map;
 import java.util.Properties;
 import javax.annotation.Nullable;
 
@@ -43,12 +41,14 @@ public class OracleConnectorConfig extends AbstractDBSpecificConnectorConfig {
 
   public OracleConnectorConfig(String host, int port, String user, String password, String jdbcPluginName,
                                String connectionArguments, String connectionType, String database) {
-    this(host, port, user, password, jdbcPluginName, connectionArguments, connectionType, database, null, null);
+    this(host, port, user, password, jdbcPluginName, connectionArguments, connectionType, database, null, null, null,
+         null);
   }
 
   public OracleConnectorConfig(String host, int port, String user, String password, String jdbcPluginName,
                                String connectionArguments, String connectionType, String database,
-                               String role, Boolean useSSL) {
+                               String role, Boolean useSSL, @Nullable Boolean treatAsOldTimestamp,
+                               @Nullable Boolean treatPrecisionlessNumAsDeci) {
 
     this.host = host;
     this.port = port;
@@ -60,6 +60,8 @@ public class OracleConnectorConfig extends AbstractDBSpecificConnectorConfig {
     this.database = database;
     this.role = role;
     this.useSSL = useSSL;
+    this.treatAsOldTimestamp = treatAsOldTimestamp;
+    this.treatPrecisionlessNumAsDeci = treatPrecisionlessNumAsDeci;
   }
 
   @Override
@@ -86,6 +88,16 @@ public class OracleConnectorConfig extends AbstractDBSpecificConnectorConfig {
   @Nullable
   public Boolean useSSL;
 
+  @Name(OracleConstants.TREAT_AS_OLD_TIMESTAMP)
+  @Description("A hidden field to handle timestamp as CDAP's timestamp micros or string as per old behavior.")
+  @Nullable
+  public Boolean treatAsOldTimestamp;
+
+  @Name(OracleConstants.TREAT_PRECISIONLESSNUM_AS_DECI)
+  @Description("A hidden field to handle precision less number as CDAP's decimal per old behavior.")
+  @Nullable
+  public Boolean treatPrecisionlessNumAsDeci;
+
   @Override
   protected int getDefaultPort() {
     return 1521;
@@ -106,6 +118,14 @@ public class OracleConnectorConfig extends AbstractDBSpecificConnectorConfig {
   public Boolean getSSlMode() {
     // return false if useSSL is null, otherwise return its value
     return useSSL != null && useSSL;
+  }
+
+  public Boolean getTreatAsOldTimestamp() {
+    return Boolean.TRUE.equals(treatAsOldTimestamp);
+  }
+
+  public Boolean getTreatPrecisionlessNumAsDeci() {
+    return Boolean.TRUE.equals(treatPrecisionlessNumAsDeci);
   }
 
   @Override
--- a/oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSource.java
+++ b/oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSource.java
@@ -63,7 +63,12 @@ public class OracleSource extends AbstractDBSource<OracleSource.OracleSourceConf
 
   @Override
   protected SchemaReader getSchemaReader() {
-    return new OracleSourceSchemaReader();
+    // PLUGIN-1893 : Based on field/properties from Oracle source and Oracle connection we will pass the flag to control
+    // handle schema to make it backward compatible.
+    boolean treatAsOldTimestamp = oracleSourceConfig.getConnection().getTreatAsOldTimestamp();
+    boolean treatPrecisionlessNumAsDeci = oracleSourceConfig.getConnection().getTreatPrecisionlessNumAsDeci();
+
+    return new OracleSourceSchemaReader(null, treatAsOldTimestamp, treatPrecisionlessNumAsDeci);
   }
 
   @Override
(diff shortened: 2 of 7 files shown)


oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSource.java
```diff
@@ -63,7 +63,24 @@ protected String createConnectionString() {
 
   @Override
   protected SchemaReader getSchemaReader() {
-    return new OracleSourceSchemaReader();
+    //PLUGIN-1893 : Based on field/properties from Oracle source and Oracle connection we will pass the flag to control
+    //handle schema to make it backward compatible.
```
itsankit-google: nit: add space after `//`

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSource.java
```diff
@@ -123,20 +140,35 @@ public static class OracleSourceConfig extends AbstractDBSpecificSourceConfig {
     @Nullable
     private Integer defaultRowPrefetch;
 
+    @Name(OracleConstants.TREAT_AS_OLD_TIMESTAMP)
+    @Description("A hidden field to handle timestamp as CDAP's timestamp micros or string as per old behavior.")
+    @Nullable
+    private String treatAsOldTimestamp;
```
itsankit-google: use `Boolean`

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSource.java
```diff
@@ -123,20 +140,35 @@ public static class OracleSourceConfig extends AbstractDBSpecificSourceConfig {
...
+
+    @Name(OracleConstants.TREAT_PRECISIONLESSNUM_AS_DECI)
+    @Description("A hidden field to handle precision less number as CDAP's decimal per old behavior.")
+    @Nullable
+    public String treatPrecisionlessNumAsDeci;
```
itsankit-google: use `Boolean`

oracle-plugin/widgets/Oracle-batchsource.json
```diff
@@ -246,6 +246,24 @@
...
+          }
+        },
+        {
+          "widget-type": "hidden",
+          "label": "Treat precision less number as Decimal(old behavior)",
+          "name": "treatPrecisionlessNumAsDeci",
+          "description": "If edited, then this will take precedence over other properties and the edited value shall be true or false only",
+          "widget-attributes": {
+            "default": "default-no"
+          }
```
itsankit-google: I think we can still use `toggle` widget properties, radio-group properties are used in `useSSL` even though it is also of type `hidden`, can we test?

This way we can read it as `Boolean` instead of string matching which is brittle.

sahusanket: Yes, this was intentional. 

My idea was how to choose when between `Connection's Flag = TRUE` and `Plugin's there is no change default, meaning FALSE` 

In that case, we should go with `TRUE` . Hence I made it a string with default ( not true or false) so we can detect this is default and not changed by user. 
If we make it boolean, how do we know if the user has made it false or was it false by default ?

itsankit-google (changes requested):
on taking a deeper look, I think we do not have fields in `Connector` in `PluginConfig`, this is to maintain single copy of those configs.

So it will always come from the connection, please check other fields present in `ConnectorConfig`.

oracle-plugin/widgets/Oracle-batchsource.json
sahusanket: This was resolved after having only 1 flag at connection level.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleConnectorConfig.java
```diff
@@ -43,12 +41,14 @@ public OracleConnectorConfig(String host, int port, String user, String password
...
-    this(host, port, user, password, jdbcPluginName, connectionArguments, connectionType, database, null, null);
+    this(host, port, user, password, jdbcPluginName, connectionArguments, connectionType, database, null, null, null,
+         null);
   }
 
   public OracleConnectorConfig(String host, int port, String user, String password, String jdbcPluginName,
                                String connectionArguments, String connectionType, String database,
-                               String role, Boolean useSSL) {
+                               String role, Boolean useSSL, Boolean treatAsOldTimestamp,
+                               Boolean treatPrecisionlessNumAsDeci) {
```
itsankit-google: nit: add `Nullable` annotation to all the params which can be null.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceSchemaReader.java
```diff
@@ -65,14 +65,17 @@ public class OracleSourceSchemaReader extends CommonSchemaReader {
...
+  private final Boolean isPrecisionlessNumAsDecimal;
 
   public OracleSourceSchemaReader() {
-    this(null);
+    this(null, false, false);
   }
-
-  public OracleSourceSchemaReader(String sessionID) {
-    super();
+  public OracleSourceSchemaReader(String sessionID, boolean isTimestampOldBehavior,
```
itsankit-google: nit: add `Nullable` annotation to all the params which can be null.

itsankit-google (approved):
minor nits, otherwise LGTM.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceSchemaReader.java
sahusanket: This cannot be null. 
Null checks are handled prior to this call.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleConnectorConfig.java
sahusanket: Done

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceSchemaReader.java
itsankit-google: sessionID is passed as null here: https://github.com/data-integrations/database-plugins/pull/601/files#diff-38a8f3375f64fe92b6ef315fd46e390956725ae5268a7f2b129aa247dfe9094aR71

sahusanket: I considered only the new changes. 
Added nullable for sessionId.

PR #619, opened 2025-09-05
PLUGIN-1925: Introduce Custom BigDecimal Splitter

sgarg-CS:
[PLUGIN-1925](https://cdap.atlassian.net/browse/PLUGIN-1925)

**Issue**

The current split point calculation logic in the Oracle Plugin generates excessively large split point lists (up to ~10M), leading to high memory consumption and OOM errors. This occurs when the difference between minVal and maxVal is smaller than the number of splits, causing division to result in `0` and fallback to `MIN_SPLIT_SIZE`, which explodes the number of splits

**Root Cause**
- No explicit scale set in `BigDecimal.divide()`.
- Division like 1 / 4 returned 0 instead of 0.25 due to lack of scale. Default: 0
- Fallback to `MIN_SPLIT_SIZE` inflated splits drastically.

**Proposed Fix**
- Determine scale from column metadata (NUMERIC/DECIMAL).
- Apply buffer of +5 digits to preserve accuracy.
- Override `BigDecimalSplitter.tryDivide()` with custom implementation.
- Overriding the DataDrivenETLDBInputFormat.getSplitter method

[PLUGIN-1925]: https://cdap.atlassian.net/browse/PLUGIN-1925

### DIFF
--- a/database-commons/src/main/java/io/cdap/plugin/db/source/DataDrivenETLDBInputFormat.java
+++ b/database-commons/src/main/java/io/cdap/plugin/db/source/DataDrivenETLDBInputFormat.java
@@ -28,6 +28,7 @@ import org.apache.hadoop.mapreduce.RecordReader;
 import org.apache.hadoop.mapreduce.TaskAttemptContext;
 import org.apache.hadoop.mapreduce.lib.db.DBConfiguration;
 import org.apache.hadoop.mapreduce.lib.db.DBInputFormat;
+import org.apache.hadoop.mapreduce.lib.db.DBSplitter;
 import org.apache.hadoop.mapreduce.lib.db.DBWritable;
 import org.apache.hadoop.mapreduce.lib.db.DataDrivenDBInputFormat;
 import org.slf4j.Logger;
@@ -39,6 +40,7 @@ import java.sql.Driver;
 import java.sql.DriverManager;
 import java.sql.SQLException;
 import java.sql.Statement;
+import java.sql.Types;
 import java.util.Properties;
 
 /**
@@ -126,6 +128,15 @@ public class DataDrivenETLDBInputFormat extends DataDrivenDBInputFormat {
   // this is added for compatibility, more information at (HYDRATOR-791)
   public Connection createConnection() {
     return getConnection();
+  }
+
+  @Override
+  protected DBSplitter getSplitter(int sqlDataType) {
+    // Use SafeBigDecimalSplitter for columns having high precision decimal or numeric columns
+    if (sqlDataType == Types.NUMERIC || sqlDataType == Types.DECIMAL) {
+      return new SafeBigDecimalSplitter();
+    }
+    return super.getSplitter(sqlDataType);
   }
 
   @Override
--- a/database-commons/src/main/java/io/cdap/plugin/db/source/SafeBigDecimalSplitter.java
+++ b/database-commons/src/main/java/io/cdap/plugin/db/source/SafeBigDecimalSplitter.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright © 2025 Cask Data, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package io.cdap.plugin.db.source;
+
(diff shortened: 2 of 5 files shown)


anup-cloudsufi:
LGTM

database-commons/src/main/java/io/cdap/plugin/db/source/CustomBigDecimalSplitter.java
```diff
@@ -0,0 +1,44 @@
...
+   * @param denominator the divisor (BigDecimal)
+   * @return quotient with derived scale
+   * @throws ArithmeticException if denominator is zero
+   */
+  @Override
+  protected BigDecimal tryDivide(BigDecimal numerator, BigDecimal denominator) {
+    // Derive scale from numerator/denominator + buffer
+    int effectiveScale = Math.max(numerator.scale(), denominator.scale()) + SCALE_BUFFER;
+    return numerator.divide(denominator, effectiveScale, RoundingMode.HALF_UP);
+  }
```
MrRahulSharma: Please add comments explaining this particular logic.

database-commons/src/main/java/io/cdap/plugin/db/source/CustomBigDecimalSplitter.java
```diff
@@ -0,0 +1,44 @@
...
+import java.math.BigDecimal;
+import java.math.RoundingMode;
+
+/**
+ * Custom implementation of {@link BigDecimalSplitter} to ensures safe and precise division of BigDecimal values while
+ * calculating split points for NUMERIC and DECIMAL types.
+ */
+public class CustomBigDecimalSplitter extends BigDecimalSplitter {
+
+  public static final int SCALE_BUFFER = 5;
```
MrRahulSharma: How is the SCALE_BUFFER determined(add a comment)? Would this value work for all scenarios ? 
If not what is the guiding principle to determine the correct scale buffer to have? If it needs to be determined externally ?

database-commons/src/main/java/io/cdap/plugin/db/source/CustomBigDecimalSplitter.java
```diff
@@ -0,0 +1,44 @@
...
+
+import org.apache.hadoop.mapreduce.lib.db.BigDecimalSplitter;
+import java.math.BigDecimal;
+import java.math.RoundingMode;
+
+/**
+ * Custom implementation of {@link BigDecimalSplitter} to ensures safe and precise division of BigDecimal values while
+ * calculating split points for NUMERIC and DECIMAL types.
+ */
+public class CustomBigDecimalSplitter extends BigDecimalSplitter {
```
MrRahulSharma: Class name like CustomBigDecimalSplitter does not give much details about the intent of the class. Can you work on  a better name for this class based on what it really provide ?

database-commons/src/main/java/io/cdap/plugin/db/source/DataDrivenETLDBInputFormat.java
```diff
@@ -128,6 +130,14 @@ public Connection createConnection() {
     return getConnection();
   }
 
+  @Override
+  protected DBSplitter getSplitter(int sqlDataType) {
+    if (sqlDataType == Types.NUMERIC || sqlDataType == Types.DECIMAL) {
+      return new CustomBigDecimalSplitter();
+    }
```
MrRahulSharma: Add comments around what column types are required to use the new splitter.

database-commons/src/main/java/io/cdap/plugin/db/source/CustomBigDecimalSplitter.java
sgarg-CS: Renamed to `SafeBigDecimalSplitter` 9f26405b226d0d3b8598c292124858486d326c3a

database-commons/src/main/java/io/cdap/plugin/db/source/CustomBigDecimalSplitter.java
sgarg-CS: Added the comment 9f26405b226d0d3b8598c292124858486d326c3a.  `5` gives a good enough buffer to maintain the accuracy during the division for all practical scenarios. This ensures correctness even for databases that support very large precision values. The same buffer was applied while fixing the OOM issue for the PostgreSQL Connector as well.

database-commons/src/main/java/io/cdap/plugin/db/source/CustomBigDecimalSplitter.java
sgarg-CS: Added the comments explaining the need for this logic to the class 9f26405b226d0d3b8598c292124858486d326c3a

database-commons/src/main/java/io/cdap/plugin/db/source/DataDrivenETLDBInputFormat.java
sgarg-CS: Added 9f26405b226d0d3b8598c292124858486d326c3a

MrRahulSharma (approved):
LGTM

PR #623, opened 2025-09-17
Prevent auto security injection on TCPS URLs

Vipinofficial11:
**Issue**
The Oracle Thin driver automatically enforces an unwanted security default when connecting via the simple TCPS URL format (`jdbc:oracle:thin:@tcps://...`). This results in an automatic security injection of `SSL_SERVER_DN_MATCH=TRUE` into the connection string, which overrides custom security settings (`SSL_SERVER_DN_MATCH=FALSE`), causing TLS connections to fail due to a forced hostname mismatch even when hostname matching is disabled.

**Root Cause**
The Oracle driver's easy connect URL parser attempts to convert the simple `tcps://` format into the verbose TNS descriptor format (`(DESCRIPTION=...)`). During this internal transformation, it prioritizes baking in its own secure defaults, corrupting our explicit certificate verification configuration.

**Proposed Fix**
We will bypass the automatic URL parser by ensuring that whenever the TCPS protocol is selected (in either service or sid mode), the application generates the full verbose TNS descriptor string itself.

This change prevents the driver from performing the unwanted transformation, allowing our per-connection properties (including `SSL_SERVER_DN_MATCH=FALSE`) to be correctly honored for secure Private CA connections.


MrRahulSharma (approved):
LGTM

PR #637, opened 2025-12-16
PLUGIN-1939: Adding nullable split support in DB plugins.

anup-cloudsufi:
**Background**: partitioned reads on an Oracle table would skip any records where the partitioning column (split-by column) contained a `NULL` value. This caused data loss and incorrect record counts because the underlying logic only generated splits for non-null value ranges. This is applicable for other DB plugins.

Changes involved:
- overriding `DataDrivenETLDBInputFormat.getSplits` to check the generated splits and if no split covering all records or a specific `IS NULL` split is found, adds a dedicated `InputSplit` using the condition `[splitByColumn] IS NULL`.
- relevant unit tests


database-commons/src/main/java/io/cdap/plugin/db/source/DataDrivenETLDBInputFormat.java
```diff
@@ -64,6 +69,49 @@ public static void setInput(Configuration conf,
...
+        .map(DataDrivenDBInputFormat.DataDrivenDBInputSplit.class::cast)
+        .map(DataDrivenDBInputFormat.DataDrivenDBInputSplit::getLowerClause)
+        .noneMatch(lowerClause ->
+            !Strings.isNullOrEmpty(lowerClause)
+                && (lowerClause.contains("IS NULL") || lowerClause.contains("1=1")));
+
+    if (shouldAddNullSplit) {
+      // Create a dedicated split to handle potential NULL values in the split-by column.
+      String splitByColumn = getDBConf().getInputOrderBy();
+      String clause = String.format("%s IS NULL", splitByColumn);
```
MrRahulSharma, line 111: Please add quotes around the column name to ensure casing is preserved in query.

database-commons/src/main/java/io/cdap/plugin/db/source/DataDrivenETLDBInputFormat.java
```diff
@@ -64,6 +69,49 @@ public static void setInput(Configuration conf,
...
+  List<InputSplit> getBaseSplits(JobContext job) throws IOException {
+    return super.getSplits(job);
+  }
+
+  @Override
+  public List<InputSplit> getSplits(JobContext job) throws IOException {
+    List<InputSplit> splits = getBaseSplits(job);
+
+    // Check if a split for NULL values or an all-encompassing split ("1=1") already exists.
+    boolean shouldAddNullSplit = splits.stream()
```
MrRahulSharma, line 100: Missing null/empty checks for splits list.

database-commons/src/main/java/io/cdap/plugin/db/source/DataDrivenETLDBInputFormat.java
```diff
@@ -64,6 +69,49 @@ public static void setInput(Configuration conf,
...
+        .map(DataDrivenDBInputFormat.DataDrivenDBInputSplit.class::cast)
+        .map(DataDrivenDBInputFormat.DataDrivenDBInputSplit::getLowerClause)
+        .noneMatch(lowerClause ->
+            !Strings.isNullOrEmpty(lowerClause)
+                && (lowerClause.contains("IS NULL") || lowerClause.contains("1=1")));
+
+    if (shouldAddNullSplit) {
+      // Create a dedicated split to handle potential NULL values in the split-by column.
+      String splitByColumn = getDBConf().getInputOrderBy();
```
MrRahulSharma, line 109: Add a log statement to log that the Null value partition was added.

database-commons/src/main/java/io/cdap/plugin/db/source/DataDrivenETLDBInputFormat.java
anup-cloudsufi: Added the check.

database-commons/src/main/java/io/cdap/plugin/db/source/DataDrivenETLDBInputFormat.java
anup-cloudsufi: Done.

database-commons/src/main/java/io/cdap/plugin/db/source/DataDrivenETLDBInputFormat.java
anup-cloudsufi: For quoting, I would raise a separate PR, doing the feasibility analysis for the same.

MrRahulSharma (approved):

PR #645, opened 2026-01-30
[PLUGIN-1949] Add Oracle UPSERT

psainics:
## Add Oracle UPSERT

Jira : [PLUGIN-1945](https://cdap.atlassian.net/browse/PLUGIN-1949)

This pull request introduces support for upsert operations in the Oracle batch sink plugin. 

* Added a new class `OracleETLDBOutputFormat` that extends `ETLDBOutputFormat` and implements the `constructUpsertQuery` method, generating an Oracle-compatible MERGE-based upsert SQL statement.
* Updated `OracleSink` to use the new `OracleETLDBOutputFormat` by adding an output context with `SinkOutputFormatProvider`.

## UI Update 
image

[PLUGIN-1945]: https://cdap.atlassian.net/browse/PLUGIN-1945?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/oracle-plugin/src/e2e-test/features/sink/OracleRunTime.feature
+++ b/oracle-plugin/src/e2e-test/features/sink/OracleRunTime.feature
@@ -272,3 +272,175 @@ Feature: Oracle - Verify data transfer from BigQuery source to Oracle sink
     Then Verify the pipeline status is "Succeeded"
     Then Validate records transferred to target table with record counts of BigQuery table
     Then Validate the values of records transferred to target Oracle table is equal to the values from source BigQuery table
+
+  @ORACLE_SOURCE_DATATYPES_TEST @ORACLE_TARGET_DATATYPES_TEST @Oracle_Required
+  Scenario Outline: To verify pipeline preview failed with invalid table key
+    Given Open Datafusion Project to configure pipeline
+    When Expand Plugin group in the LHS plugins list: "Source"
+    When Select plugin: "Oracle" from the plugins list as: "Source"
+    When Expand Plugin group in the LHS plugins list: "Sink"
+    When Select plugin: "Oracle" from the plugins list as: "Sink"
+    Then Connect plugins: "Oracle" and "Oracle2" to establish connection
+    Then Navigate to the properties page of plugin: "Oracle"
+    Then Select dropdown plugin property: "select-jdbcPluginName" with option value: "driverName"
+    Then Replace input plugin property: "host" with value: "host" for Credentials and Authorization related fields
+    Then Replace input plugin property: "port" with value: "port" for Credentials and Authorization related fields
+    Then Replace input plugin property: "user" with value: "username" for Credentials and Authorization related fields
+    Then Replace input plugin property: "password" with value: "password" for Credentials and Authorization related fields
+    Then Select radio button plugin property: "connectionType" with value: "service"
+    Then Select radio button plugin property: "role" with value: "normal"
+    Then Enter input plugin property: "referenceName" with value: "sourceRef"
+    Then Replace input plugin property: "database" with value: "databaseName"
+    Then Enter textarea plugin property: "importQuery" with value: "selectQuery"
+    Then Click on the Get Schema button
+    Then Verify the Output Schema matches the Expected Schema: "outputDatatypesSchema"
+    Then Validate "Oracle" plugin properties
+    Then Close the Plugin Properties page
+    Then Navigate to the properties page of plugin: "Oracle2"
+    Then Select dropdown plugin property: "select-jdbcPluginName" with option value: "driverName"
+    Then Replace input plugin property: "host" with value: "host" for Credentials and Authorization related fields
+    Then Replace input plugin property: "port" with value: "port" for Credentials and Authorization related fields
+    Then Replace input plugin property: "user" with value: "username" for Credentials and Authorization related fields
+    Then Replace input plugin property: "password" with value: "password" for Credentials and Authorization related fields
+    Then Select radio button plugin property: "connectionType" with value: "service"
+    Then Select radio button plugin property: "role" with value: "normal"
+    Then Enter input plugin property: "referenceName" with value: "sourceRef"
+    Then Replace input plugin property: "database" with value: "databaseName"
+    Then Replace input plugin property: "tableName" with value: "targetTable"
+    Then Replace input plugin property: "dbSchemaName" with value: "schema"
+    Then Replace input plugin property: "user" with value: "username" for Credentials and Authorization related fields
+    Then Replace input plugin property: "password" with value: "password" for Credentials and Authorization related fields
+    Then Enter input plugin property: "referenceName" with value: "targetRef"
+    Then Select radio button plugin property: "connectionType" with value: "service"
+    Then Select radio button plugin property: "role" with value: "normal"
+    Then Select radio button plugin property: "operationName" with value: "<options>"
+    Then Click on the Add Button of the property: "relationTableKey" with value:
+      | invalidOracleTableKey |
+    Then Close the Plugin Properties page
+    Then Save the pipeline
+    Then Preview and run the pipeline
+    Then Verify the preview of pipeline is "Failed"
+    Examples:
+      | options |
+      | upsert  |
+      | update  |
+
+  @ORACLE_SOURCE_TEST @ORACLE_UPDATE_TABLE @Oracle_Required
+  Scenario Outline: To verify data is getting transferred from Oracle to Oracle successfully using update operation with table key
+    Given Open Datafusion Project to configure pipeline
+    When Expand Plugin group in the LHS plugins list: "Source"
+    When Select plugin: "Oracle" from the plugins list as: "Source"
+    When Expand Plugin group in the LHS plugins list: "Sink"
+    When Select plugin: "Oracle" from the plugins list as: "Sink"
+    Then Connect plugins: "Oracle" and "Oracle2" to establish connection
+    Then Navigate to the properties page of plugin: "Oracle"
+    Then Select dropdown plugin property: "select-jdbcPluginName" with option value: "driverName"
+    Then Replace input plugin property: "host" with value: "host" for Credentials and Authorization related fields
+    Then Replace input plugin property: "port" with value: "port" for Credentials and Authorization related fields
+    Then Replace input plugin property: "user" with value: "username" for Credentials and Authorization related fields
+    Then Replace input plugin property: "password" with value: "password" for Credentials and Authorization related fields
+    Then Select radio button plugin property: "connectionType" with value: "service"
+    Then Select radio button plugin property: "role" with value: "normal"
+    Then Enter input plugin property: "referenceName" with value: "sourceRef"
+    Then Replace input plugin property: "database" with value: "databaseName"
+    Then Enter textarea plugin property: "importQuery" with value: "selectQuery"
+    Then Click on the Get Schema button
+    Then Verify the Output Schema matches the Expected Schema: "outputSchema"
+    Then Validate "Oracle" plugin properties
+    Then Close the Plugin Properties page
+    Then Navigate to the properties page of plugin: "Oracle2"
+    Then Select dropdown plugin property: "select-jdbcPluginName" with option value: "driverName"
+    Then Replace input plugin property: "host" with value: "host" for Credentials and Authorization related fields
+    Then Replace input plugin property: "port" with value: "port" for Credentials and Authorization related fields
+    Then Replace input plugin property: "user" with value: "username" for Credentials and Authorization related fields
+    Then Replace input plugin property: "password" with value: "password" for Credentials and Authorization related fields
+    Then Select radio button plugin property: "connectionType" with value: "service"
+    Then Select radio button plugin property: "role" with value: "normal"
+    Then Enter input plugin property: "referenceName" with value: "sourceRef"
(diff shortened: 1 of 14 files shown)


prince-cs (commented):
LGTM

prince-cs (approved):

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleETLDBOutputFormat.java
```diff
@@ -0,0 +1,122 @@
...
+   * WHEN NOT MATCHED THEN INSERT (id, name, age) VALUES (source.id, source.name, source.age)
+   * @param table - Name of the table
+   * @param fieldNames - All the columns of the table
+   * @param listKeys - The columns used as keys for matching
+   * @return Upsert query in the form of string
+   */
+  @Override
+  public String constructUpsertQuery(String table, String[] fieldNames, String[] listKeys) {
+    if (listKeys == null) {
+      throw new IllegalArgumentException("Column names to be updated should not be null");
```
itsankit-google: what is the name for this config in UI?

`Column names` does not help understand which config are we exactly talking about.

Also, can we move these to sink config validations when `UPSERT` is selected?

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleETLDBOutputFormat.java
```diff
@@ -0,0 +1,122 @@
...
+    } else if (fieldNames == null) {
+      throw new IllegalArgumentException("Field names should not be null");
```
itsankit-google: similar comment here.

Better to move these to sink config validations to fail early.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceDBRecord.java
```diff
@@ -116,8 +116,15 @@ protected void handleField(ResultSet resultSet, StructuredRecord.Builder recordB
   @Override
   protected void writeNonNullToDB(PreparedStatement stmt, Schema fieldSchema,
                                   String fieldName, int fieldIndex) throws SQLException {
-    int sqlType = columnTypes.get(fieldIndex).getType();
```
itsankit-google, line 119: In which case, is it possible that `fieldIndex` is greater than `columnTypes.size()-1`?

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceDBRecord.java
```diff
@@ -116,8 +116,15 @@ protected void handleField(ResultSet resultSet, StructuredRecord.Builder recordB
...
     int sqlIndex = fieldIndex + 1;
+    int sqlType = Types.OTHER;
+    // avoid OOB exception in case of mismatch between columnTypes and record schema fields
+    for (ColumnType columnType : columnTypes) {
+      if (columnType.getName().equals(fieldName)) {
```
itsankit-google: should we be using `equalsIgnoreCase` like https://github.com/data-integrations/database-plugins/blob/ec0383cd57b05b239d745471e19041312e2ec1d9/oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSinkDBRecord.java#L49?

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceDBRecord.java
```diff
@@ -116,8 +116,15 @@ protected void handleField(ResultSet resultSet, StructuredRecord.Builder recordB
   @Override
   protected void writeNonNullToDB(PreparedStatement stmt, Schema fieldSchema,
                                   String fieldName, int fieldIndex) throws SQLException {
-    int sqlType = columnTypes.get(fieldIndex).getType();
     int sqlIndex = fieldIndex + 1;
+    int sqlType = Types.OTHER;
```
itsankit-google: should we be throwing an error if none of the `fieldName` matches in the `columnType.getName()` list?

Why are we assuming `Types.OTHER` when there is no match?

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceDBRecord.java
psainics: no, oracle has case sensitive col names

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceDBRecord.java
psainics: The logic in common class adds col (that are to be updated) and then col (which we are going to match), this can lead to query having duplicate reference and the field index be more than the total col size.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceDBRecord.java
psainics: It will fail for `Types.OTHER`, but throwing an error sound like better option.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleETLDBOutputFormat.java
psainics: This is input schema

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleETLDBOutputFormat.java
psainics: Yes, we also have this validation in Abstract DB sink 

https://github.com/data-integrations/database-plugins/blob/ec0383cd57b05b239d745471e19041312e2ec1d9/database-commons/src/main/java/io/cdap/plugin/db/sink/AbstractDBSink.java#L152-L154

psainics: The naming convention is taken from `ETLDBOutputFormat` and `PostgresETLDBOutputFormat`

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleETLDBOutputFormat.java
itsankit-google: so these validations should move to source.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleETLDBOutputFormat.java
itsankit-google: if we have these validations, why will there be a point when code reaches here without these properties present?

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceDBRecord.java
psainics: throwing error if not found

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleETLDBOutputFormat.java
psainics: Not sure, this was also taken from `ETLDBOutputFormat` and `PostgresETLDBOutputFormat` following the same style.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceDBRecord.java
```diff
@@ -116,8 +122,21 @@ protected void handleField(ResultSet resultSet, StructuredRecord.Builder recordB
...
+    boolean isFieldTypeFound = false;
+    // avoid OOB exception in case of mismatch between columnTypes and record schema fields
+    for (ColumnType columnType : columnTypes) {
+      if (columnType.getName().equals(fieldName)) {
+        sqlType = columnType.getType();
+        isFieldTypeFound = true;
+        break;
+      }
```
itsankit-google: This can be made more readable & declarative using `streams`:

```
ColumnType foundColumnType = columnTypes.stream()
    .filter(columnType -> columnType.getName().equals(fieldName))
    .findFirst() // Efficiently gets the first match wrapped in an Optional
    .orElseThrow(() -> new IllegalArgumentException(
        String.format("Unable to find the column type for field '%s'", fieldName)));

int sqlType = foundColumnType.getType();
```

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceDBRecord.java
itsankit-google: which logic are you referring to here?

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleETLDBOutputFormat.java
itsankit-google: This is just an error message, we can make the error message more descriptive here to include the config name atleast.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceDBRecord.java
psainics: https://github.com/data-integrations/database-plugins/blob/ec0383cd57b05b239d745471e19041312e2ec1d9/database-commons/src/main/java/io/cdap/plugin/db/sink/ETLDBOutputFormat.java#L347

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceDBRecord.java
psainics: updated !

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleETLDBOutputFormat.java
psainics: updated

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceDBRecord.java
itsankit-google: Why don't we use the same `modifiableColumns` list to get the columnType here as used in postgres?

https://github.com/data-integrations/database-plugins/blob/ec0383cd57b05b239d745471e19041312e2ec1d9/postgresql-plugin/src/main/java/io/cdap/plugin/postgres/PostgresDBRecord.java#L151

psainics: Missed it, updated now !

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleETLDBOutputFormat.java
psainics: These are already present 

image

image

itsankit-google (commented):
Please add e2e tests for upsert / update cases.

oracle-plugin/src/e2e-test/features/sink/OracleRunTime.feature
```diff
@@ -272,3 +272,111 @@ Feature: Oracle - Verify data transfer from BigQuery source to Oracle sink
...
+    Then Close the Plugin Properties page
+    Then Save the pipeline
+    Then Preview and run the pipeline
+    Then Verify the preview of pipeline is "Failed"
+    Examples:
+      | options |
+      | upsert  |
+      | update  |
+
+  @BQ_SOURCE_TEST @ORACLE_TEST_TABLE @Oracle_Required
```
itsankit-google: This test does not actually test upsert / update. 

The `@ORACLE_TEST_TABLE` tag just creates the source table : https://github.com/data-integrations/database-plugins/blob/ec0383cd57b05b239d745471e19041312e2ec1d9/oracle-plugin/src/e2e-test/java/io.cdap.plugin/common.stepsdesign/TestSetupHooks.java#L284

But does not insert any data, then in this case `upsert / update` will work as `insert`.

itsmekumari: Done, added the required changes.

oracle-plugin/src/e2e-test/features/sink/OracleRunTimeMacro.feature
```diff
@@ -89,8 +89,8 @@ Feature: Oracle - Verify data transfer to Oracle sink with macro arguments
     Then Validate records transferred to target table with record counts of BigQuery table
     Then Validate the values of records transferred to target Oracle table is equal to the values from source BigQuery table
 
-  @ORACLE_SOURCE_DATATYPES_TEST @ORACLE_TARGET_DATATYPES_TEST @Oracle_Required
```
itsankit-google, line 92: Why removing this test and not adding a new one for upsert case?

itsmekumari: I have added a step here for upsert case in the existing test scenario.

itsankit-google (approved):

PR #650, opened 2026-04-16
provide XML support for oracle plugin

vanshikaagupta22:
XML support

### DIFF
--- a/oracle-plugin/pom.xml
+++ b/oracle-plugin/pom.xml
@@ -121,6 +121,16 @@
       <scope>test</scope>
     </dependency>
     <dependency>
+      <groupId>com.oracle.database.xml</groupId>
+      <artifactId>xdb</artifactId>
+      <version>21.1.0.0</version>
+    </dependency>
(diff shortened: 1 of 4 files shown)


oracle-plugin/pom.xml
itsankit-google: can you add `com.oracle.database.*` here in this list and see if it works?

vanshikaagupta22: I tried, but still getting the same error - 

Image

itsankit-google: How was it added?

vanshikaagupta22: Updated in the PR.

itsankit-google: Did you check if the created plugin jar actually contains classes with prefix `com.oracle.database`?

vanshikaagupta22: I tried building the project multiple times in my local but I don't see such classes - 

Image

PR #657, opened 2026-04-24
Enable xml type support for oracle for dts-connectors only

vanshikaagupta22:
This PR enables support for the **Oracle XMLType** by mapping it to the **STRING** data type in BigQuery. Due to dependency conflicts in the pom.xml affecting CDF class loading, this feature is currently enabled exclusively for the DTS side via a new boolean flag **(isXmlTypeEnabled)**.

### DIFF
--- a/oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleConnectorConfig.java
+++ b/oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleConnectorConfig.java
@@ -42,14 +42,15 @@ public class OracleConnectorConfig extends AbstractDBSpecificConnectorConfig {
   public OracleConnectorConfig(String host, int port, String user, String password, String jdbcPluginName,
                                String connectionArguments, String connectionType, String database) {
     this(host, port, user, password, jdbcPluginName, connectionArguments, connectionType, database, null, null, null,
-         null, null);
+         null, null, null);
   }
 
   public OracleConnectorConfig(String host, int port, String user, String password, String jdbcPluginName,
                                String connectionArguments, String connectionType, String database,
                                String role, Boolean useSSL, @Nullable Boolean treatAsOldTimestamp,
                                @Nullable Boolean treatPrecisionlessNumAsDeci,
-                               @Nullable Boolean treatTimestampLTZAsTimestamp) {
+                               @Nullable Boolean treatTimestampLTZAsTimestamp,
+                               @Nullable Boolean enableXmlType) {
 
     this.host = host;
     this.port = port;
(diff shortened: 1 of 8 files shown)


oracle-plugin/src/test/java/io/cdap/plugin/oracle/OracleSchemaReaderTest.java
```diff
@@ -91,4 +92,22 @@ public void getSchema_timestampLTZFieldFalse_returnDatetime() throws SQLExceptio
     Assert.assertEquals(expectedSchemaFields.get(1).getName(), actualSchemaFields.get(1).getName());
     Assert.assertEquals(expectedSchemaFields.get(1).getSchema(), actualSchemaFields.get(1).getSchema());
   }
+
+  @Test
+  public void getSchema_xmlField_returnString() throws SQLException {
```
vikasrathee-cs, line 98: write one for false scenario, also, it should fail

vanshikaagupta22: Done

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceSchemaReader.java
```diff
@@ -103,6 +108,8 @@ public Schema getSchema(ResultSetMetaData metadata, int index) throws SQLExcepti
       case INTERVAL_YM:
       case LONG:
         return Schema.of(Schema.Type.STRING);
+      case Types.SQLXML:
```
itsankit-google, line 111: we should add a comment explaining for future context that it is only added to support DTS connector and currently not in a working state in CDAP plugin.

vanshikaagupta22: Done

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceSchemaReader.java
```diff
@@ -103,6 +108,9 @@ public Schema getSchema(ResultSetMetaData metadata, int index) throws SQLExcepti
...
+        //Enabling XML type support for DTS connectors only as it is not in working state in CDAP plugin.
```
itsankit-google: Add space after `//`

vanshikaagupta22: Done

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleConnectorConfig.java
```diff
@@ -105,6 +107,12 @@ public String getConnectionString() {
   @Nullable
   public Boolean treatTimestampLTZAsTimestamp;
 
+  @Name(OracleConstants.ENABLE_XML_TYPE)
+  @Description("A hidden field to handle mapping of Oracle XML type to BQ String.")
```
itsankit-google: This should be `A hidden field to handle mapping of Oracle XML type.`

CDAP allows many other destinations it is not specific to `BQ`, we should not add such terms in CDAP.

itsankit-google (approved):

PR #659, opened 2026-05-06
Provide Struct data type support for oracle plugin

vanshikaagupta22:
This change adds native support for resolving Oracle **STRUCT** types (Object Types) into CDAP RECORD schemas. By querying the **ALL_TYPE_ATTRS** metadata table, the schema builder dynamically processes complex structures with support for up to 4 levels of nesting. Internal attributes are first translated to standard SQL data types via a dedicated mapper before being converted into the final CDAP schema.
When reading the data, the implementation overrides the setField() function, providing a custom implementation to properly extract and map individual custom object attributes according to requirements.

Note: Users will require explicit EXECUTE privileges on the custom Oracle object types to successfully resolve the schema.

Test Coverage : 
TestCoverage


oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceDBRecord.java
```diff
@@ -232,11 +235,15 @@ private Object createOracleTimestamp(Connection connection, String timestampStri
    */
   private byte[] getBfileBytes(ResultSet resultSet, String columnName) throws SQLException {
     Object bfile = resultSet.getObject(columnName);
+    return getBfileBytes(bfile, columnName);
+  }
+
+  public static byte[] getBfileBytes(Object bfile, String columnName) {
```
vikasrathee-cs: why is this static?

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceDBRecord.java
```diff
@@ -106,13 +111,101 @@ record = recordBuilder.build();
...
+      recordBuilder.set(field.getName(), attrValue.toString());
+      return;
+    }
+    if ("oracle.sql.NCLOB".equals(attrClassName)) {
+      recordBuilder.set(field.getName(), attrValue.toString());
+      return;
+    }
+    if (attrValue instanceof BigDecimal) {
+      BigDecimal bigDecimal = (BigDecimal) attrValue;
+      if (Schema.LogicalType.DECIMAL.equals(fieldSchema.getLogicalType())) {
```
vikasrathee-cs: create a separate method for handling bigdecimal

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceSchemaReader.java
```diff
@@ -45,6 +52,46 @@ public class OracleSourceSchemaReader extends CommonSchemaReader {
   public static final int LONG = -1;
   public static final int LONG_RAW = -4;
 
+  /**
+   * Maps Oracle string data type inside UDT to their corresponding java.sql.Types integer constants
+   */
+  private static final Map<String, Integer> DATA_TYPE_MAP = new HashMap<>();
+  static {
+    DATA_TYPE_MAP.put("TIMESTAMP WITH LOCAL TZ", TIMESTAMP_LTZ);
```
vikasrathee-cs, line 60: does oracle stores this type as TIMESTAMP WITH LOCAL TIMEZONE also, verify once

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceDBRecord.java
```diff
@@ -371,6 +474,20 @@ private void handleOracleSpecificType(ResultSet resultSet, StructuredRecord.Buil
...
   }
 
+  protected StructuredRecord convertStructToRecord(Struct struct, Schema schema)
+      throws SQLException {
+    Object[] attributes = struct.getAttributes();
+    List<Schema.Field> fields = schema.getFields();
+    StructuredRecord.Builder builder = StructuredRecord.builder(schema);
+
+    for (int index = 0; index < attributes.length; index++) {
+      Schema.Field field = fields.get(index);
```
vikasrathee-cs: Normal columns are resilient to schema/result-set ordering differences because they are matched by name (resultSet.findColumn(field.getName())).
STRUCT attributes rely on Struct.getAttributes() order exactly matching schema.getFields() order. If a user edits the output schema (drops, renames, or reorders an attribute — all legitimate in CDAP), this silently writes values into the wrong fields, or throws IndexOutOfBoundsException when counts differ.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceDBRecord.java
```diff
@@ -106,13 +111,101 @@ record = recordBuilder.build();
...
+    }
+    if (attrValue instanceof SQLXML) {
+      recordBuilder.set(field.getName(), ((SQLXML) attrValue).getString());
+      return;
+    }
+    if ("oracle.sql.INTERVALDS".equals(attrClassName) || "oracle.sql.INTERVALYM".equals(attrClassName)) {
+      recordBuilder.set(field.getName(), attrValue.toString());
+      return;
+    }
+    if ("oracle.sql.NCLOB".equals(attrClassName)) {
```
vikasrathee-cs: oracle.sql.NCLOB implements java.sql.Clob, so the instanceof Clob branch (getSubString) is evaluated before the "oracle.sql.NCLOB".equals(className) branch (toString()). The NCLOB-specific branch is unreachable.

oracle-plugin/src/test/java/io/cdap/plugin/oracle/OracleSchemaReaderTest.java
```diff
@@ -94,15 +106,59 @@ public void getSchema_timestampLTZFieldFalse_returnDatetime() throws SQLExceptio
     Assert.assertEquals(expectedSchemaFields.get(1).getSchema(), actualSchemaFields.get(1).getSchema());
   }
 
+  @Test
```
vikasrathee-cs, line 98: write test cases for all data types and multiple level of structs also.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceSchemaReader.java
vanshikaagupta22: It stores as 'TIMESTAMP WITH LOCAL TZ' only across different versions.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceSchemaReader.java
```diff
@@ -128,23 +190,91 @@ public Schema getSchema(ResultSetMetaData metadata, int index) throws SQLExcepti
...
-                                     metadata.getColumnName(index)));
+                                     columnName, columnTypeName, columnName));
               return Schema.of(Schema.Type.STRING);
             }
           }
           return Schema.decimalOf(precision, scale);
         }
+      case Types.STRUCT:
+        if (connection == null) {
+          throw new SQLException("Cannot resolve STRUCT schema without a database connection. "
```
vikasrathee-cs: this error message is misleading, this will come in picture due to code issue not because of user input, it should be internal error, are you able to replicate it

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceDBRecord.java
```diff
@@ -371,6 +488,57 @@ private void handleOracleSpecificType(ResultSet resultSet, StructuredRecord.Buil
...
+          (ResultSetMetaData) descriptor.getClass().getMethod("getMetaData").invoke(descriptor);
+      for (int i = 1; i <= metaData.getColumnCount() && (i - 1) < attributes.length; i++) {
+        attributeMap.put(metaData.getColumnName(i), attributes[i - 1]);
+      }
+    } catch (SQLException | NoSuchMethodException e) {
+      throw new SQLException(String.format("Failed to retrieve attribute metadata for Oracle STRUCT schema '%s': %s",
+              schema.getRecordName(), e.getMessage()), e);
+    } catch (InvocationTargetException  | IllegalAccessException e) {
+        throw new SQLException(String.format("Unable to retrieve attribute metadata for Oracle STRUCT schema '%s'. "
+              + "Ensure the Oracle JDBC driver supports JDBC StructDescriptor metadata.",
```
vikasrathee-cs: JDBC related error needs to be changed as we are using the one that supports it, is there a way to replicate it in local or it is just CDF specific not DTS specific

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceDBRecord.java
```diff
@@ -169,6 +248,34 @@ protected void writeNonNullToDB(PreparedStatement stmt, Schema fieldSchema,
...
+        recordBuilder.set(field.getName(), bigDecimal.floatValue());
+        break;
+      case INT:
+        recordBuilder.set(field.getName(), bigDecimal.intValue());
+        break;
+      case LONG:
+        recordBuilder.set(field.getName(), bigDecimal.longValue());
+        break;
+      case STRING:
+        recordBuilder.set(field.getName(), bigDecimal.toString());
```
vikasrathee-cs: use toPlainString()

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceDBRecord.java
```diff
@@ -169,6 +248,34 @@ protected void writeNonNullToDB(PreparedStatement stmt, Schema fieldSchema,
     }
   }
 
+  private void populateDecimalValue(Object attrValue, Schema fieldSchema,
```
vikasrathee-cs: change attrValue to BigDecimal and cast it while calling this method

vikasrathee-cs (approved):
LGTM

database-commons/src/main/java/io/cdap/plugin/db/DBRecord.java
```diff
@@ -188,6 +189,12 @@ protected void handleField(ResultSet resultSet, StructuredRecord.Builder recordB
   protected void setField(ResultSet resultSet, StructuredRecord.Builder recordBuilder, Schema.Field field,
                           int columnIndex, int sqlType, int sqlPrecision, int sqlScale) throws SQLException {
     Object o = DBUtils.transformValue(sqlType, sqlPrecision, sqlScale, resultSet, columnIndex);
+    setFieldValue(resultSet.getStatement().getConnection(), recordBuilder, field, o);
+  }
+
+  protected void setFieldValue(Connection connection, StructuredRecord.Builder recordBuilder,
```
Sunish-Dahiya: Where are we using the connection?

database-commons/src/main/java/io/cdap/plugin/db/DBRecord.java
```diff
@@ -188,6 +189,12 @@ protected void handleField(ResultSet resultSet, StructuredRecord.Builder recordB
...
+                               Schema.Field field, Object o)
```
Sunish-Dahiya: Better to use a more meaningful name

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceDBRecord.java
```diff
@@ -106,13 +113,90 @@ record = recordBuilder.build();
   @Override
   protected void handleField(ResultSet resultSet, StructuredRecord.Builder recordBuilder, Schema.Field field,
                              int columnIndex, int sqlType, int sqlPrecision, int sqlScale) throws SQLException {
-    if (OracleSourceSchemaReader.ORACLE_TYPES.contains(sqlType) || sqlType == Types.NCLOB) {
+    if (OracleSourceSchemaReader.ORACLE_TYPES.contains(sqlType) || sqlType == Types.NCLOB || sqlType == Types.STRUCT) {
```
Sunish-Dahiya: can abstract this in a separate function

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceDBRecord.java
```diff
@@ -106,13 +113,90 @@ record = recordBuilder.build();
...
+      recordBuilder.set(field.getName(), attrValue.toString());
+      return;
+    }
+    if (attrValue instanceof BigDecimal) {
+      populateDecimalValue((BigDecimal) attrValue, fieldSchema, recordBuilder, field);
+      return;
+    }
+
+    if (attrValue instanceof Timestamp) {
+      Timestamp timestamp = (Timestamp) attrValue;
```
Sunish-Dahiya: too much nesting, we can move some the logic to separate functions

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceDBRecord.java
```diff
@@ -371,6 +495,57 @@ private void handleOracleSpecificType(ResultSet resultSet, StructuredRecord.Buil
...
+   * @return a case-insensitive {@code Map} linking column names to their attribute values
+   * @throws SQLException if metadata extraction fails or driver-specific methods are inaccessible
+   */
+  protected Map<String, Object> getAttributeMap(Struct struct, Schema schema, Connection connection)
+          throws SQLException {
+    Map<String, Object> attributeMap = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
+    Object[] attributes = struct.getAttributes();
+    if (attributes != null) {
+      try {
+        Object descriptor = struct.getClass().getMethod("getDescriptor").invoke(struct);
```
Sunish-Dahiya, line 556: why we need reflection here?

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceSchemaReader.java
```diff
@@ -88,14 +134,27 @@ public OracleSourceSchemaReader(@Nullable String sessionID, boolean isTimestampO
   @Override
   public Schema getSchema(ResultSetMetaData metadata, int index) throws SQLException {
     int sqlType = metadata.getColumnType(index);
+    String owner = (metadata.getColumnTypeName(index) != null
```
Sunish-Dahiya: too much logic in a single line, would suggest to separate it out to improve the readability

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceSchemaReader.java
```diff
@@ -88,14 +134,27 @@ public OracleSourceSchemaReader(@Nullable String sessionID, boolean isTimestampO
...
+            && metadata.getColumnTypeName(index).contains(".")) ? metadata.getColumnTypeName(index)
+            .substring(0, metadata.getColumnTypeName(index).lastIndexOf('.')) : null;
+
+    return getSchemaMapping(sqlType, metadata.getColumnClassName(index), metadata.getPrecision(index),
+            metadata.getScale(index), metadata.getColumnName(index), metadata.getColumnTypeName(index),
+            metadata.isSigned(index), owner, 0);
+  }
+
+  public Schema getSchemaMapping(int sqlType, String columnClassName, int columnPrecision,
```
Sunish-Dahiya: too many arguments in a function

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceSchemaReader.java
```diff
@@ -128,23 +188,93 @@ public Schema getSchema(ResultSetMetaData metadata, int index) throws SQLExcepti
...
+  private Schema getStructSchema(Connection connection, String typeName, String owner, String recordName, int level)
+          throws SQLException {
+    List<Schema.Field> fields = new ArrayList<>();
+    String sql = "SELECT * FROM ALL_TYPE_ATTRS WHERE TYPE_NAME = ? AND OWNER = ? ORDER BY ATTR_NO";
+
+    try (PreparedStatement stmt = connection.prepareStatement(sql)) {
+      stmt.setString(1, typeName.substring(typeName.lastIndexOf('.') + 1));
+      stmt.setString(2, owner);
+
+      try (ResultSet attrRs = stmt.executeQuery()) {
```
Sunish-Dahiya: prefer full names instead of acronyms

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceSchemaReader.java
```diff
@@ -128,23 +188,93 @@ public Schema getSchema(ResultSetMetaData metadata, int index) throws SQLExcepti
...
+        while (attrRs.next()) {
+          String attrName = attrRs.getString("ATTR_NAME");
+          String attrTypeName = attrRs.getString("ATTR_TYPE_NAME");
```
Sunish-Dahiya: constants, apply to whole function

Sunish-Dahiya (changes requested):

database-commons/src/main/java/io/cdap/plugin/db/DBRecord.java
vanshikaagupta22: Required in the subclass (OracleAdapterDbRecord)

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceSchemaReader.java
vanshikaagupta22: Created helper class encapsulating column metadata parameters.

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceDBRecord.java
vanshikaagupta22: We use reflection to extract column names via getDescriptor() without adding a compile-time dependency on ojdbc (which is provided at runtime).

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceDBRecord.java
```diff
@@ -106,13 +113,111 @@ record = recordBuilder.build();
...
+                                    StructuredRecord.Builder recordBuilder, Schema.Field field,
+                                    Connection connection) throws SQLException {
+    if (Schema.LogicalType.DATETIME.equals(fieldSchema.getLogicalType())) {
+      recordBuilder.setDateTime(field.getName(), timestamp.toLocalDateTime());
+    } else {
+      super.populateRecordField(connection, recordBuilder, field, timestamp);
+    }
+  }
+
+  private void handleOffsetDateTimeValue(OffsetDateTime offsetDateTime, Schema fieldSchema,
```
Sunish-Dahiya, line 193: These types would be at top level/root level also and their handling should already exist, any specific reason we are handling record members separately?

Sunish-Dahiya: this comment applies to all types handling

Sunish-Dahiya (commented):
1. Can you please add the testing report.
2. UTs are not following standard convention can you please fix that.
3. Please also attach coverage details/report.
4. I think we would also need to add new ITNs to cover this logic.
5. Since this change will require new permission, have we closed the discussion with CDF team ?

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceDBRecord.java
vanshikaagupta22: Top-level columns are read directly from the **ResultSet** using column positions, but STRUCT members come from struct.getAttributes() as raw Java objects where no ResultSet is available. Also, Oracle returns all STRUCT numbers as BigDecimal, dates as Timestamp/OffsetDateTime, and LOBs as Clob/Blob objects. We need this handling in populateRecordField to convert these raw objects into the correct types so that their value can be stored properly.

vanshikaagupta22:
> 1. Can you please add the testing report.
> 2. UTs are not following standard convention can you please fix that.
> 3. Please also attach coverage details/report.
> 4. I think we would also need to add new ITNs to cover this logic.
> 5. Since this change will require new permission, have we closed the discussion with CDF team ?

1. The testing report is already attached to the buganizer.
2. Uts are updated now.
3. I have attached the Test Coverage report in the Description.
4. I will update the ITNs in the DTS 3p CL.
5. Approved from CDF team

sahusanket (approved):

oracle-plugin/src/main/java/io/cdap/plugin/oracle/OracleSourceDBRecord.java
Sunish-Dahiya: May be then we can move these in a separate Struct specific class

Sunish-Dahiya (commented):
The test report says coverage is around 30%, why it's so low?

PR #663, opened 2026-05-06
Bump up version - remove snapshot from 1.11.13

vanshikaagupta22:
- Bump up - Remove snapshot from 1.11.13

- Update cdap-e2e-framework version to 0.3.2

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -20,7 +20,7 @@
 
   <groupId>io.cdap.plugin</groupId>
   <artifactId>database-plugins-parent</artifactId>
-  <version>1.11.13-SNAPSHOT</version>
+  <version>1.11.13</version>
   <packaging>pom</packaging>
   <name>Database Plugins</name>
   <description>Collection of database plugins</description>
@@ -593,6 +593,7 @@
       <properties>
         <testSourceLocation>src/e2e-test/java</testSourceLocation>
         <TEST_RUNNER>TestRunner.java</TEST_RUNNER>
+        <guava.version>32.1.2-jre</guava.version>
       </properties>
       <build>
         <testResources>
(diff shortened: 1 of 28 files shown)


pom.xml
```diff
@@ -706,7 +707,7 @@
         <dependency>
           <groupId>io.cdap.tests.e2e</groupId>
           <artifactId>cdap-e2e-framework</artifactId>
-          <version>0.3.0</version>
+          <version>0.5.0-SNAPSHOT</version>
```
itsankit-google, line 710: This is not correct, the tests will get executed on `develop` branch of `cdap-e2e-tests` but this branch should run against `release/6.10` branch of `cdap-e2e-tests`.

Vipinofficial11: Yes, Ideally this branch should be running against a stable release of the e2e framework (like release/6.10).

There were some test failures on this PR so did some debugging, here is what is happening under the hood based on my current understanding :

- In our .github/workflows/e2e.yml, the [step that clones cdapio/cdap-e2e-tests](https://github.com/data-integrations/database-plugins/blob/release/1.11/.github/workflows/e2e.yml#L63) doesn't specify a branch ref because of this, GitHub Actions defaults to pulling the develop branch (attached a ss of this below)

[Github Action Run](https://github.com/data-integrations/database-plugins/actions/runs/26020955815/job/76482130825)
Screenshot 2026-05-19 at 5 35 10 PM

- Since the action clones develop branch, it executes the newest version of the [Python runner script](https://github.com/data-integrations/database-plugins/blob/release/1.11/.github/workflows/e2e.yml#L105). This script automatically [pulls and spins up the latest CDAP Sandbox](https://github.com/cdapio/cdap-e2e-tests/blob/develop/src/main/scripts/run_e2e_test.py#L46)

- Now the pipeline forces the tests to run against the newest CDAP UI, leaving the framework dependency on an older version results in failures (the UI locators and error message assertions are totally out of sync). 

Please share if I missed something and this is not the case on how things are actually working in the existing setup.

itsankit-google: This is incorrect, we should specify the branch in github actions as `release/6.10` here : https://github.com/data-integrations/database-plugins/blob/release/1.11/.github/workflows/e2e.yml#L67

For example in google-cloud : https://github.com/data-integrations/google-cloud/blob/release/0.24/.github/workflows/e2e.yml#L64

Vipinofficial11: I see, got it. Will raise a separate PR to checkin the Github Action change to define the release version of cdap-e2e-tests.

PR #675, opened 2026-06-29
Add TRANSACTION_ISOLATION_LEVEL config in cloudsql-postgres and cloudsql-mysql plugins for develop

vedanshugarg04:
Add TRANSACTION_ISOLATION_LEVEL config in cloudsql-postgres and cloudsql-mysql plugins

### DIFF
--- a/cloudsql-postgresql-plugin/src/main/java/io/cdap/plugin/cloudsql/postgres/CloudSQLPostgreSQLSource.java
+++ b/cloudsql-postgresql-plugin/src/main/java/io/cdap/plugin/cloudsql/postgres/CloudSQLPostgreSQLSource.java
@@ -157,6 +157,13 @@ public class CloudSQLPostgreSQLSource
     @Description("The existing connection to use.")
     private CloudSQLPostgreSQLConnectorConfig connection;
 
+    @Name(TRANSACTION_ISOLATION_LEVEL)
+    @Description("Transaction isolation level for queries run by this source.")
+    @Nullable
+    private String transactionIsolationLevel;
+
+    private static final String DEFAULT_TRANSACTION_ISOLATION_LEVEL = "TRANSACTION_READ_COMMITTED";
+
     @Override
     protected Map<String, String> getDBSpecificArguments() {
       return Collections.emptyMap();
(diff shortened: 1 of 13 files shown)


cloudsql-postgresql-plugin/src/main/java/io/cdap/plugin/cloudsql/postgres/CloudSQLPostgreSQLSource.java
```diff
@@ -173,6 +178,11 @@ protected CloudSQLPostgreSQLConnectorConfig getConnection() {
       return connection;
     }
 
+    @Override
+    public String getTransactionIsolationLevel() {
+      return transactionIsolationLevel;
```
vikasrathee-cs: Add a constant for default isolation level and then return that in case of null for both DBs.

cloudsql-postgresql-plugin/widgets/CloudSQLPostgreSQL-batchsink.json
```diff
@@ -84,6 +84,19 @@
           "label": "Password",
           "name": "password"
         },
+        {
+          "widget-type": "select",
```
vikasrathee-cs: this property is already available in this plugin, why it is added again, remove it

cloudsql-postgresql-plugin/src/main/java/io/cdap/plugin/cloudsql/postgres/CloudSQLPostgreSQLSource.java
vedanshugarg04: done

cloudsql-postgresql-plugin/widgets/CloudSQLPostgreSQL-batchsink.json
vedanshugarg04: done

cloudsql-postgresql-plugin/widgets/CloudSQLPostgreSQL-batchsink.json
```diff
@@ -174,13 +174,12 @@
           "label": "Transaction Isolation Level",
           "name": "transactionIsolationLevel",
           "widget-attributes": {
-            "default": "TRANSACTION_READ_COMMITTED",
```
vikasrathee-cs, line 177: This should not change, keep it as it is. and keep the documentation also same

vedanshugarg04: done

shruthi713 (approved):

PR #680, opened 2026-08-23
Addition of Databricks Plugin

vikasrathee-cs:
Addition of Databricks Plugin


databricks-plugin/src/main/java/io/cdap/plugin/databricks/DatabricksConnector.java
```diff
@@ -0,0 +1,162 @@
...
+      return String.format("`%s`.`%s`", schema, table);
+    }
+    if (schema == null) {
+      return String.format("`%s`.`%s`", database, table);
+    }
+    return String.format("`%s`.`%s`.`%s`", database, schema, table);
+  }
+
+  @Override
+  protected String getRandomQuery(String tableName, int limit) {
```
sahusanket, line 128: getRandomQuery() is only ever reached from AbstractDBSpecificConnector.getTableQuery(..., sampleType, ...), which is only invoked when the UI passes a sampleType. 

The UI only offers a sample type that the connector declares via ConnectorSpec.Builder.addSupportedSampleType(...) 
and DatabricksConnector.setConnectorSpec() never calls it (compare MysqlConnector, which calls .addSupportedSampleType(SampleType.RANDOM) and STRATIFIED). 

So the override can never fire.

databricks-plugin/src/main/java/io/cdap/plugin/databricks/DatabricksConnector.java
```diff
@@ -0,0 +1,162 @@
...
+    return record.getRecord();
+  }
+
+  @Override
+  protected SchemaReader getSchemaReader(String sessionID) {
+    return new DatabricksSchemaReader(sessionID);
+  }
+
+  @Override
+  protected String getTableName(String database, String schema, String table) {
```
sahusanket, line 114: If the value has a backtick (`) in it, it will malform things going forward. 
Please handle validation.

databricks-plugin/src/main/java/io/cdap/plugin/databricks/DatabricksConstants.java
```diff
@@ -0,0 +1,31 @@
...
+package io.cdap.plugin.databricks;
+
+/** Databricks constants. */
+public final class DatabricksConstants {
+
+  private DatabricksConstants() {
+  }
+
+  public static final String PLUGIN_NAME = "Databricks";
+  public static final String DRIVER_CLASS_NAME = "com.databricks.client.jdbc.Driver";
```
sahusanket, line 26: Is this used any where?

databricks-plugin/src/test/java/io/cdap/plugin/databricks/DatabricksSchemaReaderTest.java
```diff
@@ -0,0 +1,90 @@
...
+    Assert.assertEquals(Schema.of(Schema.Type.STRING), schemaReader.getSchema(metadata, 6));
+    Assert.assertEquals(Schema.of(Schema.Type.STRING), schemaReader.getSchema(metadata, 7));
+    Assert.assertEquals(Schema.of(Schema.Type.STRING), schemaReader.getSchema(metadata, 8));
+    Assert.assertEquals(Schema.of(Schema.Type.STRING), schemaReader.getSchema(metadata, 9));
+  }
+
+  @Test
+  public void testShouldIgnoreColumn() throws SQLException {
+    DatabricksSchemaReader schemaReader = new DatabricksSchemaReader("sessionID");
+    Map<Integer, String> names = new java.util.HashMap<>();
```
sahusanket, line 79: Please use import. Don't use FQNs unless necessary .

databricks-plugin/src/test/java/io/cdap/plugin/databricks/DatabricksSchemaReaderTest.java
```diff
@@ -0,0 +1,90 @@
...
+    typeNames.put(2, "BIGINT");
+    typeNames.put(3, "TIMESTAMP");
+    typeNames.put(4, "TIMESTAMP_NTZ");
+    typeNames.put(5, "DATE");
+    typeNames.put(6, "VARIANT");
+    typeNames.put(7, "STRUCT");
+    typeNames.put(8, "ARRAY");
+    typeNames.put(9, "MAP");
+
+    ResultSetMetaData metadata = createMockMetadata(typeNames, java.util.Collections.emptyMap());
```
sahusanket, line 63: Please use import. Don't use FQNs unless necessary .

databricks-plugin/src/test/java/io/cdap/plugin/databricks/DatabricksSchemaReaderTest.java
```diff
@@ -0,0 +1,90 @@
...
+import java.lang.reflect.Proxy;
+import java.sql.ResultSetMetaData;
+import java.sql.SQLException;
+import java.util.Map;
+
+public class DatabricksSchemaReaderTest {
+
+  private ResultSetMetaData createMockMetadata(Map<Integer, String> columnTypeNames,
+                                                Map<Integer, String> columnNames) {
+    return (ResultSetMetaData) Proxy.newProxyInstance(
```
sahusanket, line 32: Why this approach ? is it not diable via Mockito ?  Mockito.mock(ResultSetMetaData.class) with when(...).thenReturn(...) would be a lot more readable ..

databricks-plugin/src/main/java/io/cdap/plugin/databricks/DatabricksSchemaReader.java
```diff
@@ -0,0 +1,79 @@
...
+
+    if (typeName != null) {
+      if (typeName.equalsIgnoreCase("INT") || typeName.equalsIgnoreCase("INTEGER") ||
+          typeName.equalsIgnoreCase("SMALLINT") || typeName.equalsIgnoreCase("TINYINT")) {
+        return Schema.of(Schema.Type.INT);
+      }
+      if (typeName.equalsIgnoreCase("BIGINT")) {
+        return Schema.of(Schema.Type.LONG);
+      }
+      if (typeName.equalsIgnoreCase("TIMESTAMP") || typeName.equalsIgnoreCase("TIMESTAMP_NTZ") ||
```
sahusanket, line 54: Databricks TIMESTAMP is timestamp-with-local-time-zone (an instant); TIMESTAMP_NTZ is the zone-less one. Mapping both to CDAP DATETIME drops the zone and will shift values based on the executor's default TZ. I think this should be TIMESTAMP → LogicalType.TIMESTAMP_MICROS and TIMESTAMP_NTZ → DATETIME. Also, TIMESTAMPTZ isn't a Databricks type name, and the BIGINT/DATE branches are redundant — CommonSchemaReader already maps those from the SQL type.

sahusanket (changes requested):

databricks-plugin/src/main/java/io/cdap/plugin/databricks/DatabricksConnector.java
vishwasvaidya-cloudsufi: Added SampleType.RANDOM and SampleType.STRATIFIED in setConnectorSpec() and updated getRandomQuery() and getStratifiedQuery() similar to MysqlConnector

databricks-plugin/src/main/java/io/cdap/plugin/databricks/DatabricksConstants.java
vishwasvaidya-cloudsufi: Removed the unused DRIVER_CLASS_NAME constant from DatabricksConstants.

databricks-plugin/src/main/java/io/cdap/plugin/databricks/DatabricksConnector.java
vishwasvaidya-cloudsufi: Tested this on Databricks — Databricks does not allow creating catalog, schema, or table names with an unescaped backtick (`) and fails with [PARSE_SYNTAX_ERROR]. Also, getTableName() here is consistent with MysqlConnector and AbstractDBSpecificConnector, where the identifiers come directly from browsing the connection metadata.

databricks-plugin/src/test/java/io/cdap/plugin/databricks/DatabricksSchemaReaderTest.java
vishwasvaidya-cloudsufi: Done

databricks-plugin/src/test/java/io/cdap/plugin/databricks/DatabricksSchemaReaderTest.java
vishwasvaidya-cloudsufi: Done

databricks-plugin/src/test/java/io/cdap/plugin/databricks/DatabricksSchemaReaderTest.java
vishwasvaidya-cloudsufi: Updated DatabricksSchemaReaderTest to use Mockito.mock(ResultSetMetaData.class) with Mockito.when(...).thenReturn(...) instead of Proxy.newProxyInstance.

databricks-plugin/src/main/java/io/cdap/plugin/databricks/DatabricksSchemaReader.java
vishwasvaidya-cloudsufi: Updated DatabricksSchemaReader so TIMESTAMP_NTZ maps to Schema.LogicalType.DATETIME, removed TIMESTAMPTZ and redundant SQL type branches (BIGINT, DATE, INT, SMALLINT, TINYINT, TIME, TIMESTAMP), and delegated standard JDBC types (including TIMESTAMP $\rightarrow$ TIMESTAMP_MICROS) to super.getSchema()

sahusanket:
Hi, I don't think you have pushed the changes yet .

database-commons/src/main/java/io/cdap/plugin/db/connector/AbstractDBSpecificConnector.java
```diff
@@ -104,8 +104,12 @@ public InputFormatProvider getInputFormatProvider(ConnectorContext context, Samp
     String tableQuery = getTableQuery(path.getDatabase(), path.getSchema(), path.getTable(), request.getLimit(),
       request.getProperties().get("sampleType"), request.getProperties().get("strata"), sessionID);
     DataDrivenETLDBInputFormat.setInput(connectionConfigAccessor.getConfiguration(), getDBRecordType(),
-      tableQuery, null, false);
+      tableQuery, null, isAutoCommitEnabled());
```
sahusanket: getInputFormatProvider() runs for every DB connector in this repo — MySQL, Postgres, Oracle, SQL Server, Redshift — not just Databricks.

The defaulting looks correct to me (isAutoCommitEnabled() → false reproduces the old hard-coded false, and a null isolation level leaves TransactionIsolationLevel.CONF_KEY unset so getLevel(null) still yields SERIALIZABLE). But nothing asserts that, so a future change to either default would silently alter connection behaviour for every other plugin with no test failing.

Could you add coverage in database-commons for:

- the base hooks returning false / null;
- a non-overriding connector leaving AUTO_COMMIT_ENABLED and TransactionIsolationLevel.CONF_KEY exactly as before this change;
an overriding connector setting both.
database-commons/src/test/java/io/cdap/plugin/db/source/DataDrivenETLDBInputFormatTest.java is a reasonable place

databricks-plugin/pom.xml
```diff
@@ -0,0 +1,127 @@
...
+      <groupId>io.cdap.plugin</groupId>
+      <artifactId>hydrator-common</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>com.google.guava</groupId>
+      <artifactId>guava</artifactId>
+    </dependency>
+
+    <!-- test dependencies -->
+    <dependency>
```
sahusanket: This dependency isn't referenced by any test in the PR — nothing loads the driver, so it's currently dead weight in the build.

That's really a symptom of the wider gap: there's no DatabricksPluginTestBase, DatabricksPluginTestSuite, DatabricksSourceTestRun, DatabricksFailedConnectionTest or DatabricksDBRecordUnitTest here. Every other plugin in the repo ships that set — see amazon-redshift-plugin/src/test/.../RedshiftPluginTestBase.java (218 lines) as the closest template. The two unit tests in this PR don't touch DatabricksSource, DatabricksDBRecord, or any connection path.

For a brand-new plugin I'd like at least a DatabricksPluginTestBase following the Redshift pattern, which would also give this dependency a purpose. Separately, could you attach evidence of a real end-to-end run (browse → get schema → sample → pipeline read) against a SQL warehouse, over a table containing TIMESTAMP, TIMESTAMP_NTZ, DATE, DECIMAL(38,10), ARRAY, MAP, STRUCT, VARIANT and a NULL-only column? That's the set most likely to break and none of it is covered today.

